PR #502, opened 2020-09-24
Fix single node setup using setup_singlenode API

mazinamdar:

### DIFF
--- a/api/python/provisioner/commands/configure_setup.py
+++ b/api/python/provisioner/commands/configure_setup.py
@@ -137,7 +137,7 @@ class StorageEnclosureParamsValidation:
 
     def __attrs_post_init__(self):
         params = attr.asdict(self)
-        types = ['JBOD', 'virtual']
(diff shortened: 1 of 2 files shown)


api/python/provisioner/commands/configure_setup.py
```diff
@@ -137,7 +137,8 @@ class StorageEnclosureParamsValidation:
 
     def __attrs_post_init__(self):
         params = attr.asdict(self)
-        if params['type'] == 'JBOD':
+        types = ['JBOD', 'virtual']
```
ypise, line 140: Here we need:
['JBOD', 'RBOD', 'virtual', 'other']

ypise (approved):

api/python/provisioner/commands/configure_setup.py
mazinamdar: comment is addressed

PR #521, opened 2020-09-25
[EOS-13584][EOS-13228] Improves default file logging for setup cmds

andkononykhin2:
- improved default configuration for for setup commands:
  - file logging is turned on by default
  - rsyslog logging is turned off by default (fix for EOS-13584)


mazinamdar (approved):
looks good to me , do we need to add auto_deploy command in LOG_FORCED_LOGFILE_CMDS in config.py ?

ypise (approved):

PR #522, opened 2020-09-25
EOS-13679: ssh config file not found: /root/.ssh/config

ypise:


83bhp (approved):
looks good

PR #523, opened 2020-09-25
Set target build  in release.sls with value --cortx-url

mazinamdar:


api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1299,39 +1299,49 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
...
-            ssh_client.cmd_run(
-                (
-                    'provisioner pillar_set --fpath release.sls'
-                    f' release/target_build \'"{run_args.target_build}"\''
-                ), targets=run_args.primary.minion_id
-            )
-
-            logger.info("Get release factory version")
+        logger.info("Updating target build pillar")
+        if run_args.url_cortx:
```
andkononykhin2, line 1303: That all should checked in  __post_attr_init_. Just imagine: setup almost completed with some wrong values that we may check at the very beggining and failed then.

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1299,39 +1299,49 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
...
+            url = run_args.url_cortx
+        elif run_args.target_build:
             if run_args.dist_type == config.DistrType.BUNDLE:
                 url = f"{run_args.target_build}/cortx_iso"
             else:
                 url = run_args.target_build
+        elif run_args.iso_cortx:  # Fix Me update with proper path
+            logger.info("Update target with mounted path")
```
andkononykhin2, line 1311: In that case url would be undefined and we will fail few lines later. Please test that code.

andkononykhin2 (changes requested):
I don't understand the logic. We are fixing only one case: source is rpm and urls are passed.
Questions:
- should we support only one url for cortx
- in case of 2 urls how our states will reflect that

Also you should't break all other setups: bundled local isos, bundled remote isos, local repo.

andkononykhin2:
@ypise @mazinamdar target_build is not only problem here,  let's close that PR on behalf of #529

andkononykhin2:
superseded by #530

PR #525, opened 2020-09-25
EOS-13687: Use grains to check if server is physical or vm

83bhp:


ipoddubnyy (approved):
approved

dmitrydb (approved):

PR #530, opened 2020-09-26
[EOS-13651][EOS-13686] improves rpm based setup support

andkononykhin2:
- adds support for separate remotely mounted ISO files for 3rd party dependencies (deps bundle url)
- adds support of non local source of provisioner setup for replace node logic

### DIFF
--- a/srv/components/defaults.yaml
+++ b/srv/components/defaults.yaml
@@ -17,10 +17,28 @@
 
 tmp_dir: /tmp
 
+
+# the following logic evaluates:
+#   - release_repo: a repo of the product release
+#   - 3rd party dependencies repositories:
+#     - os repositories (optionally)
+#     - commons_repo
+#     - epel_repo
+#     - saltstack_repo
+
+{% set os_repos = [] %}
+{% set target_build = pillar['release']['target_build'] %}
+{% set deps_bundle_url = salt['pillar.get']('release:deps_bundle_url') %}
+{% set release_repo = target_build %}
+
+# for bundle release target_build is a base url of mounted repositories,
+# might be local (file://) or remote (http(s)://)
 {% if pillar['release']['type'] == 'bundle' %}
+  # FIXME os repos for redhat with subscription disabled
 
-  {% set base_bundle_url = pillar['release']['target_build'] %}
+  {% set base_bundle_url = target_build %}
 
+  # FIXME too complicated command
   {% if ("RedHat" in grains['os']) and (salt['cmd.shell']('subscription-manager list | grep -m1 -A4 -Pe "Product Name:.*Red Hat Enterprise Linux Server"|grep -Pe "Status:.*Subscribed"')) %}
     {% set system_repo = '/'.join([base_bundle_url, 'rhel7.7']) %}
   {% else %}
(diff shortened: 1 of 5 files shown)


srv/components/defaults.yaml
```diff
@@ -17,122 +17,145 @@
...
   {% else %}
-    {% set cortx_commons_url = pillar['commons']['cortx_commons']['CentOS'] %}
+    {% set commons_repo = pillar['commons']['cortx_commons']['CentOS'] %}
   {% endif %}
 
+  {% set epel_repo = "http://ssc-satellite1.colo.seagate.com/pulp/repos/EOS/Library/custom/EPEL-7/EPEL-7/" %}
+    # url: https://dl.fedoraproject.org/pub/epel/7/x86_64/
+  {% set saltstack_repo = "https://repo.saltstack.com/py3/redhat/$releasever/$basearch/3001" %}
+
+{% endif %}
```
ypise, line 81: In the past I have seen a JINJA without a closing else after an if/elif causes minion to return false.
Not sure it would happen the same way in this case, but there's no harm in addition of an empty else block with a blank or default statement

ypise (changes requested):
@andkononykhin2 
We need a discussion with @ipoddubnyy regarding the options that we should support.
I feel we should support only 2:

1. Downloadable ISO for Cortx stack and 3rd_Party
2. A remote hosted repo representing a bundle of cortx_iso (this is not a very good name choice now that I think of it again) and 3rd_party repos

ypise:
@andkononykhin2 
Don't rush with frequent changes. I understand you are trying to improvise, but that would confuse QA.
Let's take time until Tuesday EoD to finalize the CLI args and address the fixes once and for all.

Also, I would like to drop the nodes arguments for all commands except for `replace_node`, which doesn't have config.ini file.
For the rest the srvnode args should be picked from the config.ini file as default. The CLI args would be over-rides.
It's not good from usability aspect if the user has to mention the same data twice.

srv/components/defaults.yaml
andkononykhin2: Done

andkononykhin2:
> @andkononykhin2
> Don't rush with frequent changes. I understand you are trying to improvise, but that would confuse QA.
> Let's take time until Tuesday EoD to finalize the CLI args and address the fixes once and for all.

Let's don't expect that. Each sw (should) change and versioning is the only and necessary remedy here.
For now we can just decide on some good design for nearest milestones.

> Also, I would like to drop the nodes arguments for all commands except for `replace_node`, which doesn't have config.ini file.
> For the rest the srvnode args should be picked from the config.ini file as default. The CLI args would be over-rides.
> It's not good from usability aspect if the user has to mention the same data twice.
Generally you raised good points but here are my concerns:
- config.ini is to declare the whole setup, command args might be used for some partial (subset) setup and partial re-appliance (as long as we have idempotent logic now)
- replace_node command doesn't need config.ini since everything is already set in pillar but it does need targetting in command line
- command args may override some generic config (agree with you here)

So, I think we shouldn't address that in the current PR.

ypise (approved):

PR #534, opened 2020-09-27
[EOS-13228] disables regression of #521

andkononykhin2:
hotfix to disable logging configuration regression of #521


ypise:
@andkononykhin 
I've changed the base to pre-cortx-1.0 branch as we would use this branch as buffer, just like we used to do with dev branch.

PR #537, opened 2020-09-27
EOS-13228: Better log defaults

ypise:


83bhp (approved):

api/python/provisioner/__main__.py
```diff
@@ -39,6 +39,9 @@
     log
 )
 from .commands import commands
+from .commands.setup_provisioner import (
+    SetupCmdBase, RunArgsSetupProvisionerGeneric
```
ypise, line 43: @andkononykhin 
This was the reason the commit was reverted.
Due to this, nodes was expected as a mandatory positional argument and the commands (none of the API ones) never proceeded beyond this point.

We need to understand the flow and fix it

api/python/provisioner/__main__.py
```diff
@@ -114,18 +117,54 @@ def _set_logging(output_type, log_args=None):
...
+            # enable file logging
             setattr(log_args, config.LOG_FILE_HANDLER, True)
             # set file to log if default value is set
             filename_attr = f'{config.LOG_FILE_HANDLER}_filename'
             if (
                 getattr(log_args, filename_attr) ==
                 attr.fields_dict(type(log_args))[filename_attr].default
             ):
-                filename = _generate_logfile_filename(log_args.cmd)
+                if isinstance(cmd_inst, SetupCmdBase):
```
ypise, line 146: Separate logging fixes and design changes into different PRs.
Don't club the two as it becomes difficult to trace the changes.

ypise:
@83bhp 
You have approved a change that had broken the API functionality completely.

andkononykhin2:
> @83bhp
> You have approved a change that had broken the API functionality completely.

> @83bhp
> You have approved a change that had broken the API functionality completely.

@ypise The change #521 impacted only one special case when one doesn't specify file logging settings. Since in all our docs we declare that (`--logfile --logfile-filename somefile.log`) the issue won't appear for end users most of the time. So, `broken the API functionality completely` is not true.

Makes sense to mention that #521 was approved and merged by you.
Please either adjust or remove such inaccurate and blaming statement.

ypise:
@andkononykhin2 
Merge when you are ready.

PR #539, opened 2020-09-27
EOS-13679: ssh config file not found: /root/.ssh/config

ypise:
--------------------------------------------------------------


cli/cortx-prvsnr-cli.spec
```diff
@@ -62,19 +88,29 @@ mkdir -p /root/.ssh
 
 # Ensure update replaces the keys
 if [[ -e /root/.ssh/id_rsa_prvsnr ]]; then
-  rm -f /root/.ssh/id_rsa_prvsnr
+  rm -f /root/.ssh/id_rsa_prvsnr || true
   rm -f /root/.ssh/id_rsa_prvsnr.pub || true
 fi
 
-cp -pr /opt/seagate/cortx/provisioner/files/.ssh/* /root/.ssh/
+cp -pr /opt/seagate/cortx/provisioner/files/.ssh/id_rsa_prvsnr /root/.ssh/
```
ypise, line 95: Create `/root/.ssh` if not exists.

cli/cortx-prvsnr-cli.spec
```diff
@@ -20,6 +20,28 @@ CORTX Provisioner Command line interface. Provides utilities to deploy CORTX Obj
 %prep
 %setup -n %{name}-%{version}-%{_cortx_prvsnr_git_ver}
 
+/bin/cat <<EOL > ssh_config
+Host srvnode-1 srvnode-1.colo.seagate.com
```
ipoddubnyy, line 24: no, no, no. please DO NOT hard code colo.seagate.com here. our lab is not the only environment where the systems will operate

cli/cortx-prvsnr-cli.spec
```diff
@@ -20,6 +20,28 @@ CORTX Provisioner Command line interface. Provides utilities to deploy CORTX Obj
...
+    HostName srvnode-1.colo.seagate.com
```
ipoddubnyy, line 25: same as above - do not hardcode "colo.seagate.com"

cli/cortx-prvsnr-cli.spec
```diff
@@ -20,6 +20,28 @@ CORTX Provisioner Command line interface. Provides utilities to deploy CORTX Obj
...
+    User root
+    UserKnownHostsFile /dev/null
+    StrictHostKeyChecking no
+    IdentityFile /root/.ssh/id_rsa_prvsnr
+    IdentitiesOnly yes
+    LogLevel ERROR
+    BatchMode yes
+
+Host srvnode-2 srvnode-2.colo.seagate.com
```
ipoddubnyy, line 34: no on hardcoding "colo.seagate.com"

cli/cortx-prvsnr-cli.spec
```diff
@@ -20,6 +20,28 @@ CORTX Provisioner Command line interface. Provides utilities to deploy CORTX Obj
...
+    HostName srvnode-2.colo.seagate.com
```
ipoddubnyy, line 35: no on hardcoding "colo.seagate.com"

cli/cortx-prvsnr-cli.spec
```diff
@@ -54,27 +80,38 @@ rm -rf %{buildroot}
...
+if [[ ! -e /root/.ssh/ssh_config ]]; then
+  /bin/cp -pr /opt/seagate/cortx/provisioner/files/.ssh/ssh_config /root/.ssh/
 fi
 
-cp -pr /opt/seagate/cortx/provisioner/files/.ssh/* /root/.ssh/
-cat /root/.ssh/id_rsa_prvsnr.pub >>/root/.ssh/authorized_keys
-chmod 700 /root/.ssh/
-chmod 600 /root/.ssh/*
+/bin/chmod 700 /root/.ssh/
+/bin/chmod 644 /root/.ssh/*
```
ipoddubnyy, line 105: config file should have permissions 600. this line will cause ssh to faile with "bad permission or ownership of the file" error. 
config file should be configured separately in this section

ipoddubnyy (changes requested):
changes requested

cli/cortx-prvsnr-cli.spec
sradhanand-pati: Yes.. you are right. i will update it. Thanks
```     
~/.ssh/config
             This is the per-user configuration file.  The file format and
             configuration options are described in ssh_config(5).  Because of
             the potential for abuse, this file must have strict permissions:
             read/write for the user, and not writable by others.
```
http://linuxcommand.org/lc3_man_pages/ssh1.html

cli/cortx-prvsnr-cli.spec
sradhanand-pati: This is just a template.. we would be updating it with hostnames in later stages of provisioning.

cli/cortx-prvsnr-cli.spec
sradhanand-pati: it creates automatically

cli/cortx-prvsnr-cli.spec
```diff
@@ -28,20 +50,24 @@ CORTX Provisioner Command line interface. Provides utilities to deploy CORTX Obj
...
+
+/bin/mkdir -p %{buildroot}/opt/seagate/cortx/provisioner/{cli,files/etc,files/.ssh}
 
-mkdir -p %{buildroot}/opt/seagate/cortx/provisioner/{cli,files/etc}
+/bin/cp -pr cli/src %{buildroot}/opt/seagate/cortx/provisioner/cli
+/bin/cp -pr files/etc/yum.repos.d %{buildroot}/opt/seagate/cortx/provisioner/files/etc
 
-cp -pr cli/src %{buildroot}/opt/seagate/cortx/provisioner/cli
-cp -pr files/etc/yum.repos.d %{buildroot}/opt/seagate/cortx/provisioner/files/etc
+/bin/ssh-keygen -o -q -t rsa -b 4096 -a 100 -N '' -f id_rsa_prvsnr
```
andkononykhin2, line 61: @sradhanand-pati 
>This is only run when creating a package, not when the end-user installs the package
(from [here](https://rpm-packaging-guide.github.io/#:~:text=A%20SPEC%20file%20can%20be,are%20used%20in%20the%20Body.))

so generated keys would be part of rpm package, I think it's a wrong direction in scope of the secutiry. If you want a dynamic way you need to use scriptlets that are called during actual installation.

andkononykhin2 (changes requested):

cli/cortx-prvsnr-cli.spec
```diff
@@ -54,27 +54,71 @@ rm -rf %{buildroot}
...
-  rm -f /root/.ssh/id_rsa_prvsnr.pub || true
+## Ensure update replaces the keys
+#if [[ -e /root/.ssh/id_rsa_prvsnr ]]; then
+#  /bin/rm -f /root/.ssh/id_rsa_prvsnr || true
+#  /bin/rm -f /root/.ssh/id_rsa_prvsnr.pub || true
+#fi
+
+
+if [[ ! -e /root/.ssh/ssh_config || ! -e /root/.ssh/authorized_keys 
+  || ! -e /root/.ssh/id_rsa_prvsnr.pub || ! -e /root/.ssh/id_rsa_prvsnr ]]; then
```
ypise, line 96: This won't work.
We are checking all or none. We should just check `-e /root/.ssh/id_rsa_prvsnr`
ssh_config should be tested separately for it's presense.
authorized_keys shouldn't be checked.

cli/cortx-prvsnr-cli.spec
ypise: @sradhanand-pati change this to:  
`Host srvnode-1 srvnode-1.localdomain`

cli/cortx-prvsnr-cli.spec
ypise: @sradhanand-pati change this to:
`Host srvnode-1 srvnode-1.localdomain`

cli/cortx-prvsnr-cli.spec
ypise: @sradhanand-pati change this to:
`Host srvnode-1 srvnode-1.localdomain`

cli/cortx-prvsnr-cli.spec
ypise: @sradhanand-pati change this to:
`Host srvnode-1 srvnode-1.localdomain`

ypise:
@sradhanand-pati 
I'm good with this PR. But, since I'm the creator of this PR, I'm unable to approve it.

@83bhp, @ipoddubnyy, @andkononykhin2 
Give it another look please. This scenario is now obsolete with the new Python API based bootstrap+deployment approach, but we still need to land it for the sake of cleaning up the repo of these files.

cli/cortx-prvsnr-cli.spec
```diff
@@ -54,27 +54,70 @@ rm -rf %{buildroot}
...
+## Ensure update replaces the keys
+#if [[ -e /root/.ssh/id_rsa_prvsnr ]]; then
+#  /bin/rm -f /root/.ssh/id_rsa_prvsnr || true
+#  /bin/rm -f /root/.ssh/id_rsa_prvsnr.pub || true
+#fi
+
+
+if [[ ! -e /root/.ssh/config ]]; then
+
+  /bin/ssh-keygen -o -q -t rsa -b 4096 -a 100 -N '' -f id_rsa_prvsnr
```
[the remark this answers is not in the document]
ypise: @andkononykhin2 
Yes @sradhanand-pati has managed to cover that in setup-provisioner script.
Since we do not plan to use this method to deploy in future, it is just a placeholder for now.

PR #540, opened 2020-09-29
EOS-13717: Round maxconn up instead of down in haproxy.cfg

t7ko-seagate:
Signed-off-by: Ivan Tishchenko <ivan.tishchenko@seagate.com>

This addresses EOS-13717.  Issue there is: current implementation before the fix will do "floor division", and so maxconn calculated will be smaller than needed.  We have to have total at least 150 connections per node, and with floor division we're getting less than that (nmbproc=12, 150 // 12 = 12, so total = nbproc * maxconn = 12*12 = 144, less than 150).

So I'm changing this to do round-up division, which will give expected results.

### DIFF
--- a/srv/components/ha/haproxy/files/haproxy.cfg
+++ b/srv/components/ha/haproxy/files/haproxy.cfg
@@ -8,7 +8,7 @@ global
 
     chroot      /var/lib/haproxy
     pidfile     /var/run/haproxy.pid
-    maxconn     {{ 150 // salt['pillar.get']('haproxy:nbproc', 2) }}
+    maxconn     {{ (150 / salt['pillar.get']('haproxy:nbproc', 2)) | round(0, 'ceil') | int }}
     user        haproxy             #Haproxy running under user and group "haproxy"
     group       haproxy
 {% if "physical" in grains['virtual'] %}


srv/components/ha/haproxy/files/haproxy.cfg
```diff
@@ -8,7 +8,7 @@ global
 
     chroot      /var/lib/haproxy
     pidfile     /var/run/haproxy.pid
-    maxconn     {{ 150 // salt['pillar.get']('haproxy:nbproc', 2) }}
+    maxconn     {{ int(0.999 + 150.0 / salt['pillar.get']('haproxy:nbproc', 2)) }}
```
83bhp, line 11: This won't work - {{ int() }}. I tried with this change and jinja failed to render with following error:
```
----------
          ID: Setup HAProxy config
    Function: file.managed
        Name: /tmp/haproxy.cfg
      Result: False
     Comment: Unable to manage file: Jinja variable 'int' is undefined
     Started: 15:54:44.946491
    Duration: 187.445 ms
     Changes:   
```
Need to have different logic to round up to the desired value.

andkononykhin2: I guess [|int](https://jinja.palletsprojects.com/en/2.11.x/templates/#int) filter might be used or a temp variable using `{% set ... %}` ...

sradhanand-pati: yes. for jinja you cant parse that way..
@t7ko-seagate  you need to have something like `{{ (0.999 + 150.0 / salt['pillar.get']('haproxy:nbproc', 2))| int }}`

t7ko-seagate: Fixed.

andkononykhin2 (approved):

ypise (approved):

t7ko-seagate:
Fix has been tested on sm26-r19 with help of @83bhp .

PR #541, opened 2020-09-29
Add to build and source metadata to api rpm package

shailesh-vaidya:
Updated package name as `cortx-prvsnr-api` and version as `1.0.0` to make api package name inline with  cortx-prvsnr and cortx-prvsnr-cli packages. 

During build time we need to execute, 

`bash -x ./devops/rpms/api/build_python_api.sh -vv --out-dir /root/rpmbuild/RPMS/x86_64/ --pkg-ver ${BUILD_NUMBER}_git$(git rev-parse --short HEAD`

Package will be generated as,
`python36-cortx-prvsnr-api-1.0.0-10_git5712d544.x86_64.rpm`

Please review.

### DIFF
--- a/api/python/provisioner/__metadata__.py
+++ b/api/python/provisioner/__metadata__.py
@@ -15,8 +15,8 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-__title__ = 'cortx-prvsnr'
-__version__ = '0.36.0'
+__title__ = 'cortx-prvsnr-api'
+__version__ = '1.0.0'
 __author__ = "Seagate"
 __author_email__ = 'support@seagate.com'  # TODO
 __maintainer__ = 'Seagate'
--- a/api/python/provisioner/__main__.py
+++ b/api/python/provisioner/__main__.py
@@ -39,6 +39,9 @@ from . import (
     log
 )
 from .commands import commands
+from .commands.setup_provisioner import (
+    SetupCmdBase, RunArgsSetupProvisionerGeneric
+)
 from .base import prvsnr_config
 
 from . import cli_parser
@@ -103,7 +106,7 @@ def _generate_logfile_filename(cmd):
     return (config.LOG_ROOT_DIR / f'{cmd}.{ts}.{pid}.{threadid}.log')
 
 
-def _set_logging(output_type, log_args=None):
+def _set_logging(output_type, log_args=None, other_args=None):
     if log_args is None:
         log_args = log.LogArgs()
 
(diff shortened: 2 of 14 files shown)


api/python/provisioner/__metadata__.py
```diff
@@ -15,8 +15,8 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-__title__ = 'cortx-prvsnr'
-__version__ = '0.36.0'
```
andkononykhin2, line 19: That versioning has some meaning. Setting 1.0.0 is not ok currently. Please check https://semver.org/#faq. 
If we decide to set 1.0.0 it should be a common decision and related to some milestone. E.g. GA-1.

Currently it looks like a quick patch. Why do you need that? If you want some specific relation with other packages please may consider to use snapshots in Release: tag. That doc is a quite good set of rules https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_snapshots

api/python/provisioner/__metadata__.py
```diff
@@ -15,8 +15,8 @@
...
+__title__ = 'cortx-prvsnr-api'
```
andkononykhin2, line 18: It will just break our setup logic.

andkononykhin2 (changes requested):
Hello Shailesh. I understand the change you want to submit but it's not a good time for that. The task EOS-13766 doesn't require that. All wr need is to add a build number as a revision version to unblock good comparison for upstream source code. Sha1 will help to rekate builds with upstream. So it is only about the build not the code.

shailesh-vaidya:
> Hello Shailesh. I understand the change you want to submit but it's not a good time for that. The task EOS-13766 doesn't require that. All wr need is to add a build number as a revision version to unblock good comparison for upstream source code. Sha1 will help to relate builds with upstream. So it is only about the build not the code.

Hi @andkononykhin2  - Regarding using 1.0.0 it was in accordance with [Naming specifications](https://seagatetechnology.sharepoint.com/:p:/r/sites/gteamdrv1/tdrive1224/_layouts/15/Doc.aspx?sourcedoc=%7B8E177E3D-B776-484A-8E8F-6263FCDC7830%7D&file=EOS%20Specifications%20-%20EESV1.pptx&nav=eyJzSWQiOjI2NSwiY0lkIjowfQ&action=edit&mobileredirect=true) from architecture team. I feel we should have all CORTX package names and versions inline. 

For EOS-13766 - I have updated pipelines to use 
` bash ./devops/rpms/api/build_python_api.sh -vv --out-dir /root/rpmbuild/RPMS/x86_64/ --pkg-ver ${BUILD_NUMBER}_git$(git rev-parse --short HEAD)`

Hence package generated would be `python36-cortx-prvsnr-0.36.0-42_git13f4725.x86_64.rpm`  

Hope this is fine.

andkononykhin2:
> > Hello Shailesh. I understand the change you want to submit but it's not a good time for that. The task EOS-13766 doesn't require that. All wr need is to add a build number as a revision version to unblock good comparison for upstream source code. Sha1 will help to relate builds with upstream. So it is only about the build not the code.
> 
> Hi @andkononykhin2 - Regarding using 1.0.0 it was in accordance with [Naming specifications](https://seagatetechnology.sharepoint.com/:p:/r/sites/gteamdrv1/tdrive1224/_layouts/15/Doc.aspx?sourcedoc=%7B8E177E3D-B776-484A-8E8F-6263FCDC7830%7D&file=EOS%20Specifications%20-%20EESV1.pptx&nav=eyJzSWQiOjI2NSwiY0lkIjowfQ&action=edit&mobileredirect=true) from architecture team. I feel we should have all CORTX package names and versions inline.

It might be. But let's park that for now since this is an API by the nature and it's critical to follow clear versioning.  always 1.0.0 breaks some other (widely accepted and followed) naming conventions.

> For EOS-13766 - I have updated pipelines to use
> ` bash ./devops/rpms/api/build_python_api.sh -vv --out-dir /root/rpmbuild/RPMS/x86_64/ --pkg-ver ${BUILD_NUMBER}_git$(git rev-parse --short HEAD)`
> 
> Hence package generated will be `python36-cortx-prvsnr-0.36.0-42_git13f4725.x86_64.rpm`

Thanks, this is exactly what we need.

PR #542, opened 2020-09-29
EOS-13772: PATH updated for admin user access

divya-vijay:

### DIFF
--- a/api/python/provisioner/commands/__init__.py
+++ b/api/python/provisioner/commands/__init__.py
@@ -1163,6 +1163,30 @@ class CreateUser(CommandParserFillerMixin):
             _generate_ssh_config()
             _copy_minion_nodes()
 
+        # Update PATH just as a precaution
+        StateFunExecuter.execute(
+            'file.blockreplace',
+            fun_kwargs=dict(
+                name='~/.bashrc',
+                marker_start='# DO NOT EDIT: Start',
+                marker_end='# DO NOT EDIT: End',
+                content='export PATH=/usr/local/bin:$PATH',
+                append_if_not_found=True,
+                append_newline=True,
+                backup=False
+            ),
+            targets=targets
+        )
+
+        StateFunExecuter.execute(
+            'cmd.run',
+            fun_kwargs=dict(
+                name='source ~/.bashrc'
+            ),
+            targets=targets
+        )
+
+
         # Creating a new group with limited access for csm admin users
         StateFunExecuter.execute(
             'group.present',
@@ -1194,9 +1218,12 @@ class CreateUser(CommandParserFillerMixin):
                           '/usr/sbin/subscription-manager, /usr/bin/cat, '
                           '/usr/bin/cd, /usr/bin/ls, '
                           f'/usr/sbin/pcs, {SEAGATE_USER_HOME_DIR}, '
-                          '/usr/bin/salt, /usr/bin/systemctl '
-                          '/usr/bin/yum, /usr/bin/dir, /usr/bin/cp '
-                          '/opt/seagate/cortx/csm/bin/cortxcli, '
+                          '/usr/bin/salt, /usr/local/bin/salt, '
+                          '/usr/bin/salt-call, /usr/local/bin/salt-call, '
+                          '/usr/bin/yum, /usr/bin/dir, /usr/bin/cp, '
+                          '/opt/seagate/cortx/csm/lib/cortxcli, '
+                          '/usr/bin/cortxcli, /var/log, '
+                          '/usr/bin/systemctl, '
                           f'{PRVSNR_CLI_DIR}/factory_ops/boxing/init,'
                           f'{PRVSNR_CLI_DIR}/factory_ops/unboxing/init'),
                 create=True,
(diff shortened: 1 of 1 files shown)


api/python/provisioner/commands/__init__.py
```diff
@@ -1163,6 +1163,30 @@ def _passwordless_ssh():
...
             _copy_minion_nodes()
 
+        # Update PATH just as a precaution
+        StateFunExecuter.execute(
+            'file.blockreplace',
+            fun_kwargs=dict(
+                name='~/.bashrc',
+                marker_start='# DO NOT EDIT: Start',
+                marker_end='# DO NOT EDIT: End',
+                content='export PATH=/usr/local/bin:$PATH',
```
ipoddubnyy, line 1173: place `/usr/local/bin` after $PATH (e.g., `export PATH=$PATH:/usr/local/bin`)

ipoddubnyy (changes requested):
minor change requested

api/python/provisioner/commands/__init__.py
divya-vijay: yes, updated now!

ipoddubnyy (approved):
approved

api/python/provisioner/commands/__init__.py
```diff
@@ -1163,6 +1163,30 @@ def _passwordless_ssh():
...
             _copy_minion_nodes()
 
+        # Update PATH just as a precaution
+        StateFunExecuter.execute(
+            'file.blockreplace',
+            fun_kwargs=dict(
+                name='~/.bashrc',
+                marker_start='# DO NOT EDIT: Start',
+                marker_end='# DO NOT EDIT: End',
+                content='export PATH=$PATH:/usr/local/bin',
```
andkononykhin2, line 1173: @ipoddubnyy @divya-vijay I think we shold mostly rely on `/usr/bin` in production. So makes sense to prepend (not append) both `/usr/bin` should be the first

api/python/provisioner/commands/__init__.py
```diff
@@ -1163,6 +1163,30 @@ def _passwordless_ssh():
             _generate_ssh_config()
             _copy_minion_nodes()
 
+        # Update PATH just as a precaution
+        StateFunExecuter.execute(
+            'file.blockreplace',
+            fun_kwargs=dict(
+                name='~/.bashrc',
```
andkononykhin2, line 1170: It will work not for only shells since it might not be read by shell in some cases. @divya-vijay please create a task to address that better later. For now extended list of sudo commands should anyway fix the issue.

andkononykhin2 (approved):

api/python/provisioner/commands/__init__.py
```diff
@@ -1194,9 +1218,12 @@ def _passwordless_ssh():
...
                           '/usr/bin/cd, /usr/bin/ls, '
                           f'/usr/sbin/pcs, {SEAGATE_USER_HOME_DIR}, '
-                          '/usr/bin/salt, /usr/bin/systemctl '
-                          '/usr/bin/yum, /usr/bin/dir, /usr/bin/cp '
-                          '/opt/seagate/cortx/csm/bin/cortxcli, '
+                          '/usr/bin/salt, /usr/local/bin/salt, '
+                          '/usr/bin/salt-call, /usr/local/bin/salt-call, '
+                          '/usr/bin/yum, /usr/bin/dir, /usr/bin/cp, '
+                          '/opt/seagate/cortx/csm/lib/cortxcli, '
+                          '/usr/bin/cortxcli, /var/log, '
```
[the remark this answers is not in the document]
divya-vijay: just like /users path, I gave /var/log, just in case the user has access issues while viewing logs

api/python/provisioner/commands/__init__.py
divya-vijay: @andkononykhin yeah, but /usr/bin is, by default, present in all PATH variables. so I did not explicitly mention it. but in cases where /usr/local/bin is also present, precedence goes to local/. 
you suggest to add it this way -->  "/usr/bin:/usr/local/bin:$PATH " ?

api/python/provisioner/commands/__init__.py
```diff
@@ -1194,9 +1218,12 @@ def _passwordless_ssh():
                           '/usr/sbin/subscription-manager, /usr/bin/cat, '
                           '/usr/bin/cd, /usr/bin/ls, '
                           f'/usr/sbin/pcs, {SEAGATE_USER_HOME_DIR}, '
-                          '/usr/bin/salt, /usr/bin/systemctl '
-                          '/usr/bin/yum, /usr/bin/dir, /usr/bin/cp '
-                          '/opt/seagate/cortx/csm/bin/cortxcli, '
+                          '/usr/bin/salt, /usr/local/bin/salt, '
+                          '/usr/bin/salt-call, /usr/local/bin/salt-call, '
+                          '/usr/bin/yum, /usr/bin/dir, /usr/bin/cp, '
```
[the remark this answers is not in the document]
divya-vijay: Yes, I tried that first. it looked literally too many lines. I'll still handle it in the next PR to pre-cortx.

PR #546, opened 2020-09-30
[EOS-13737] - Set logging level to WARNING for get commands

dmitrydb:
Set logging level to WARNING for `get_params` command.


ypise (approved):

ypise (approved):

api/python/provisioner/config.py
```diff
@@ -184,6 +184,14 @@
     'replace_node'
 ]
 
+
+# List of commands for which we purposely reduce log level severity to suppress
+# logging noise and prevent log files size growing.
+LOG_SUPPRESSION_CMDS = [
```
andkononykhin2, line 190: Please add here all get commands that we  have currently

api/python/provisioner/__main__.py
```diff
@@ -132,6 +132,11 @@ def _set_logging(output_type, log_args=None, other_args=None):
         setattr(log_args, config.LOG_RSYSLOG_HANDLER, False)
         log_args.update_handlers()
 
+    if cmd in config.LOG_SUPPRESSION_CMDS:
+        setattr(log_args, "rsyslog_level", logging.WARNING)
```
andkononykhin2, line 136: please use here constants instead of hard coded rsyslog, logfile or console

api/python/provisioner/__main__.py
```diff
@@ -132,6 +132,11 @@ def _set_logging(output_type, log_args=None, other_args=None):
...
+        setattr(log_args, "logfile_level", logging.WARNING)
```
andkononykhin2, line 137: I believe we don not need that:
- logfile by default is not turned on for get command
- in case of custom logfile settings we should allow logs to custom path

andkononykhin2 (changes requested):

api/python/provisioner/config.py
dmitrydb: Done

api/python/provisioner/__main__.py
dmitrydb: Done

api/python/provisioner/__main__.py
dmitrydb: Left suppression only for **rsyslog** type of logging.

api/python/provisioner/config.py
```diff
@@ -184,9 +184,36 @@
...
+    'get_params',
+    "get_result",
+    "grains_get",
+    "pillar_get",
+    "get_params",
+    "get_cluster_id",
+    "get_node_id",
+    "get_release_version",
+    "get_release_version",
```
andkononykhin2, line 199: please remove duplicates

api/python/provisioner/__main__.py
```diff
@@ -132,6 +132,9 @@ def _set_logging(output_type, log_args=None, other_args=None):
         setattr(log_args, config.LOG_RSYSLOG_HANDLER, False)
         log_args.update_handlers()
 
+    if cmd in config.LOG_SUPPRESSION_CMDS:
+        setattr(log_args, config.LogLevelTypes.RSYSLOG.value, logging.WARNING)
```
andkononykhin2, line 136: please add a conditional `if hasattr(log_args, config.LOG_RSYSLOG_HANDLER):` for cases when rsyslog is not listed in handlers

api/python/provisioner/__main__.py
```diff
@@ -169,7 +172,10 @@ def _set_logging(output_type, log_args=None, other_args=None):
                 else:
                     filename = _generate_logfile_filename(log_args.cmd)
                 setattr(log_args, filename_attr, str(filename))
-            log_args.update_handlers()
+
+    if (cmd in config.LOG_SUPPRESSION_CMDS or
```
andkononykhin2, line 176: what if we can accumulate all changes in some key-value manner and apply them all here using `setattr` so we won't need to check these conditionals one more time?

andkononykhin2 (changes requested):
@dmitrydb please address few comments

api/python/provisioner/__main__.py
dmitrydb: Refactoring this. But python needs to support `finally` for if-else conditional :slightly_smiling_face:

api/python/provisioner/__main__.py
dmitrydb: Done.
But as improvement we need to consider a way to combine `hasattr` and `setattr` method in single call. For example, `attr` module support `on_setattr` callback
https://www.attrs.org/en/stable/api.html

api/python/provisioner/config.py
dmitrydb: Done

api/python/provisioner/__main__.py
andkononykhin2: possibly... but I meant that we can:
```
log_changes = {}
if(cond1):
   if (conf2):
      log_changes[attr1] = val1

...

if (cond3):
    log_changes[attr2] =- val2

if log_changes:
    for lattr, lval in ...:
        setatt(log_args, lattr, lval)
    log_args.update_handlers()  
```

api/python/provisioner/__main__.py
andkononykhin2: sure, you can do the poc here

andkononykhin2 (approved):

PR #553, opened 2020-09-30
EOS-13644: BMC Watchdog implementation

divya-vijay:
Tested on VM for Salt flow. Once there is a HW available, can check for final ipmi command, as well.


ipoddubnyy (approved):
approved

srv/components/misc_pkgs/ipmi/bmc_watchdog/install.sls
```diff
@@ -0,0 +1,39 @@
...
+
+Update watchdog timeout:
+  file.replace:
+    - name: /etc/sysconfig/bmc-watchdog
+    - pattern: '900'
+    - repl: '300'
+
+Reload kernel with Modprobe:
+  file.managed:
+    - name: /etc/modules
```
ypise, line 30: I don't have a good feeling about this file: `/etc/modules`
Change this to: `/etc/modules-load.d/ipmi_mods.conf`
Ref: https://unix.stackexchange.com/a/189696

srv/components/misc_pkgs/ipmi/bmc_watchdog/start.sls
```diff
@@ -0,0 +1,27 @@
...
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+Start and enable bmc-watchdog:
+  service.running:
+    - name: bmc-watchdog
+    - enable: True
+
+Verify bmc-watchdog:
```
ypise, line 23: This should go into a separate `sanity_check.sls`.

ypise (changes requested):

srv/components/misc_pkgs/ipmi/bmc_watchdog/start.sls
divya-vijay: That's how I'd added it in previous commit, but I have placed a require condition in start file. only when it is enabled, we can check the command. So, added it to start.sls directly.

srv/components/misc_pkgs/ipmi/bmc_watchdog/install.sls
```diff
@@ -0,0 +1,39 @@
...
+#
+
+Install bmc-watchdog:
+  pkg.installed:
+    - name: freeipmi-bmc-watchdog
+
+Update watchdog timeout:
+  file.replace:
+    - name: /etc/sysconfig/bmc-watchdog
+    - pattern: '900'
```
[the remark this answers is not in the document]
divya-vijay: @ypise The entire file has only one line explaining different options. Since the optional parameter placing may vary, I'd checked the pattern directly for sleep time.
Ex: OPTIONS="-d -u 4 -p 0 -a 1 -F -P -L -S -O -i 300 -e 60"

srv/components/misc_pkgs/ipmi/bmc_watchdog/install.sls
divya-vijay: This is an optional step to reload the kernel before ipmi command can be run. And it accepts the options on /modules file and doesnt not check in any .conf files. So, I created one new and added. I'll test this on HW anyway and see if it works on .conf.

srv/components/misc_pkgs/ipmi/bmc_watchdog/install.sls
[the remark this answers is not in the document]
ypise: Can we not replace `-i 900` to `-i 300`?

srv/components/misc_pkgs/ipmi/bmc_watchdog/start.sls
ypise: No it has to be in sanity_check.sls.
you can use `include` to set the dependency on start.sls from sanity.sls

srv/components/misc_pkgs/ipmi/bmc_watchdog/install.sls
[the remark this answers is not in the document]
divya-vijay: Done! Updated all the comments in the new commit.

PR #555, opened 2020-10-01
EOS-13764 : deploy stack on vm using python API

mazinamdar:

single node and dual node vm deployment


api/python/provisioner/commands/deploy_vm.py
```diff
@@ -93,6 +93,38 @@ class DeployVM(Deploy):
     input_type: Type[inputs.NoParams] = inputs.NoParams
     _run_args_type = run_args_type
 
+    def set_pillar_data(self, setup_type):
+        minions_ids = ['srvnode-1']
```
andkononykhin2, line 97: please add some newlines to improve readability

andkononykhin2 (changes requested):
@mazinamdar Looks good. But some changes are required:
- please check [that](https://github.com/Seagate/cortx-prvsnr/blob/1a62803c14d07f316de601f22155f140689734c1/api/python/provisioner/commands/deploy_vm.py#L135-L136) part and update how it is implemented in deploy_dual.py
- what about auto_deploy_vm command which should orchestrate that from bootstrap state in the same way as auto-deploy-vm bash script

Also please double-check bash deploy-vm and this one are in sync.

api/python/provisioner/commands/deploy_vm.py
mazinamdar: comment is addressed

ypise (approved):

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -100,14 +146,14 @@ def _run_states(self, states_group: str, run_args: run_args_type):
         states = deploy_states[states_group]
         stages = run_args.stages
 
-        primary = local_minion_id()
+        primary = self._primary_id()
```
andkononykhin2, line 149: that should be replaced everywhere, not only here
also:
- all direct calls to `provisioner.salt` API should be replaced with calls to parent (`Deploy`) class:
    - cmd_run -> self._cmd_run
    - local_minion_id() -> self._primary_id()
    - function_run -> self._function_run

The only missed thing in Deploy is StateFunExecuter. You can either:
- extend Deploy to provide that api to child classes
- use `self._function_run` since actually StateFunExecuter is just a wrepper over provisioner.salt.function_fun for `state.single` function

The general idea is support 2 cases of setup:
- remote setup so it can be combined with SetupProvisioner
- local setup so it might be run on a primary node locally

Deploy as a parent class takes care about that fro children classes.

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -194,25 +241,29 @@ def run(self, **kwargs):
...
                     logger.info(
                         f"Mounting partition {metadata_device} "
-                        "into {mount_point} (with fstab record)"
+                        f"into {mount_point} (with fstab record)"
                     )
                     StateFunExecuter.execute(
                         'mount.mounted',
-                        fun_args=[mount_point, metadata_device, 'ext4'],
-                        fun_kwargs=dict(mkmnt=True, persist=True)
+                        fun_kwargs=dict(
```
andkononykhin2, line 255: please check my comment above

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -92,6 +94,50 @@
...
+                )
+                self._cmd_run(
+                    "provisioner pillar_set "
+                    f" cluster/{minion_id}/storage/data_devices "
+                    f'[\\"{disk[-1]}\\"]',
+                    targets=self._primary_id()
+                )
+
+            self._cmd_run(
+                "provisioner pillar_set "
```
andkononykhin2, line 129: two comments:
- looks complicated, better to use single quotes for all parts to avoid any side effects
- please use additional parenthesis:
```
                self._cmd_run(
                    (
                        'provisioner pillar_set '
                        f' cluster/{minion_id}/storage/metadata_device '
                        f'[\\"{disk[-2]}\\"]'
                    ),
                    targets=self._primary_id()
                )
```

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -92,6 +94,50 @@
...
+                f"cluster/{minion_id}/network/data_nw/roaming_ip  "
+                '\\"127.0.0.1\\"',
+                targets=self._primary_id()
+            )
+
+        self._cmd_run(
+            "provisioner pillar_set "
```
andkononykhin2, line 136: the same as above

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -92,6 +94,50 @@
...
+            "lsblk -ndp | grep disk | awk '{ print $1 }'",
+            targets=self._primary_id()
+        )
+        if res[self._primary_id()]:
+            disk = res[self._primary_id()].split("\n")
+
+        for minion_id in minions_ids:
+            if len(disk) > 1:
+                self._cmd_run(
+                    "provisioner pillar_set "
```
andkononykhin2, line 116: please check my comment below

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -92,6 +94,50 @@
...
+                    f" cluster/{minion_id}/storage/metadata_device "
+                    f'[\\"{disk[-2]}\\"]',
+                    targets=self._primary_id()
+                )
+                self._cmd_run(
+                    "provisioner pillar_set "
```
andkononykhin2, line 122: please check my comment below

andkononykhin2 (changes requested):
@mazinamdar Please address the comments. Thank you.

api/python/provisioner/commands/deploy_vm.py
mazinamdar: replaced all local function call with Deploy func calls.

api/python/provisioner/commands/deploy_vm.py
mazinamdar: comment is addressed

api/python/provisioner/commands/deploy_vm.py
mazinamdar: Done

api/python/provisioner/commands/deploy_vm.py
mazinamdar: Done

api/python/provisioner/commands/deploy_vm.py
mazinamdar: comment is addressed

api/python/provisioner/commands/deploy_vm.py
mazinamdar: comment is addressed

api/python/provisioner/commands/deploy_vm.py
andkononykhin2: that won't work for remote setup, try the following instead:
```
               fun_kwargs=dict(...)

               # TODO IMPROVE move that conditional split into Deploy class
               if self.setup_ctx:
                   return self.setup_ctx.ssh_client.run(
                        'state.single',
                        fun_args=['mount.mounted'],
                        fun_kwargs=fun_kwargs,
                    )
               else:
                 StateFunExecuter.execute(
                        'mount.mounted',
                        fun_kwargs=fun_kwargs
                )   
```

api/python/provisioner/commands/deploy_vm.py
andkononykhin2: please check one last place

api/python/provisioner/commands/deploy_vm.py
mazinamdar: comment is addressed

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -252,16 +252,25 @@ def run(self, **kwargs):
...
+                                   persist=True)
+                    if self.setup_ctx:
+                        self.setup_ctx.ssh_client.run(
+                            'state.single',
+                            fun_args=['mount.mounted'],
+                            fun_kwargs=fun_kwargs,
+                            targets=self._primary_id()
+                        )
+                    else:
+                        function_run(
```
andkononykhin2, line 269: there is no such function in `mount` salt's module, `mount.mounted` is a state. Please double check the example I posted before.

mazinamdar: comment is addressed

andkononykhin2 (approved):

PR #567, opened 2020-10-01
Revert "[WIP][EOS-12334]-ISO files with updates should be located outside installation directory"

dmitrydb:
Reverts Seagate/cortx-prvsnr#556. After discussion with @andkononykhin2 we decided to revert this PR for the additional testing


andkononykhin2 (approved):
We need to address the following:
- make accordant changes for paths in states
- implement a migration logic for the case of sw update when some repos are already installed

In the current state the change will likely break the sw update logic.

ipoddubnyy (approved):
approved

PR #572, opened 2020-10-02
[EOS-13853] adds iso support for on field setup

andkononykhin2:


ypise (approved):
Tested on system to work.

PR #574, opened 2020-10-03
EOS 13626: Rotate rabbitmq-server's log once it crosses specified size in the config file

VijayThakkar132:


srv/components/misc_pkgs/rabbitmq/files/rabbitmq-server
```diff
@@ -1,5 +1,5 @@
 /var/log/rabbitmq/*.log {
-    daily
+    hourly
```
andkononykhin2, line 2: @VijayThakkar132 am I understanding that correctly: each log file for rabbitmq will survive only for 7 hours. So in case of issues we would have not more than 7 hours to collect the logs. If that is true then don't you think it's an enough value.
@SomwanshiAnjali  @sradhanand-pati could you verify please?

SomwanshiAnjali: @andkononykhin2  @VijayThakkar132  
with these changes rabbitmq log file will persist only for 7 hours, I believe we should keep it at least for 2 days

VijayThakkar132: As far as, I know rotation only happens once it cross the specified size (In case of rabbitmq it's 20MB). So, It persists minimum for 7 hours in case logs size are more than 20 MB per hour. I think that case is very rare and  happens only due to major issue. @indrajitzagade  Any comment on this ?

indrajitzagade: @andkononykhin2 With this change we are checking more frequently that if we need a log rotation, it does not mean always it will be rotated every hour. So we allow 7 files of 20MB each to be accumulated, if logs are not flooded then the existing log files could stay for longer.

saurabhkhanvilkar: @andkononykhin2, please confirm if these changes look good, or if you have any more comments. If all ok, then please approve so we can merge. Thanks.

ypise: @saurabhkhanvilkar 
These changes are partial.
Logrotate by default doesn't support hourly logs:

1. https://stackoverflow.com/questions/25485047/hourly-rotation-of-files-using-logrotate/25485313
1. https://stackoverflow.com/a/56449479

@SomwanshiAnjali  could you help get this to work hourly?
@saurabhkhanvilkar we would need a separate ticket for the Provisioner task.

My only worry here is if this would trigger all log rotations on hourly basis.

ypise (changes requested):

srv/components/misc_pkgs/rabbitmq/files/rabbitmq-server
saurabhkhanvilkar: @ypise, as i understand there is this drive to change log rotation check frequency (all components) from daily to hourly... theory is that daily check could be too relaxed, as logs could over-shoot allotted size within a day and hence hourly check is required.

saurabhkhanvilkar: discussed with @ypise and concluded that this will need some work on provisioner side. the change in this PR is very trivial and won't work as-is. instead of creating a separate ticket for provisioner, this same ticket will be transferred to provisioner team and this current PR will be cancelled.

VijayThakkar132: Closing this PR and created [EOS-14206](https://jts.seagate.com/browse/EOS-14206) for the provisioner.

ypise (approved):

srv/components/misc_pkgs/rabbitmq/files/rabbitmq-server
andkononykhin2: > As far as, I know rotation only happens once it cross the specified size (In case of rabbitmq it's 20MB). So, It persists minimum for 7 hours in case logs size are more than 20 MB per hour. I think that case is very rare and happens only due to major issue. @indrajitzagade Any comment on this ?

@VijayThakkar132  yes, looks like so. From [logrotate man](https://linux.die.net/man/8/logrotate):
>size option is similar except that it is mutually exclusive with the time interval options, and it causes log files to be rotated without regard for the last rotation time.

and

>size size
Log files are rotated only if they grow bigger then size bytes.

But as I understand having that `size` we can't actually predict the time. We just know that it will keep up to 7*20MB logs. 

I believe makes sense:
- remove `hourly` since logrotate will just ignore that statement on behalf of `size`. Actually periodical check will be configured by the change that @ypise mentioned (#622)
- rename the title of the PR since it is about size based rotation

VijayThakkar132: @andkononykhin2 Agree with you. Addressed comments.

PR #577, opened 2020-10-04
EOS-13072: Disable firewalld during bootstrap

ypise:
---------------------------------------------------------------
salt-ssh being used during bootstrap fails when network restarts
as a follow-up action for firewall config.

### DIFF
--- a/api/python/provisioner/srv/salt/firewall/init.sls
+++ b/api/python/provisioner/srv/salt/firewall/init.sls
@@ -17,6 +17,7 @@
 
 include:
   - .prepare
-  - .install
-  - .config
-  - .start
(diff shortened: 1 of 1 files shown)


ypise:
@andkononykhin2 
Previous commits for EOS-13072 causes auto-deploy or any provisioner bootstrap commands to be stuck.
Bootstrap is performed using salt-ssh. When firewall is configured on management network, it restarts the network. This causes salt-ssh to lose connectivity and breaks the process.

Since the first related commit was landed directly to cortx-1.0, I'm overriding this PR as a special case to avoid any builds issues in cortx-1.0.

PR #579, opened 2020-10-05
EOS-13072: Firewall to be disabled during bootstrap

ypise:


ypise:
Addressed a leftover file in: #577

PR #580, opened 2020-10-05
EOS-13897: GlusterFS failure on Isolated system

ypise:


api/python/provisioner/srv/salt/glusterfs/cluster/config.sls
```diff
@@ -18,7 +18,7 @@
 glusterfs_servers_peered:
   glusterfs.peered:
     - names:
-{% for peer in salt['pillar.get']('glusterfs_peers', []) %}
+{% for peer in (salt['pillar.get']('glusterfs_peers', []) | difference(grains['fqdn'])) %}
```
andkononykhin2, line 21: It is assumed that the list is already prepared enough. So:
- bad to do same in multiple places and if you want to adjust it you need to check [here](https://github.com/Seagate/cortx-prvsnr/blob/37cdc7a745644fe8b67c4a13fc88758e86d33475/api/python/provisioner/commands/setup_provisioner.py#L1327)
- I don't think that it's a good idea to filter fqdns as they are enough stable and in other parts of the project we usually rely on them (not IP)

Why do you think it's necessary?

andkononykhin2 (changes requested):

api/python/provisioner/srv/salt/glusterfs/cluster/config.sls
ypise: Not necessary. It was the fix I found to work during testing.
I agree, filtering should be moved to source of the call.

However, it's a limitation in GlusterFS peering that it is not required on the node it is run from.
So a thought here might help.

andkononykhin2: @ypise I believe we should debug that more. Let's hold the PR for now.

ypise:
@andkononykhin2 
This fix is not necessary although, it would help if we avoid the primary node (current node ideally) in the list of pillars to be set.

Ref: https://docs.gluster.org/en/latest/Quick-Start-Guide/Quickstart/ 
**Step 5 - Configure the trusted pool**

PR #583, opened 2020-10-05
EOS-13904: Remove redunent config of STORAGE_ENCLOSURE from consul sspl config

mariyappanp:


pillar/components/sspl.sls
```diff
@@ -96,12 +96,6 @@ sspl:
     threaded: "true"
     logging_interval: "3600"
   STORAGE_ENCLOSURE:
-    primary_controller_ip: "127.0.0.1"
```
cdeshmukh, line 99: Am ok with this. But is this all? are we still referring from sspl.sls here, i remember we reduced use from sspl.sls earlier, but seems entries are lying around. Or sspl using rest of the stuff here still ?

cdeshmukh (commented):
Need this patch in cortx-1.0 branch first & then later can target for main branch

pillar/components/sspl.sls
ypise: Ideally we should have only these entries in this file:
```
sspl:
  health_map_path: /opt/seagate/cortx/provisioner/generated_configs/healthmap/
  health_map_file: ees-schema.json
  rmq:
    user: rmq
    secret: "B4kf1T6Qso"
```

cdeshmukh: Agree, it should be very minimal & i remember we shifted sspl config already, but stale entries remained here seems. @mariyappanp please confirm (check with @shubhamseagate too) and address

mariyappanp: @shubhamseagate patch is not having this change. There STORAGE_ENCLOSURE configs are still remained.

mariyappanp: Testing the same for cortx-1.0 on VM. I will post it once test done. BTW, I will test once in HW setup too.

mariyappanp: PR for cortx:
**pre-cortx-1.0**
https://github.com/Seagate/cortx-prvsnr/pull/591

thavanathan (approved):

pillar/components/sspl.sls
cdeshmukh: Ok you mean @shubhamseagate  patch taking care of the rest ? If yes lets move ahead with this change

mariyappanp: Yes

mariyappanp: Changed the ticket number as per the assignment. (EOS-13904 - **removing unused/duplicate kv pairs in consul**)

PR #585, opened 2020-10-05
EOS-13858: extending permissions to csm admin user

divya-vijay:


divya-vijay:
@andkononykhin @ypise I feel we can improve this csm-user permissions handling with states and calling it in CreateUser API. 
We may even get more requirements on this. I would like your views on this, please.

api/python/provisioner/commands/__init__.py
```diff
@@ -1207,6 +1207,15 @@ def _passwordless_ssh():
             targets=targets
         )
 
+        # Ensure rsync is installed, if not already
+        StateFunExecuter.execute(
```
ypise, line 1211: Common package installation should be covered in:
`cortx-prvsnr/srv/components/system/install.sls`
Which is already present. Although it's idempotent, it should not be repeated unless there's a good reason for doing so.

api/python/provisioner/commands/__init__.py
```diff
@@ -1236,6 +1245,13 @@ def _passwordless_ssh():
                           '/usr/bin/provisioner, '
                           '/var/log, '
                           '/tmp, '
+                          '/usr/bin/lsscsi, '
+                          '/usr/sbin/mdadm, '
+                          '/usr/sbin/sfdisk, '
+                          '/usr/sbin/mkfs, '
+                          '/usr/bin/rsync, '
+                          '/bin/rsync, '
+                          '/usr/local/bin/rsync, '
```
ypise, line 1254: `/usr/local/bin/rsync` not requried.
rsync installation is using yum. The binary would land in `/usr/bin` only.

ypise (changes requested):

ipoddubnyy (approved):
approved (providing that comment from Yash about /usr/local/bin/rsync is addressed)

api/python/provisioner/commands/__init__.py
andkononykhin2: I agree. Just for better understanding: all these issues with `/usr/local/bin` that we faced before were about pip vs yum ways of installation:
-  pip installed tool to /usr/local/bin
- yum installed to /usr/bin

For now it might be ok to make our life safer and cover both cases for salt but let's not abuse. We need a very strong concerns for any extensions here.

api/python/provisioner/commands/__init__.py
andkononykhin2: Agree. That command should base on assumption: user is created on field, it means the sw stack is fully deployed and all our necessary states are applied. Otherwise you would need to install here all packages that provides the tools listed for sudo.

andkononykhin2 (changes requested):

andkononykhin2 (approved):

ypise (approved):

PR #600, opened 2020-10-07
EOS-13330: lower stop-timeout for STONITH is too dangerous

knekrasov:
JIRA: https://jts.seagate.com/browse/EOS-13330

Setting smaller timeout for STONITH resources was not quite justified.
If (by any reason) STONITH resource can't be stopped within the given
timeout, Pacemaker will fence the node. This will definitely not make
failover/failback to happen faster.

Solution: revert #595


83bhp:
What is the default timeout value?

83bhp (commented):
Looks ok, but if this is removed what would be the default value for timeout?

andkononykhin2 (approved):

andkononykhin2:
> Looks ok, but if this is removed what would be the default value for timeout?

@83bhp sorry, hasn't noticed you question which is reasonable actually
I believe we may consider better change (explicit default timeout) as a different change. Current one is mostly to revert the previous change that might introduce a regression for hare configuration.

PR #338, opened 2020-09-08
WIP: Adding unit tests for inputs.py classmethods.

sudarshana27:
Signed-off-by: Sudarshana Babar <sudarshana.babar@seagate.com>

PR #604, opened 2020-10-08
EOS-12439: Implemented 8 unit tests for inputs.py classmethods.

sudarshana27:
Signed-off-by: Sudarshana Babar <sudarshana.babar@seagate.com>

### DIFF
--- a/test/api/python/provisioner/test_inputs.py
+++ b/test/api/python/provisioner/test_inputs.py
@@ -16,6 +16,7 @@
 #
 
 import pytest
+from unittest.mock import call
 import argparse
 from typing import Union, List
 from pathlib import Path
@@ -40,7 +41,8 @@ from provisioner.inputs import (
     ParamGroupInputBase,
     NTP, Network, NetworkParams,
     ParamDictItemInputBase,
-    SWUpdateRepo
+    SWUpdateRepo,
+    ParserFiller
 )
 from provisioner.pillar import PillarKey
 from provisioner.api_spec import param_spec as _param_spec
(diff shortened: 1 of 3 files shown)


test/api/python/provisioner/test_inputs.py
```diff
@@ -744,3 +746,152 @@ def _check(*args):
...
+
+def test_ParserFiller_fill_parser_with_no_metadata_argparser():
+    parser = argparse.ArgumentParser()
+    SC = attr.make_class("SC", {
+        "y": attr.ib(
+            default=123,
+            type=int
+        )
+    })
+    ParserFiller.fill_parser(SC, parser)
```
andkononykhin2, line 826: you don't do any checks here, what is the purpose of that test?

test/api/python/provisioner/test_inputs.py
```diff
@@ -744,3 +746,152 @@ def _check(*args):
...
+
+
+def test_ParserFiller_fill_parser_input_checks(mocker):
+    parser = argparse.ArgumentParser()
+    SC = attr.make_class("SC", {
+        "y": attr.ib(
+            default=123,
```
andkononykhin2, line 833: a boolean is expected here as a default value

test/api/python/provisioner/test_inputs.py
```diff
@@ -744,3 +746,152 @@ def _check(*args):
...
+    })
+    ret = ParserFiller.extract_positional_args(SC, attr.fields_dict(SC))
+    assert ret == ([attr.fields(SC).x], {})
+
+
+def test_ParserFiller_extract_positional_args_no_metadata_argparser():
+    SC = attr.make_class("SC", {
+        "x": attr.ib(
+            type=int,
+            repr=False,
```
andkononykhin2, line 876: why do we need `repr` here?

andkononykhin2 (commented):
@sudarshana27 Looks good. Please just address minor questions. Thank you.

test/api/python/provisioner/test_inputs.py
sudarshana27: It was a suggestion given in the old PR : https://github.com/Seagate/cortx-prvsnr/pull/338/commits/f039adba116fb689a81dd706a99cd9f810a1d45c
ParserFiller is not covered but as long as it just passes result of the AttrParserArgs to argparse we should focus on special cases:

- a class has some argument that doesn't specify METADATA_ARGPARSER - that argument should be ignored, so calls to parser.parse_args should produce an error

This unit test is implemented to cover above test scenario, but it doesn't throw any error.
Please suggest further, whether should we keep it or remove ?

test/api/python/provisioner/test_inputs.py
sudarshana27: have added some random attributes to the variable as test scenario requirement is not to add metadata attr.

test/api/python/provisioner/test_inputs.py
sudarshana27: Done

test/api/python/provisioner/test_inputs.py
andkononykhin2: >so calls to parser.parse_args should produce an error
but you don't do that here

sudarshana27: working on it

sudarshana27: When I am calling parser.parse_args(), it is throwing below error:
`>           args = parser.parse_args(['--y', 'some-value'])

test/api/python/provisioner/test_inputs.py:828:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/local/lib/python3.7/argparse.py:1758: in parse_args
    self.error(msg % ' '.join(argv))
/usr/local/lib/python3.7/argparse.py:2508: in error
    self.exit(2, _('%(prog)s: error: %(message)s\n') % args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = ArgumentParser(prog='pytest', usage=None, description=None, formatter_class=<class 'argparse.HelpFormatter'>, conflict_handler='error', add_help=True)
status = 2, message = 'pytest: error: unrecognized arguments: --y some-value\n'

    def exit(self, status=0, message=None):
        if message:
            self._print_message(message, _sys.stderr)
>       _sys.exit(status)
E       SystemExit: 2

/usr/local/lib/python3.7/argparse.py:2495: SystemExit
`
It is not raising any exception as such in order to catch within pytest.raises.
Should I mark this test with pytest  marker as @pytest.mark.xfail ?

andkononykhin2: @sudarshana27 yes, that is what we should expect, try the following:
```
    with pytest.raises(SystemExit) as excinfo:                                                                                                                               
       args = parser.parse_args(['--y', 'some-value'])
```
you may also check `unrecognized arguments: --y some-value'` in `excinfo`

sudarshana27: Addressed

andkononykhin2 (approved):

PR #606, opened 2020-10-08
[WIP] EOS-13921 : [SWUpdate] add restart minion logic at the end of sw_update

mazinamdar:

### DIFF
--- a/api/python/provisioner/commands/__init__.py
+++ b/api/python/provisioner/commands/__init__.py
@@ -540,6 +540,7 @@ class SWUpdate(CommandParserFillerMixin):
         #      options: set up temp ssh config and rollback yum + minion config
         #      via ssh as a fallback
         rollback_ctx = None
+        minion_conf_changes = None
         try:
             ensure_cluster_is_healthy()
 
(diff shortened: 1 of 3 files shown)


api/python/provisioner/commands/__init__.py
```diff
@@ -678,6 +679,14 @@ def run(self, targets):  # noqa: C901 FIXME
                 update_exc, rollback_error=rollback_error
             ) from update_exc
 
+        finally:
+            if minion_conf_changes:
+                logger.debug("Restarting salt minions ")
```
andkononykhin2, line 684: please use INFO level here

andkononykhin2: we shouldn't ever fail here actually, I want to change my mind : let's move that  ([here](https://github.com/Seagate/cortx-prvsnr/blob/52ef5b9528246b1d4308ea23da65b90e27de6abd/api/python/provisioner/commands/__init__.py#L598)) and use try-except block
```
try:
....
except Exception as exc:
           raise SWStackUpdateError(exc) from exc
```

andkononykhin2 (changes requested):

andkononykhin2 (approved):

ypise (approved):

PR #607, opened 2020-10-08
[EOS-14063]: Delete nic parameter for ClusterIP resource creation

dmitrydb:
Based on parameters description for IPaddr2 resource agent in source code:
  
> https://github.com/ClusterLabs/resource-agents/blob/e3c89b03fc8c696498b62b2c2ba807f78842a941/heartbeat/IPaddr2#L172-L181

and this answer in the official GitHub Issues page of the `pcs` tool:
 > https://github.com/ClusterLabs/pcs/issues/144

I decided to delete the `nic` parameter for the `pcs resource create ClusterIP` command in cluster_ip.sls

if omit `nic` parameter, it trigger the IPaddr2 resource agent itself to figure out what interface should be used based on the node routing table. To determine network interfaces the IPaddr2 uses script `/usr/lib/ocf/lib/heartbeat/findif.sh` and following logic of IPaddr2 script:
>  https://github.com/ClusterLabs/resource-agents/blob/e3c89b03fc8c696498b62b2c2ba807f78842a941/heartbeat/IPaddr2#L532-L554

### DIFF
--- a/srv/components/ha/corosync-pacemaker/config/cluster_ip.sls
+++ b/srv/components/ha/corosync-pacemaker/config/cluster_ip.sls
@@ -15,7 +15,7 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-{% if "primary" in grains['roles'] 
+{% if "primary" in grains['roles']
   and pillar['cluster']['cluster_ip']
   and "physical" in grains['virtual']
 %}
@@ -38,7 +38,7 @@ Update ClusterIP:
 
 Setup ClusterIP resouce:
   cmd.run:
-    - name: pcs resource create ClusterIP ocf:heartbeat:IPaddr2 ip={{ pillar['cluster']['cluster_ip'] }} mac=$(echo "01:00:5e:`echo {{ grains['hwaddr_interfaces'][data_if] }} | cut -d ":" -f 1-3`") nic={{ data_if }} op monitor interval=30s
+    - name: pcs resource create ClusterIP ocf:heartbeat:IPaddr2 ip={{ pillar['cluster']['cluster_ip'] }} mac=$(echo "01:00:5e:`echo {{ grains['hwaddr_interfaces'][data_if] }} | cut -d ":" -f 1-3`") op monitor interval=30s
 
 Add stickiness metadata to ClusterIP resource:
   cmd.run:


ypise:
@ajaysrivas, @knekrasov, @mssawant, @indrajitzagade 
Could you have a look at this change and help us ensure, if all is OK with it?
We have limited knowledge in this area and would like to have your opinion.

In the past the **_nic_** parameter was added to ensure stickiness. But it seems that is not really necessary.

Appreciate your help!

knekrasov:
I also have not so much of expertise here. 

According to what I see [here](http://www.linux-ha.org/doc/man-pages/re-ra-IPaddr2.html) for `nic` parameter:
> The base network interface on which the IP address will be brought online. **If left empty, the script will try and determine this from the routing table**. Do NOT specify an alias interface in the form eth0:1 or anything here; rather, specify the base interface only. Prerequisite: There must be at least one static IP address, which is not managed by the cluster, assigned to the network interface. If you can not assign any static IP address on the interface, modify this kernel parameter: sysctl -w net.ipv4.conf.all.promote_secondaries=1 (or per device) (optional, string, default eth0).

The only question I have is whether we are sure that the script will manage to determine the interface correctly. Are we sure about that?

And will it fail if it can't infer the right interface? Would be nice if yes..

srv/components/ha/corosync-pacemaker/config/cluster_ip.sls
```diff
@@ -38,7 +38,7 @@ Update ClusterIP:
 
 Setup ClusterIP resouce:
   cmd.run:
-    - name: pcs resource create ClusterIP ocf:heartbeat:IPaddr2 ip={{ pillar['cluster']['cluster_ip'] }} mac=$(echo "01:00:5e:`echo {{ grains['hwaddr_interfaces'][data_if] }} | cut -d ":" -f 1-3`") nic={{ data_if }} op monitor interval=30s
+    - name: pcs resource create ClusterIP ocf:heartbeat:IPaddr2 ip={{ pillar['cluster']['cluster_ip'] }} mac=$(echo "01:00:5e:`echo {{ grains['hwaddr_interfaces'][data_if] }} | cut -d ":" -f 1-3`") op monitor interval=30s
```
ajaysrivas, line 41: I guess that the ClusterIP needs to be bound with data nic. After the change, it may get resolved to other interface or mac value makes sure that it is resolved to data interface?

srv/components/ha/corosync-pacemaker/config/cluster_ip.sls
```diff
@@ -15,7 +15,7 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-{% if "primary" in grains['roles'] 
+{% if "primary" in grains['roles']
```
ypise, line 18: This needs to be `{% if pillar["cluster"][grains["id"]]["is_primary"] `

ypise (commented):
@dmitrydb 
It seems we don't have a choice here: https://serverfault.com/questions/818147/pacemaker-2-floatingip

I've marked the JIRA as impediment for now

dmitrydb:
Ticket was closed without that solution. Closing this PR.

PR #608, opened 2020-10-08
EOS-14103: Update cortx-prereqs.sh for centos7.8 support

SomwanshiAnjali:


cli/src/cortx-prereqs.sh
```diff
@@ -251,16 +251,21 @@ fi
 
 echo -n "INFO: Checking if kernel version is correct.........................." 2>&1 | tee -a ${LOG_FILE}
 
+grep -q "Red Hat" /etc/*-release && {
```
ipoddubnyy, line 254: does CentOS also return "Red Hat"?
and is it "Red Hat" or "RedHat"?

ipoddubnyy (approved):
approved (providing my question is answered)

cli/src/cortx-prereqs.sh
SomwanshiAnjali: 1. On centos above grep command will return nothing and hence kernel version will be set to 3.10.0-1127.el7.x86_64
2. In /etc/os-release and /etc/redhat-release the release name  is mentioned as "Red Hat Enterprise Linux Server"

83bhp:
The changes look ok to me. But as discussed in the morning please add following changes:
1. make -t argument compulsory for deployment on CentOS, if not provided error out and exit.
2. Use the OS repos from Seagate hosted mirrors. This will lock the versions for packages installed from OS repos, remember the standard repos keeps changing and unavailability of one mirror makes yum to try the next one in the list, this ends up consuming more time. So internal hosted mirrors are required.

SomwanshiAnjali:
> The changes look ok to me. But as discussed in the morning please add following changes:
> 
> 1. make -t argument compulsory for deployment on CentOS, if not provided error out and exit.
> 2. Use the OS repos from Seagate hosted mirrors. This will lock the versions for packages installed from OS repos, remember the standard repos keeps changing and unavailability of one mirror makes yum to try the next one in the list, this ends up consuming more time. So internal hosted mirrors are required.



> The changes look ok to me. But as discussed in the morning please add following changes:
> 
> 1. make -t argument compulsory for deployment on CentOS, if not provided error out and exit.
> 2. Use the OS repos from Seagate hosted mirrors. This will lock the versions for packages installed from OS repos, remember the standard repos keeps changing and unavailability of one mirror makes yum to try the next one in the list, this ends up consuming more time. So internal hosted mirrors are required.


@83bhp 
> first point I have addressed with latest changes but instead of making -t compulsory can we update default value of 
url_local_repo_commons="http://cortx-storage.colo.seagate.com/releases/cortx/uploads/centos/centos-7.7.1908/" to
http://cortx-storage.colo.seagate.com/releases/cortx/uploads/centos/centos-7.8.2003/

because if we consider deployment using mounted iso we won't have this build url (base url of hosted iso)

> For second I believe we provisioner do not add/update CentOS based repos.

andkononykhin2 (approved):

ypise (approved):

ipoddubnyy (approved):
looks ok to me

PR #609, opened 2020-10-08
EOS-14113, EOS-14119: Storage Enclosure Health view issue

ypise:

### DIFF
--- a/api/python/provisioner/utils.py
+++ b/api/python/provisioner/utils.py
@@ -244,15 +244,14 @@ def node_hostname_validator(
     parser_obj.read(config_path)
 
     for section in parser_obj.sections():
-        if ("srvnode" in section
-            and
-            (
-                node_dict[section]
-                != parser_obj[section]["hostname"]
-            )
-        ):
-                msg = (
-                    "Hostname values from config.ini and CLI did not match. "
-                    f"{node_dict[parser_obj[section]]} != {parser_obj[section]['hostname']}"
+        if (
+            "srvnode" in section
+            and (
+                    node_dict[section] != parser_obj[section]["hostname"]
                 )
-                raise ValueError(msg)
+        ):
+            msg = (
+                "Hostname values from config.ini and CLI did not match. "
+                f"{node_dict[section]} != {parser_obj[section]['hostname']}"
+            )
+            raise ValueError(msg)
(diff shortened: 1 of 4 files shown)


api/python/provisioner/utils.py
```diff
@@ -218,3 +222,26 @@ def repo_tgz(
...
     return dest
+
+
+# Validates the hostname string in config.ini
+# against hostname in CLI args
+def node_hostname_validator(
+    nodes: List,
+    config_path: str
+):
+    logger.debug(f"Validating list of nodes: {'srvnode-1 : '.join(nodes)}")
```
mazinamdar, line 233: nodes  is list of Node class object from setup_provisioner so we need to convert that Node object list to list of hostnames.
`nodes = [node.host for node in nodes]`

api/python/provisioner/utils.py
```diff
@@ -218,3 +222,26 @@ def repo_tgz(
...
+    logger.debug(f"Config file path: {config_path}")
+
+    parser_obj = configparser.ConfigParser()
+    parser_obj.read(config_path)
+
+    for section in parser_obj.sections():
+        if "srvnode" in section:
+            node_index = int(parser_obj[section].split("-")) - 1
```
mazinamdar, line 241: split command will result in list so we need only numder from srvnode-1 so 

 `node_index = int(parser_obj[section].split("-")) - 1`
 this line should be something like:
`node_index = int(section.split("-")[1]) - 1`

api/python/provisioner/utils.py
```diff
@@ -218,3 +222,37 @@ def repo_tgz(
...
+    logger.debug(f"Config file path: {config_path}")
+
+    parser_obj = configparser.ConfigParser()
+    parser_obj.read(config_path)
+
+    for section in parser_obj.sections():
+        if ("srvnode" in section
+            and
+            (
+                node_dict[parser_obj[section]]
```
mazinamdar, line 250: `node_dict[parser_obj[section]] ` In this section will give value as srvnode-1/srvndoe-2
so we need to make change:
`node_dict[section]`

api/python/provisioner/utils.py
```diff
@@ -218,3 +222,37 @@ def repo_tgz(
...
+                != parser_obj[section]["hostname"]
+            )
+        ):
+                msg = (
+                    "Hostname values from config.ini and CLI did not match. "
+                    f"{node_dict[parser_obj[section]]} != {parser_obj[section]['hostname']}"
```
mazinamdar, line 256: same here

mazinamdar (approved):

PR #611, opened 2020-10-09
Patch 2

pujamudaliar:
Problem Statement: License link needs to be changed.
Solution: https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/LICENSE link added.


andkononykhin2 (changes requested):
@pujamudaliar please:
- rebase you feature branch onto main, so we can see only your changes
- add clear title about the change and would be great to have a description as well
- if you have some task in Jira please add a label [EOS-XXXXX] into the title

pujamudaliar:
I will reopen another PR for this.

PR #612, opened 2020-10-09
[EOS-14096][Beta] sw update hardening beta

andkononykhin2:


ypise (approved):
@andkononykhin2 
The changes are good and should be necessary in cortx-1.0 branch.
However, do we really need this change in Beta, now that we have a fairly stable code?

83bhp:
@ypise  yes, we need this change when update is done in absence of internet.

PR #627, opened 2020-10-13
Fixes in Boxing and Unboxing

sradhanand-pati:
Added following fixes:
1. Add sleeps in between salt commands
2. Don't call function to prompt for controller IPs.
3. Don't disable root user.
    This fix is because somehow minion wasn't able to get/lost the private ipaddrs in its grain data.
    As discussed with @83bhp our private ip addr will be in static ip so we modified to fetch ipaddrs from cluster file.
    The patch to fix the root user disablement is now fixed but it needs additional fixes, so disabling it until the fixes from Beta branch are ported.

### DIFF
--- a/srv/components/system/config/sshd_boxing.sls
+++ b/srv/components/system/config/sshd_boxing.sls
@@ -15,4 +15,4 @@ Set private ips root login:
     - name: /etc/ssh/sshd_config
     - text:
       -  Match Address {{ private_ip_list|join(',') }}
-      -  "      PermitRootLogin yes"
+      -  "    PermitRootLogin yes"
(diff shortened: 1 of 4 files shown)


srv/components/system/config/sshd_boxing.sls
```diff
@@ -1,22 +1,16 @@
...
-
-
 Set private ips root login:
   file.append:
     - name: /etc/ssh/sshd_config
     - text:
       -  Match Address {{ private_ip_list|join(',') }}
-      -  "	PermitRootLogin yes"
-
+      -  "      PermitRootLogin yes"
```
83bhp, line 18: There's an extra spaces there, hope that won't matter.

83bhp (approved):

PR #628, opened 2020-10-13
EOS-14050: Unboxing misc fixes

83bhp:
Following fixes added in this patch:
1. Don't prompt for controller IPs as they are configured in-band now.
2. Add sleep between consecutive Salt commands
3. Run CSM post install, config and init scripts during unboxing.

### DIFF
--- a/srv/components/csm/render_ha_template.sls
+++ b/srv/components/csm/render_ha_template.sls
@@ -1,0 +1,26 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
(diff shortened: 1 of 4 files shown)


srv/components/csm/render_ha_template.sls
```diff
@@ -0,0 +1,26 @@
+#
```
ypise, line 1: @83bhp 
Why have we created this new file?

ypise (changes requested):

srv/components/csm/render_ha_template.sls
83bhp: To be able to run "Render CSM ha input params template" state explicitly during unboxing.
If csm.prepare is called the repos get created.

ipoddubnyy (approved):
looks ok to me

PR #634, opened 2020-10-14
[EOS-14134]: Commit the initial implementation of checker/validator API

dmitrydb:
Add initial implementation of the checker/validator for Provisioner API. Implement basic SW Update CORTX pre-check validations.


api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+            else:
+                res[str(key.keypath)] = cfg.CheckVerdict.PASSED.value
+
+        return res
+
+    @staticmethod
+    def _communicability(*, args: str, targets: str = cfg.ALL_MINIONS,
+                         dry_run: bool = False) -> dict:
+        """
+        Check if nodes can communicate using `salt test.ping`
```
andkononykhin2, line 132: Test.ping verifies  master minion connection.please adjust the title a bit. Also test.ping is not the best choice since it will hang if a minion is down. Consider to use that call https://github.com/Seagate/cortx-prvsnr/blob/9e5677a67a14480acc3f7690f33e003fee3a346f/api/python/provisioner/salt_minion.py#L49 instead.

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+                        target nodes
+        :return:
+        """
+        if dry_run:
+            return dict()
+
+        res = dict()
+        minion_id = local_minion_id()
+        try:
+            _res = function_run("cmd.run", targets=minion_id,
```
andkononykhin2, line 173: https://github.com/Seagate/cortx-prvsnr/blob/9e5677a67a14480acc3f7690f33e003fee3a346f/api/python/provisioner/salt.py#L948

cmd_run has been improved and can now provide you all necessary features, you may use it directly

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+            :param _targets
+            :return:
+            """
+            for phrase in key_phrases:
+                cmd = self._LOGS_ARE_GOOD_CHECK_CMD_TMP.format(
+                    log_filename=log_file,
+                    key_phrase=phrase)
+                try:
+                    function_run("cmd.run", targets=_targets, fun_args=[cmd])
+                    res[minion_id] = {
```
andkononykhin2, line 244: Are you sure these outer scope vars would work? In any case looks awkward.

andkononykhin2 (changes requested):
Looks good for the moment. But some changes are desired. Thanks

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+    Base class for system and system components health and status validation
+    """
+
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+    _run_args_type = CheckArgs
+
+    _PRV_METHOD_MOD = "_"  # private method modificator
+    # -c 1 means count of sending ECHO_REQUEST packets
+    # -W 1 means timeout to wait for a response
+    _PING_CMD_TMP = "ping -c 1 -W 1 {server_addr}"  # ping command template
```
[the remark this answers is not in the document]
dmitrydb: Done

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+            inputs.METADATA_ARGPARSER: {
+                'help': ("optional arguments and parameters for "
+                         "particular check")
+            }
+        },
+        default=""  # empty string
+    )
+
+    targets: str = RunArgs.targets
+    dry_run: bool = RunArgs.dry_run
```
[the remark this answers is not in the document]
dmitrydb: Fixed

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+            _res = function_run("cmd.run", targets=minion_id,
+                                fun_args=[self._CLUSTER_STATUS_CHECK_CMD])
+            # TODO: parse command if necessary
+            res[minion_id] = f"{cfg.CheckVerdict.PASSED.value}: {str(_res)}"
+        except SaltCmdResultError as e:
+            res[minion_id] = f"{cfg.CheckVerdict.FAIL.value}: {str(e)}"
+
+        return res
+
+    def check_all(self, check_args: str = "", targets: str = cfg.ALL_MINIONS,
```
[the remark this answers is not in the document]
dmitrydb: Deleted this method

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+    _run_args_type = CheckArgs
+
+    _PRV_METHOD_MOD = "_"  # private method modificator
+    # -c 1 means count of sending ECHO_REQUEST packets
+    # -W 1 means timeout to wait for a response
+    _PING_CMD_TMP = "ping -c 1 -W 1 {server_addr}"  # ping command template
+    _BMC_CHECK_CMD = 'ipmitool chassis status | grep "System Power"'
+    _CLUSTER_STATUS_CHECK_CMD = "pcs status --full cluster"
```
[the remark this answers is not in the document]
dmitrydb: Done

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+        res = dict()
+        for addr in servers:
+            # NOTE: check ping of 'srvnode-2' from 'srvnode-1'
+            # and vise versa
+            # TODO: which targets do we need to use? Because we need to
+            #  check cross connectivity if nodes > 2
+            targets = set(servers) - {addr}
+
+            targets = (cfg.LOCAL_MINION if not targets
+                       else next(iter(targets)))  # takes just one node
```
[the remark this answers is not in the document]
dmitrydb: Yes, this to-do label already exists in the code

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+        if check_name in cfg.CHECKS:
+            _res = getattr(self,
+                           self._PRV_METHOD_MOD + check_name)(args=check_args,
+                                                              targets=targets,
+                                                              dry_run=dry_run)
+            res[check_name] = _res  # aggregate result to simple dict form
+        else:
+            raise ValueError(f'Check "{check_name}" is not supported')
+
+        return res
```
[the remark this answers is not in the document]
dmitrydb: As we discuss with Ivan, we need to return all aggregate results. I will improve code to wrap this result in some new class which will simply allow to go through all passed/failed checks.

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,392 @@
...
+                                                              targets=targets,
+                                                              dry_run=dry_run)
+            res[check_name] = _res  # aggregate result to simple dict form
+
+        return res
+
+    def run(self, check_name: cfg.Checks, check_args: str = "",
+            targets: str = cfg.ALL_MINIONS, dry_run: bool = False) -> dict:
+        """
+        Basic run method to execute remote commands on targets nodes:
```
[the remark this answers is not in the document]
dmitrydb: Fixed method description.

api/python/provisioner/commands/check.py
dmitrydb: Done

api/python/provisioner/commands/check.py
dmitrydb: Done

api/python/provisioner/commands/check.py
dmitrydb: Yes, moved vars definition before this function

andkononykhin2:
@dmitrydb please update with the latest changes from the target branch. Thanks

dmitrydb:
@andkononykhin2 Done

andkononykhin2:
@dmitrydb please update with latest changes from target branch and resolve the conflicts

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+
+        res: CheckEntry = CheckEntry(cfg.Checks.BMC_ACCESSIBILITY.value)
+
+        minion_id = local_minion_id()
+
+        try:
+            _res = cmd_run(_BMC_CHECK_CMD, targets=minion_id,
+                           fun_kwargs=dict(python_shell=True))
+            # TODO: parse command output if necessary
+            res.set_passed(target=minion_id, comment=str(_res))
```
andkononykhin2, line 361: better to use `else` block for that since an exception inside `set_passed` will lead to `set_fail` call which will shift an attention internal logic to a check fail

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+            # and vise versa
+            # TODO: which targets do we need to use? Because we need to
+            #  check cross connectivity if nodes > 2 and all possible pairs
+            targets = set(servers) - {addr}
+            targets = (cfg.LOCAL_MINION if not targets
+                       else next(iter(targets)))  # takes just one node
+
+            cmd = _PING_CMD_TMPL.format(server_addr=addr)
+            try:
+                cmd_run(cmd, targets=targets)
```
andkononykhin2, line 395: here `targets` are nodes on which ping is run, and `addr` is the node which is pinged. So, is it ok to set `target` to that value?

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+        # grep "unable to stop resource" /var/log/pacemaker.log
+        # grep "reboot" /var/log/corosync.log
+        # grep "error" /var/log/corosync.log"
+
+        # they should be a constants
+        corosync_log = "/var/log/cluster/corosync.log"
+        pacemaker_log = "/var/log/pacemaker.log"
+
+        corosync_key_phrases = (
+            "reboot",
```
andkononykhin2, line 461: I think these may lead to false positives since:
- logs may include outdated records (e.g. an issue has been fixed already and cluster is good now)
- some words might not be directly related to a bad state of the cluster since it is very specific to how different components treat different cases (e.g. 'reboot might be required') or depends on context (e.g. '0 errors were detected')

@azheregelya Could you provide your vision on that: is it ok at all to rely on any phrases in logs to make a desicion regarding the whole system state? Do you think there are some good phrares for that?

@dmitrydb Generally I think it would be safer to exclude that check from SWUpdate routine:
- either skip it totally
- or do not fail if any phrases are found and just warn

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+
+            targets = (cfg.LOCAL_MINION if not targets
+                       else next(iter(targets)))  # takes just one node
+
+            cmd = _PSWDLESS_SSH_CHECK_CMD_TMPL.format(user=user,
+                                                      hostname=addr)
+            try:
+                _res = cmd_run(cmd, targets=targets)
+            except SaltCmdResultError:
+                res.set_fail(target=targets,
```
andkononykhin2, line 514: the same concern as above for ping command

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+        :return:
+        """
+        # NOTE: per my discussion with Andrey Kononykhin replace
+        #  `pcs status --full cluster` and its complex parsing by existing
+        #  ensure_cluster_is_healthy call
+        res: CheckEntry = CheckEntry(cfg.Checks.CLUSTER_STATUS.value)
+
+        try:
+            ensure_cluster_is_healthy()
+            res.set_passed(target=local_minion_id())
```
andkononykhin2, line 538: the same concern as above

api/python/provisioner/commands/sw_update.py
```diff
@@ -0,0 +1,244 @@
...
+# TODO consider to use RunArgsUpdate and support dry-run
+@attr.s(auto_attribs=True)
+class SWUpdate(CommandParserFillerMixin):
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+
+    def run(self, targets):  # noqa: C901 FIXME
+        # logic based on https://jts.seagate.com/browse/EOS-6611?focusedCommentId=1833451&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1833451  # noqa: E501
+
+        checker = Check()
+        check_res = checker.run(targets=LOCAL_MINION)
```
andkononykhin2, line 67: That block should be moved inside the following try block. Actually `ensure_cluster_is_healthy` check at the beginning it is exactly the same but with less deepness.

Also I believe it's not safe to make all checks mandatory since the new logic is very new and not enough tested BUT it may breaks SW update logic in false positives cases.

So, I would suggest:
- do all checks
- but do not fail and just post on INFO level if everything is passed and on WARNING level if anything is failed

andkononykhin2 (changes requested):
@dmitrydb This looks like a big change with a set good ideas. Thanks.
But I have few major concerns:
- it's dangerous to fully enable these checks for SW Update logic, we need time to monitor how it works and fix if any false positives are detected
- you moved SW Update logic into a separate file, better to do that in a separate PR since currently it mixes non-related changes and breaks unit tests

api/python/provisioner/commands/check.py
azheregelya: Interesting phrases:
`Forcing <resource-name> away from <node-name>`
`node is unclean`
`can fence`
`node <node-name> will be fenced`

Existing example with "unable to stop resource" is good as well.
Recommended file /var/log/cluster/corosync.log

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+
+            return _res
+
+        # grep "unable to stop resource" /var/log/pacemaker.log
+        # grep "reboot" /var/log/corosync.log
+        # grep "error" /var/log/corosync.log"
+
+        # they should be a constants
+        corosync_log = "/var/log/cluster/corosync.log"
+        pacemaker_log = "/var/log/pacemaker.log"
```
azheregelya, line 458: Normally `pacemaker.log` does not contain anything interesting. From my experience it wasn't helpful.

api/python/provisioner/commands/check.py
azheregelya: Another important moment - we need make sure that only new events are considered. This becomes relevant in case of update when system already has some "history" in corosync.log

api/python/provisioner/commands/check.py
dmitrydb: Done

api/python/provisioner/commands/check.py
dmitrydb: Done

api/python/provisioner/commands/check.py
dmitrydb: Yes, we are trying to check cross-connectivity between nodes. Also in future we need to improve this to check all possible combinations of pairs (target host which performs ping command and destination host). The code already had the necessary TODO comments:
```python
            # TODO: which targets do we need to use? Because we need to
            #  check cross connectivity if nodes > 2 and all possible pairs
            targets = set(servers) - {addr}
            targets = (cfg.LOCAL_MINION if not targets
                       else next(iter(targets)))  # takes just one node
```

api/python/provisioner/commands/check.py
dmitrydb: Done

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+    def __init__(self, check_name: str):
+        """
+        Setup check
+        Args:
+            str check_name: check/validation name or label
+        """
+        self._check_name = check_name
+        self._verdict = None
+        self._comment = None
+        self._target = None
```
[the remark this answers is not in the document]
dmitrydb: Done

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+from ..vendor import attr
+
+
+SRVNODE1 = "srvnode-1"
+SRVNODE2 = "srvnode-2"
+
+PENDING_SERVERS = frozenset({SRVNODE1, SRVNODE2})  # to make constant immutable
+
+
+class CheckEntry:
```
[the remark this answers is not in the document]
dmitrydb: Added TODO comment for the future. Will create a new ticket for that.

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+        try:
+            ensure_cluster_is_healthy()
+            res.set_passed(target=local_minion_id())
+        except Exception:
+            res.set_fail(target=local_minion_id())
+
+        return res
+
+    def run(self, check_name: cfg.Checks = None, check_args: str = "",
+            targets: str = cfg.ALL_MINIONS) -> CheckResult:
```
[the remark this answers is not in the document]
dmitrydb: Yes, it was planned to be used but previously I leaved this parameter for the future reuse. Delete it for the current moment.

api/python/provisioner/api_spec.yaml
```diff
@@ -122,3 +122,5 @@ deploy_jbod:
   type: DeployJBOD
 deploy_dual:
   type: DeployDual
+check:
```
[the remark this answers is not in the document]
dmitrydb: Fixed

api/python/provisioner/commands/check.py
andkononykhin2: the check fails if any of source services fails to connect to destination one.
Let's imaging that:
- we have 5 nodes
- node 1 and node 2 are not ping-able from some of other ones.

What would be passed to `set_fail`? My guess:

- set_fail(target={node2, node3, node4, node5})
- set_fail(target={node1, node2, node4, node5})

Don't you think that:
- set_fail(target=node1)
- set_fail(target=node2)

would be clearer and easier to use if we want to extract a node that fails the check?

dmitrydb: > * set_fail(checked_target=node1)
> * set_fail(checked_target=node2)

Yes, this method will be used to create a validation result. At the current moment, we take just one node for validation. During our improvement, we will have as many records as the number of pairs. I used this approach in other checks such as `_logs_are_good`, `_network`. It is the most simple way to do such validation instead of the creation of some new check instance which encapsulates a series of check results.

dmitrydb: So this method will return list of `CheckEntry` instances.

api/python/provisioner/commands/sw_update.py
dmitrydb: Done. Isolate this logic in separate `SWUpdateDecisionMaker` class.

api/python/provisioner/commands/check.py
dmitrydb: I guess to move these improvements in separate PR to check and debug that logic separately from the main flow.

api/python/provisioner/commands/check.py
```diff
@@ -0,0 +1,578 @@
...
+
+        """
+        for check in self._check_entries:
+            yield check
+
+    def __str__(self):
+        """String representation of all checks"""
+        # return str(self.to_dict())
+        # TODO: maybe it is better to use json.dump
+        return json.dumps(self.to_dict(), indent=4)
```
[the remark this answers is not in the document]
dmitrydb: Will do it in separate PR to carefully debug and test serialization for `CheckResult` classes with output option and without.

andkononykhin2 (approved):

ypise (approved):

PR #641, opened 2020-10-15
[EOS-6078] adds base logic for sw update repo dynamic validation

andkononykhin2:
- sets the framework for set commands dynamic validation
- implements logic of temporarily dynamic configuration for a new repo candidate to allow safe pre-checks before actual repo activation

### DIFF
--- a/api/python/provisioner/commands/__init__.py
+++ b/api/python/provisioner/commands/__init__.py
@@ -474,8 +474,9 @@ class SetSWUpdateRepo(Set):
     # TODO at least either pre or post should be defined
     input_type: Type[inputs.SWUpdateRepo] = inputs.SWUpdateRepo
 
+    @staticmethod
     def _prepare_repo_for_apply(
-        self, repo: inputs.SWUpdateRepo, enabled: bool = True
+        repo: inputs.SWUpdateRepo, enabled: bool = True
     ):
         # if local - copy the repo to salt user file root
         # TODO consider to use symlink instead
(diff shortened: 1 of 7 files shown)


api/python/provisioner/commands/__init__.py
```diff
@@ -465,20 +474,93 @@ class SetSWUpdateRepo(Set):
...
+            #   - there is no the same release repo already active EOS-13715
+            #   - (optionally) new version is higher then currently installed
+
+            logger.debug(f"Resolved metadata {metadata}")
+            repo.metadata = metadata
+
+            # TODO IMPROVE EOS-6078
+            # - set 'release' to version from resolved metadata,
+            #   that would make `release` param unnecessary at all
+            #   repo.release = ...
```
ypise, line 545: Why are we not handling the exception here if we are putting it in a try block?

ypise (approved):

api/python/provisioner/commands/__init__.py
andkononykhin2: we need finally block here mostly (we might consider context manager for that in future).
Exceptions will just go outside. That might be also improved later if needed since it is related to current efforts for messaging improvement.

andkononykhin2: we need finally block here mostly (we might consider context manager for that in future).
Exceptions will just go outside. That might be also improved later if needed since it is related to current efforts for messaging improvement.

PR #643, opened 2020-10-15
[WIP] EOS-14136: Factory Deployment validations

divya-vijay:
1. Kept this entire implementation in plain python, without other dependencies, for now.
2. In case of validation failure, captured the error message in a common json format and not raised Exception, so as to continue with other checks.
3. Check/error/log messages will be made common and generic and will go into user_messages.py, (similar to config.py), which can be imported to code.
4. This is the skeletal structure of Validation Framework for Factory deployment (covering pre- and post- checks), over which other layers will be built. This is open to alterations before arriving at the final structure. 
5. To Test and finalise output parsing.


andkononykhin2 (changes requested):
@divya-vijay Hello Divya. I see you have done a significant job. But that copying from api is not that is actually necessary here. Please either move your checks to api or remove all these api modules from here since as I understand you need only subprocess.run method for all your checks.

ipoddubnyy (approved):
it looks ok to me. but someone from with better python skills should review as well.

divya-vijay:
> it looks ok to me. but someone from with better python skills should review as well.

Thanks Ivan, this is a base structure over which we'll build the rest of the validations on. So, wanted to get this verified, before proceeding. I'll add for others and update the PR.

divya-vijay:
> @divya-vijay Hello Divya. I see you have done a significant job. But that copying from api is not that is actually necessary here. Please either move your checks to api or remove all these api modules from here since as I understand you need only subprocess.run method for all your checks.

Thanks Andrey! Sure, just had them as placeholder incase needed for other implementations. Will update !

validations/scripts/utils/common.py
```diff
@@ -0,0 +1,151 @@
...
+def get_repo_archive_exclusions():
+    exclude = []
+    for d in config.REPO_BUILD_DIRS + ['*.swp']:
+        exclude.extend(['--exclude', str(d)])
+    return exclude
+
+
+# Validates the hostname string in config.ini
+# against hostname in CLI args
+def node_hostname_validator(
```
ypise, line 123: This should not be part of **_commons.py_**, but part of _**network.py**_

validations/scripts/utils/config.py
```diff
@@ -0,0 +1,293 @@
...
+    PROJECT_PATH = CONFIG_MODULE_DIR.parents[2]
+except IndexError:
+    PROJECT_PATH = None
+else:
+    # TODO IMPROVE more accurate way for that
+    for path in ('srv', 'pillar', 'files'):
+        if not (PROJECT_PATH / path).is_dir():
+            PROJECT_PATH = None
+            break
+PRVSNR_ROOT_DIR = Path('/opt/seagate/cortx/provisioner')
```
ypise, line 35: Add an empty line before line #[35](https://github.com/Seagate/cortx-prvsnr/pull/643/files#diff-dbab3b67a4207ea48b16836575a1bccc3b18aa8c28de3d478d083a278d1c0438R35)

validations/scripts/utils/config.py
```diff
@@ -0,0 +1,293 @@
...
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+from enum import Enum
+from pathlib import Path
+from typing import Union, Dict, Optional
+
+CONFIG_MODULE_DIR = Path(__file__).resolve().parent
+
```
ypise, line 22: This file has a lot of data specific to Provisioner.
It should not be part of `validations/scripts/utils/config.py`
It should be maintained in Provisioner API.
`validations` directory should be independent of Provisioner and Salt. It is supposed to be stand alone, no hard bindings

validations/scripts/utils/errors.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def __str__(self):
+        return (
+            'decode failed for {}, reason: {!r}'
+            .format(self.spec, self.reason)
+        )
+
+
+class SWUpdateRepoSourceError(ProvisionerError, ValueError):
+    _prvsnr_type_ = True
```
ypise, line 70: Again this is specific to Provisioner and should be part of Provisioner API

validations/scripts/utils/errors.py
```diff
@@ -0,0 +1,201 @@
...
+        self.reason = reason
+
+    def __str__(self):
+        return (
+            'failed to apply Hare post_update logic, reason: {!r}'
+            .format(self.reason)
+        )
+
+
+class ClusterNotHealthyError(ProvisionerError):
```
ypise, line 144: If at all we are planning to implement this, it should be Python version of:
https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/cli/src/common_utils/utility_scripts.sh#L51

Also, no Provisioner specific implementation: `ProvisionerError`

validations/scripts/utils/errors.py
```diff
@@ -0,0 +1,201 @@
...
+        self.reason = reason
+
+    def __str__(self):
+        return (
+            'failed to apply Hare post_update logic, reason: {!r}'
+            .format(self.reason)
+        )
+
+
+class SWUpdateError(ProvisionerError):
```
ypise, line 157: This is not a Pre-Check
This is unnecessary in Validations framework

validations/scripts/utils/errors.py
```diff
@@ -0,0 +1,201 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+from typing import Dict, Union, Any
```
ypise, line 18: This file is unnecessary for now

ypise (changes requested):

ypise:
> @divya-vijay Hello Divya. I see you have done a significant job. But that copying from api is not that is actually necessary here. Please either move your checks to api or remove all these api modules from here since as I understand you need only subprocess.run method for all your checks.

I agree.

validations/scripts/field/__init__.py
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
ypise, line 2: Why do we need this file?

validations/scripts/utils/__init__.py
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
```
ypise, line 3: Why do we need this file?

ypise (changes requested):

validations/scripts/utils/errors.py
divya-vijay: Hello @ypise , Thank you. 
Yes, like I had mentioned previously, I created the initial draft as a sample for how other files will look in Validations Framework. since we're continuously implementing new ideas and changes to the framework, I just had them as placeholders. I've updated it all to latest version in the today's commit. Some functions are yet to be tested and changed accordingly. But this latest commit gives an overall idea on the complete structure of the framework.

validations/scripts/factory/cluster_readiness_check.py
```diff
@@ -0,0 +1,180 @@
...
+        if res[0]:
+            response["message"]= str(PACEMAKER_HEALTH_ERROR)
+        else:
+            for node in nodes:
+                if node in res[1]:
+                    response["message"]= str(PACEMAKER_HEALTH_CHECK)
+                    response["ret_code"]= 0
+                else:
+                    response["message"]= str(PACEMAKER_HEALTH_ERROR)
+                    response["ret_code"]= 1
```
sanjognaik, line 55: Please break from the loop in case of error to avoid overwriting of 'message'.

validations/scripts/factory/cluster_readiness_check.py
```diff
@@ -0,0 +1,180 @@
...
+        nodes = node_res['response']
+        res = self.cluster.nodes_status()
+        if res[0]:
+            response["message"]= str(NODES_OFFLINE)
+        else:
+            for node in nodes:
+                if node in res[1]:
+                    response["message"]= str(NODES_ONLINE)
+                else:
+                    response["message"]= str(NODES_OFFLINE)
```
sanjognaik, line 77: Same as above. 'break'

validations/scripts/factory/cluster_readiness_check.py
```diff
@@ -0,0 +1,180 @@
...
+            print ("TEST", response["message"])
+            response["message"]= str(CLUSTER_HEALTH_ERROR)
+        else:
+            for node in nodes:
+                if node in res[1]: 
+                    response["message"]= str(CLUSTER_HEALTH_CHECK)
+                    response["ret_code"]= 0
+                else:
+                    response["message"]= str(CLUSTER_HEALTH_ERROR)
+                    response["ret_code"]= 1
```
sanjognaik, line 124: Same as above. 'break'

validations/scripts/factory/cluster_readiness_check.py
```diff
@@ -0,0 +1,180 @@
...
+        response["response"]= ping_check[1]
+        response["error_msg"]= ping_check[2]
+        return response
+
+    def check_bmc_accessible(self):
+        ''' Validations for BMC accessibility
+        '''
+        response = {}
+        bmc_res = self.bmc.bmc_accessible()
+        bmc_ping_res = self.bmc.ping_bmc()
```
sanjognaik, line 172: Function bmc.bmc_accessible() already calls ping_bmc() along with checking power status. So no need to call ping_bmc() here again. Also check_bmc_accessible() function is already implemented in unboxing_readiness_check.py with same name. I would suggest to modify bmc.bmc_accessible() to bmc.check_power_status() to just check power status, so that you can call both (ping_bmc() and check_power_status()) functions separately whenever required. Ref verify_bmc_accessible() validations/scripts/factory/bmc_checks.py [PR #655 ](https://github.com/Seagate/cortx-prvsnr/pull/655/files)

validations/scripts/factory/unboxing_readiness_check.py
```diff
@@ -0,0 +1,118 @@
...
+        response["error_msg"]= bmc_res["error_msg"]
+        return response
+
+
+    def check_controller_mc_accessible(self):
+        ''' Validations for Controllers accessibility
+        '''
+        response = {}
+        node_res = PillarGet.get_pillar("cluster:node_list")
+        if not node_res['ret_code']:
```
sanjognaik, line 99: I would prefer return node_res immediately 'if node_res['ret_code']:'.

validations/scripts/factory/unboxing_readiness_check.py
```diff
@@ -0,0 +1,118 @@
...
+                ctrl_ip = PillarGet.get_pillar("cluster:{node}:bmc:ip")
+                user = PillarGet.get_pillar("cluster:{node}:bmc:user")
+                enc_secret = PillarGet.get_pillar("cluster:{node}:bmc:secret")
+                get_decrypt_secret = decrypt_secret("ldap", enc_secret)
+                if not get_decrypt_secret["ret_code"]:
+                    passwd = get_decrypt_secret["response"]
+                    ssh_ctrl_ip = ssh_remote_machine(hostname, user, passwd)
+                    response["message"]= ssh_ctrl_ip["message"]
+                else:
+                    response["message"]= str(DECRYPT_PASSWD_FAILED)
```
sanjognaik, line 111: Please 'break' here to avoid overwriting of response.

validations/scripts/factory/unboxing_readiness_check.py
```diff
@@ -0,0 +1,118 @@
...
+        else:
+            response["message"]= str(CTRL_IP_ACCESSIBLE_ERROR)
+
+        response["ret_code"]= res[0]
```
sanjognaik, line 115: Remove these 3 lines as 'res' is not defined.

validations/validators.py
```diff
@@ -0,0 +1,92 @@
...
+        if args.postcheck:
+            check_list = config.FACTORY_POST_CHECK
+        elif args.precheck:
+            check_list = config.FACTORY_PRE_CHECK
+        elif args.replacenode:
+            check_list = config.REPLACE_NODE_CHECK
+        elif args.fwupdate:
+            check_list = config.FW_UPDATE_CHECK
+        elif args.unboxing:
+            check_list = config.UNBOXING_CHECK
```
sanjognaik, line 42: We need to separate this into something like '--preunboxing' and '--postunboxing' as we have pre and post checks for unboxing. Please take a look at validations/validators.py and validations/config.py in PR#655.

validations/scripts/utils/hardware.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __init__(self):
+        ''' Validations for pre-factory HW steps
+        '''
+        pass
+
+    def mlnx_ofed_installed(self):
+        ''' Validations for Mellanox drivers
+        '''
+        cmd = "yum list yum list mlnx-ofed-all | grep mlnx-ofed*"
```
mazinamdar, line 36: can we just check rpm is installed or not using " rpm -qa | grep mlnx-ofed-all " command

validations/scripts/factory/pre_flight_check.py
```diff
@@ -0,0 +1,141 @@
...
+        response["response"]= res[1]
+        response["error_msg"]= res[2]
+        return response
+
+    def check_mlnx_hca_req_ports(self):
+        ''' Validations for Mellanox drivers
+        '''
+        response = {}
+        res = self.hw.mlnx_hca_req_ports()
+        if res[0] != 0:
```
mazinamdar, line 68: this should be res ?

divya-vijay:
Stale PR - created as Provisioner-specific checks before Validations Framework was in place. Closing this as we have new implementation in place.

PR #645, opened 2020-10-16
EOS-14342: Install jq package in system component

SomwanshiAnjali:


srv/components/system/install.sls
```diff
@@ -25,6 +25,7 @@ Install_base_packages:
       - rsync
       - ftp
       - sshpass
+      - jq
```
andkononykhin2, line 28: @SomwanshiAnjali why we don't add cleanup in teardown.sls here?

ypise (approved):

PR #649, opened 2020-10-16
EOS-13909 Refactoring replace internal code  with replace_node API call

IgorPivovarovDSR:
1) disable internal function calls
2) enable replace_node API call


IgorPivovarovDSR:
Will be recreated

PR #659, opened 2020-10-19
EOS-14392: Update permissions for rabbitmq rotated files

SomwanshiAnjali:

### DIFF
--- a/srv/components/misc_pkgs/rabbitmq/files/rabbitmq-server
+++ b/srv/components/misc_pkgs/rabbitmq/files/rabbitmq-server
@@ -1,4 +1,5 @@
 /var/log/rabbitmq/*.log {
+    create 644 rabbitmq rabbitmq
     missingok
     rotate 7
     compress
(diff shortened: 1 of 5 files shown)


srv/components/misc_pkgs/rabbitmq/files/rabbitmq-server
```diff
@@ -1,4 +1,5 @@
 /var/log/rabbitmq/*.log {
+    create 644 rabbitmq rabbitmq
     missingok
```
SomwanshiAnjali, line 3: @ypise 
su command was throwing below error and permissions was not getting set properly hence used create command.
error setting owner of /var/log/rabbitmq/rabbit@srvnode-1.log to uid 0 and gid 10: Operation not permitted

ypise: @SomwanshiAnjali 
Looks better. This actually would apply to all files in the directory: `/var/log/rabbitmq/`

ypise (approved):

PR #661, opened 2020-10-19
[EOS-14406] Adds support for bootstrap key

andkononykhin2:
- introduces `--bootstrap-key PATH` to specify location of a ssh private key which would be used to access a system at the very beginning
- automates setup-provisioner routine for a test


ypise (approved):
@andkononykhin 
Please do support this new functionality with a User Doc/Guide

PR #663, opened 2020-10-20
[EOS-13909] Update unboxing logic to support active-active Provisioner configuration

IgorPivovarovDSR:
-disable internal function calls
-enable replace_node API call

### DIFF
--- a/cli/src/factory_ops/unboxing/init
+++ b/cli/src/factory_ops/unboxing/init
@@ -204,13 +204,18 @@ check_mgmt_ip
 # Run pre_unbox validations
 pre_unbox
 
+echo "Running provisioner replace node api" >> ${LOG_FILE}
+provisioner replace_node --node-host 192.168.0.2 --logfile --logfile-filename ${LOG_FILE}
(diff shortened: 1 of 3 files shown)


cli/src/factory_ops/unboxing/init
```diff
@@ -204,13 +204,18 @@ check_mgmt_ip
 # Run pre_unbox validations
 pre_unbox
 
+echo "Running provisioner replace node api" >> ${LOG_FILE}
+provisioner replace_node --node-host 192.168.0.2 --logfile --logfile-filename ${LOG_FILE}
```
andkononykhin2, line 208: @83bhp Can we always rely on that hard-coded IP?

andkononykhin2 (commented):
@IgorPivovarovDSR Looks good to me. I have just one concern

cli/src/factory_ops/unboxing/init
83bhp: for now, yes, but can be put in a variable, will do that.

PR #664, opened 2020-10-20
[EOS-6078]: Add ISO validation for SW update

dmitrydb:
Add ISO repo validation for the `set_swupdate_repo` command: 

- check metadata info
- there is no the same sw update repo activated at current moment
- yum repo base checks


api/python/provisioner/commands/__init__.py
```diff
@@ -48,7 +48,9 @@
     CONTROLLER_BOTH,
     SSL_CERTS_FILE,
     SEAGATE_USER_HOME_DIR, SEAGATE_USER_FILEROOT_DIR_TMPL,
-    REPO_CANDIDATE_NAME
+    REPO_CANDIDATE_NAME,
+    RELEASE_INFO,
+    ReleaseInfo
```
ypise, line 53: There's repetition and also the case discipline isn't followed.
Any specific reason?

api/python/provisioner/commands/__init__.py
```diff
@@ -536,13 +536,63 @@ def dynamic_validation(self, params: inputs.SWUpdateRepo, targets: str):
...
+                   'list available')
+            try:
+                salt_cmd_run(cmd, targets=LOCAL_MINION)
+            except SaltCmdRunError as e:
+                # SW Update repo is malformed
+                logger.debug("Can't enable SW update candidate repo: "
+                             f"'{str(e)}'")
+
+            # TODO: take it from pillar
+            iso_mount_dir = f'/opt/seagate/cortx/updates/{REPO_CANDIDATE_NAME}'
```
ypise, line 550: Can this base directory be set in config.py?
There's a possibility for change regarding this path.

api/python/provisioner/config.py
```diff
@@ -64,6 +64,7 @@
 
 
 REPO_CANDIDATE_NAME = 'candidate'
+RELEASE_INFO = 'RELEASE.INFO'
```
ypise, line 67: This is a file name and this const can be elaborated to avoid confusion with the ENUM below

ypise (changes requested):

ypise (commented):
@dmitrydb 
My comments can be overlooked if we have @andkononykhin2's approval

api/python/provisioner/config.py
dmitrydb: Fixed

api/python/provisioner/commands/__init__.py
dmitrydb: There is pillar value for this directory. It is used for on-fly mount of ISO during validation and I can reuse it.
Added logic to retrieve this value from pillars

api/python/provisioner/commands/__init__.py
dmitrydb: I renamed `RELEASE_INFO` constant for file name to `RELEASE_INFO_FILE` to avoid confusing. For classes in Python according to PEP8 CamelCase is used. As `ReleaseInfo` is class-enumeration I use this camel case for it.

dmitrydb:
@ypise Resolved your comments

ypise (approved):

api/python/provisioner/commands/__init__.py
```diff
@@ -493,7 +496,52 @@ def _prepare_repo_for_apply(
...
-    def dynamic_validation(self, params: inputs.SWUpdateRepo, targets: str):
+    @staticmethod
+    def _get_mount_dir():
+        """
+        Return the base mount directory for ISO from pillar values
+
+        :return:
+        """
+        release_path = KeyPath('release')
+        update_dir = PillarKey(release_path / 'update/base_dir')
```
t7ko-seagate, line 507: Could be simplified as:

```
update_dir = PillarKey('release/update/base_dir')
```

dmitrydb: Fixed

andkononykhin2:
> general comment/request: this functionality should also support OS ISO. We'll be using this model for controlled OS updates in the [near] future.

@ipoddubnyy From functionality perceptive we should be good. But it will require some design changes in code and review of our variables naming. I suggest to move that requirement to a separate task.

ypise (approved):

api/python/provisioner/commands/__init__.py
```diff
@@ -514,7 +585,18 @@ def dynamic_validation(self, params: inputs.SWUpdateRepo, targets: str):
...
         # TODO IMPROVE VALIDATION EOS-14350
-        #   - there is no other candidate that is being verified
+        #   - there is no other candidate that is being verified:
+        #     if found makes sense to raise an error in case the other
+        #     logic is still running, if not - forcibly remove the previous
+        #     candidate
+        #   - after first mount 'sw_update_candidate' listed in disabled repos
+        if self._does_repo_exist(f'sw_update_{candidate_repo.release}'):
+            logger.warning(
+                'other repo candidate was found, proceed with force removal'
```
andkononykhin2, line 595: @dmitrydb could you update the message: `proceeding`

api/python/provisioner/errors.py
```diff
@@ -115,13 +115,11 @@ class SWUpdateRepoSourceError(ProvisionerError, ValueError):
 
     def __init__(self, source: str, reason: str):
         self.source = source
-        self.reason = reason
+        self.reason = (f'repo source "{self.source}" is not acceptable, '
```
andkononykhin2, line 118: why do we need that change: initially that attribute kept original reason for later use as per consumer needs. I don't think we should construct a string ready for __str__, it would limit the usage of the attribute.

andkononykhin2 (changes requested):

api/python/provisioner/errors.py
dmitrydb: I did that for proper error message understanding on console output. Returned code back for now.

api/python/provisioner/commands/__init__.py
dmitrydb: Fixed

andkononykhin2 (approved):

PR #669, opened 2020-10-21
EOS-13827: Add provisioner API overview doc

t7ko-seagate:


ypise (approved):

docs/Provisioner-Introduction.md
```diff
@@ -0,0 +1,71 @@
...
+
+Provisioner exposes an API which other components may use to run certain operations through Provisioner.
+
+Main page: [Northbound-Interface](https://github.com/Seagate/cortx-prvsnr/wiki/Northbound-Interface).
+
+This API is exposed in 3 forms:
+
+* Library (python module).
+* CLI (main entry point is command-line utility called `provisioner`; run `provisioner --help` for list of available commands).
+* CLI-wrapper library.  It's a python module, but it's "decoupled" from main library listed above, and does not change frequently with versions which allows for special strict linkage of this python module, e.g. in a form of python frozen app (and there is no need to re-link every time there is new version of main Provisioner library).
```
andkononykhin2, line 62: @t7ko-seagate I think we shouldn't emphasize the stability of the API here since it actually doesn't follow any specific rules here: if API needs to be changed we do that. The main rule is to keep it in-sync with underlying parts (cli and lowest level python API).
In the current state the main reason of that 3rd level stability (and random updates) is that it exposes not all commands available for lower levels since:
- the main consumer (CSM) doesn't require them for now
- while updating that level usually requires just few lines of code it will lead to additional testing and for sure some specific bugs might appear due to additional shell-related layer

andkononykhin2 (approved):
@t7ko-seagate Thank you. Looks like a good entry point doc for new people which we have been missing for a long. I have just one minor concern regarding API levels explanation.

PR #677, opened 2020-10-23
[EOS-14427] masks secrets in logs

andkononykhin2:


andkononykhin2:
@ypise @mazinamdar when you will be reviewing the PR please don't wonder changes in test_errors.py : I applied `dos2unix` to that file since it had been saved in windows editor by some reason. That led to some inconvenience.

api/python/provisioner/salt.py
```diff
@@ -798,6 +809,7 @@ def _salt_runner_cmd(  # noqa: C901 FIXME
     fun_args: Union[Tuple, None] = None,
     fun_kwargs: Union[Dict, None] = None,
     nowait=False,
+    secure=False,
```
ypise, line 812: Should the default value of secure not be True?

api/python/provisioner/salt.py
```diff
@@ -146,9 +151,11 @@ class SaltRunnerArgs(SaltArgsMixin):
     )
     nowait: bool = False
     kw: Dict = attr.Factory(dict)
+    secure: bool = False
```
ypise, line 154: Should the default value of secure not be True?

ypise (approved):

api/python/provisioner/salt.py
andkononykhin2: No, it would impact many places where we don't need that.

api/python/provisioner/salt.py
andkononykhin2: No, it would impact many places where we don't need that.

PR #681, opened 2020-10-27
EOS 14121: Add support for custom python index repository

SomwanshiAnjali:


SomwanshiAnjali:
@andkononykhin2  @ypise 

By default, custom python repo would be enabled for --dist-type bundle and single iso installation

in other cases we need to provide --disable-pip-repo to enable python installation from pypi.org.

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -417,6 +429,15 @@ def __attrs_post_init__(self):  # noqa: C901
         if self.source == 'local':
             if not self.local_repo:
                 raise ValueError("local repo is undefined")
+            if (
+                not self.disable_pip_repo and
```
andkononykhin2, line 433: As per the comment above that options meaning is not clear. So this and similar conditionals are not clear as well. Please clarify.

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -417,6 +429,15 @@ def __attrs_post_init__(self):  # noqa: C901
...
+                self.dist_type != config.DistrType.BUNDLE
+            ):
+                logger.warning(
+                    "Custom pip repo cannot be set for cortx dist-type "
```
andkononykhin2, line 437: better to say `for non bundled distributions`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -432,6 +453,12 @@ def __attrs_post_init__(self):  # noqa: C901
                         "ISO files for CORTX and CORTX dependencies "
                         f"have the same name: {self.iso_cortx.name}"
                     )
+                if not self.disable_pip_repo:
+                    logger.warning(
+                        "Custom pip repo cannot be set "
```
andkononykhin2, line 458: better to say `Custom python repo can be used only for single ISO based setup`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -483,6 +510,15 @@ def __attrs_post_init__(self):  # noqa: C901
                     "for bundle distribution type"
                 )
                 self.url_cortx_deps = None
+            if (
+                not self.disable_pip_repo and
+                self.dist_type != config.DistrType.BUNDLE
+            ):
+                logger.warning(
+                    "Custom pip repo cannot be set "
```
andkononykhin2, line 518: please use a similar warning message as for local setup

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -303,6 +303,17 @@ class RunArgsSetup:
         # default='github/release/rhel-7.7.1908/last_successful',
         # converter=(lambda v: f'{config.CORTX_REPOS_BASE_URL}/{v}')
     )
+    disable_pip_repo: bool = attr.ib(
```
ypise, line 306: Rename this to `pypi_repo`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -303,6 +303,17 @@ class RunArgsSetup:
...
+        metadata={
+            inputs.METADATA_ARGPARSER: {
+                'help': (
+                    "Disable custom pip repository for "
+                    "pip installations"
```
ypise, line 311: Update help to:
`Use pypi.org as source for pip installations.`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -373,6 +384,7 @@ class RunArgsSetupProvisionerBase:
     url_cortx_deps: str = attr.ib(init=False, default=None)
     dist_type: str = RunArgsSetup.dist_type
     target_build: str = RunArgsSetup.target_build
+    disable_pip_repo: bool = RunArgsSetup.disable_pip_repo
```
ypise, line 387: `pypi_repo`

api/python/provisioner/commands/setup_provisioner.py
ypise, line 433: `pypi_repo`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -432,6 +453,12 @@ def __attrs_post_init__(self):  # noqa: C901
                         "ISO files for CORTX and CORTX dependencies "
                         f"have the same name: {self.iso_cortx.name}"
                     )
+                if not self.disable_pip_repo:
```
ypise, line 456: `if not self.pypi_repo`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -483,6 +510,15 @@ def __attrs_post_init__(self):  # noqa: C901
                     "for bundle distribution type"
                 )
                 self.url_cortx_deps = None
+            if (
+                not self.disable_pip_repo and
```
ypise, line 514: `pypi_repo`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -483,6 +510,15 @@ def __attrs_post_init__(self):  # noqa: C901
...
+                self.dist_type != config.DistrType.BUNDLE
+            ):
```
ypise, line 516: Should this if be modified to:
```
if (
        self.pypi_repo or
        self.dist_type == config.DistrType.BUNDLE
):
```

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -432,6 +453,12 @@ def __attrs_post_init__(self):  # noqa: C901
                         "ISO files for CORTX and CORTX dependencies "
                         f"have the same name: {self.iso_cortx.name}"
                     )
+                if not self.disable_pip_repo:
+                    logger.warning(
+                        "Custom pip repo cannot be set "
+                        "hence pip will be configured to install from pypi.org"
+                    )
+                    self.disable_pip_repo = True
```
ypise, line 461: `self.pypi_repo = True`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1298,6 +1344,10 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
         else:
             ssh_client.state_apply('cortx_repos')
 
+        if not run_args.disable_pip_repo:
```
ypise, line 1347: `if not run_args.pypi_repo:`

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -483,6 +510,15 @@ def __attrs_post_init__(self):  # noqa: C901
...
                 self.url_cortx_deps = None
+            if (
+                not self.disable_pip_repo and
+                self.dist_type != config.DistrType.BUNDLE
+            ):
+                logger.warning(
+                    "Custom pip repo cannot be set "
+                    "hence pip will be configured to install from pypi.org"
+                )
+                self.disable_pip_repo = True
```
ypise, line 521: `self.pypi_repo = True`

ypise (changes requested):

api/python/provisioner/config.py
```diff
@@ -83,7 +84,9 @@
 PRVSNR_CORTX_DEPS_ISO = (
     PRVSNR_CORTX_REPOS_BASE_DIR / f'{CORTX_3RD_PARTY_ISO_DIR}.iso'
 )
-
+PRVNSR_PIP_ISO_DIR = (
+    PRVSNR_CORTX_REPOS_BASE_DIR / f'{CORTX_PIP_ISO_DIR}.iso'
```
andkononykhin2, line 88: as I understand we are not going to have (to support for now)  a standalone python repo ISO assuming only a directory inside a single ISO (either locally or remotely mounted).

So, I think we don't need that for now.

api/python/provisioner/srv/salt/repos/pip_config.sls
```diff
@@ -0,0 +1,29 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+{% set pip_repo_path = pillar['release']['base']['repos']['python_deps']['source'] %}
```
andkononykhin2, line 18: as it was mentioned before: we can add already evaluated value to release pillar so the path would be like that `pillar['release']['python_repo']`

andkononykhin2 (changes requested):
@SomwanshiAnjali  Thank you for the working solution! But I feel some variables naming might confuse. Also pillar management requires an adjustment. Please check my comments.

api/python/provisioner/commands/setup_provisioner.py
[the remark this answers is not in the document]
SomwanshiAnjali: updated parameter to pypi_repo

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -303,6 +303,17 @@ class RunArgsSetup:
         # default='github/release/rhel-7.7.1908/last_successful',
         # converter=(lambda v: f'{config.CORTX_REPOS_BASE_URL}/{v}')
     )
+    disable_pip_repo: bool = attr.ib(
+        metadata={
+            inputs.METADATA_ARGPARSER: {
+                'help': (
+                    "Disable custom pip repository for "
```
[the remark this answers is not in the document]
SomwanshiAnjali: @andkononykhin2 --custom-python-repo would be to explicitly enable custom python repository

As suggested by @ypise  we are will always install from cutom python repo (should be enabled by default) and user will explicitly inform when to use pypi.org hence I have used --pypi-repo 

according to this the validations are added as below
if --pypi-repo not provided
	1. source = local OR rpm and dist-type != bundle
			WARN: using pypi.org
			pypi_repo = True
	2. source = ISO and --iso-cortx-deps 
			WARN: using pypi.org
			pypi_repo = True

And create/update pip.conf if pypi_repo is False

api/python/provisioner/commands/setup_provisioner.py
SomwanshiAnjali: Clarified in above comment
https://github.com/Seagate/cortx-prvsnr/pull/681#discussion_r513377085

api/python/provisioner/commands/setup_provisioner.py
SomwanshiAnjali: @ypise NOT to both with AND operator does not work as expected, hence didn't change.
https://stackoverflow.com/questions/33565519/python-if-not-this-and-not-that/

ypise (approved):

ypise (approved):

ypise (approved):

PR #691, opened 2020-10-29
EOS-13915:Salt log filtering

AshwiniS-Borse:
rsyslog handler, file handler: need to configure to filter out salt logs lower than INFO level
one more file handle: need to route all salt logs from DEBUG level and higher there


api/python/provisioner/log.py
```diff
@@ -60,6 +61,30 @@ def filter(self, record):
         record.prvsnrcmd = self._cmd
         return True
 
+class LogFileFilter(logging.Filter):
```
andkononykhin2, line 64: please don't combine two filters, python handlers [accept](https://docs.python.org/3.6/library/logging.config.html#dictionary-schema-details) list of filters

api/python/provisioner/log.py
```diff
@@ -60,6 +61,30 @@ def filter(self, record):
...
+
+    @cmd.setter
+    def cmd(self, cmd: str):
+        self._cmd = cmd
+
+    def filter(self, record):
+        record.prvsnrcmd = self._cmd
+
+        if ( record.name.startswith("salt.") ) and ( record.levelno < logging.INFO ):
+            return False
```
andkononykhin2, line 80: I believe you may combined that to simplify:
```
return not ( record.name.startswith("salt.") and (record.levelno < logging.INFO) )
```

api/python/provisioner/log.py
```diff
@@ -60,6 +61,30 @@ def filter(self, record):
...
+        return True
+
+class SaltLogFileFilter(logging.Filter):
+    def filter(self, record):
+        if record.name.startswith("salt."):
```
andkononykhin2, line 85: that might be simplified as well

api/python/provisioner/log.py
```diff
@@ -151,13 +176,13 @@ class _ConsoleStreamLogHandler(_LogHandler):
                     converter=(lambda stream: 'ext://sys.{}'.format(stream))
                 )
             return _ConsoleStreamLogHandler
-        elif hname == logfile_handler:
+        elif ( hname == logfile_handler ) or ( hname == saltlogfile_handler ):
```
andkononykhin2, line 179: possible other way of the same
```
hname in (logfile_handler, saltlogfile_handler)
```

api/python/provisioner/log.py
```diff
@@ -151,13 +176,13 @@ class _ConsoleStreamLogHandler(_LogHandler):
...
             @attr.s(auto_attribs=True)
             class _FileLogHandler(_LogHandler):
                 filename: str = attr.ib(
                     metadata={
                         inputs.METADATA_ARGPARSER: {
-                            'help': f"{logfile_handler} handler file path",
+                            'help': f"{saltlogfile_handler} handler file path",
```
andkononykhin2, line 185: you need to support both log files, not only salt's one

api/python/provisioner/log.py
```diff
@@ -167,7 +192,7 @@ class _FileLogHandler(_LogHandler):
                     metadata={
                         inputs.METADATA_ARGPARSER: {
                             'help': (
-                                f"{logfile_handler} handler "
+                                f"{saltlogfile_handler} handler "
```
andkononykhin2, line 195: the same as above

api/python/provisioner/log.py
```diff
@@ -181,7 +206,7 @@ class _FileLogHandler(_LogHandler):
...
```
andkononykhin2, line 209: the same as above

api/python/provisioner/provisioner.conf
```diff
@@ -24,17 +30,26 @@ logging:
       (): provisioner.log.NoErrorSysLogHandler
       facility: local1
       formatter: full
-      filters: [cmd_filter]
+      filters: [log_filter]
```
andkononykhin2, line 33: as I mentioned above, please use multiple filters here instead of a super one

api/python/provisioner/provisioner.conf
```diff
@@ -11,6 +11,12 @@ logging:
     cmd_filter:
       (): provisioner.log.CommandFilter
       cmd: unknown
+    log_filter:
+      (): provisioner.log.LogFileFilter
+      cmd: unknown
+    salt_filter:
+      (): provisioner.log.SaltLogFileFilter
+
```
andkononykhin2, line 19: please either add new lines between other blocks or remove that one

andkononykhin2 (changes requested):
@AshwiniS-Borse Very good PR, thank you! It follows the related task's design very closely. But some changes are needed. Please check and address comments.

Also please update the title to refer the task. e.g. [EOS-12345] ... title ...

api/python/provisioner/log.py
```diff
@@ -61,6 +62,17 @@ def filter(self, record):
         return True
 
 
+class LogFileFilter(logging.Filter):
+    def filter(self, record):
+        return not(record.name.startswith("salt.")
```
andkononykhin2, line 67: I believe flake8 will complain here

andkononykhin2 (approved):
Thank you. Looks good now. Please ensure flake8 passes though.

ypise (approved):

PR #697, opened 2020-10-30
EOS-14470: Flip deploy sequence for CSM-UDS

ypise:


andkononykhin2 (approved):
@ypise shoudn't we adjust an [sw update](https://github.com/Seagate/cortx-prvsnr/blob/b20f2cd76f65583feb4eb4bf217e7176a287f362/api/python/provisioner/commands/__init__.py#L821) logic as well? Possibly it doesn't matter there but only to be aligned with the current changes.

83bhp (approved):

PR #708, opened 2020-11-02
EOS-14658: add verbosity to stonith resources

azheregelya:
This makes fence_ipmilan to log messages to system log about power
status.
Note: --debug-file option represented by debug_file option is more
appropriate here, however actual tests showed that it is working only
along with verbose option and does not prevent logging to syslog.
Also debug_file option does not pass validation and requires --force to
be applied when stonith resource is created. This looks fishy, thus
unreliable.

Signed-off-by: Andrei Zheregelia <andrei.zheregelia@seagate.com>


azheregelya:
@andkononykhin could you please take a look and merge? This tweak may help with stonith issues investigation.

andkononykhin2 (approved):

ypise (approved):

azheregelya:
Hello @ypise , @andkononykhin2. Just want to clarify the process: do I need to perform additional actions to get this merged to cortx-1.0 from pre-cortx-1.0?

azheregelya:
Question above is not relevant anymore. Builds starting 409 have this patch.

PR #710, opened 2020-11-03
Enabling RabbitMQ cluster partition autoheal

thavanathan:


cdeshmukh (approved):
Double confirm with Provisioner / S3/CSM team that they are not adding anything to this rabbitmq.conf file which would break it again

cdeshmukh (changes requested):
@thavanathan please update & resubmit this for latest rabbitmq-3.8.9 version

ypise:
Very old PR. Changes are no more relevant after RMQ version update.

PR #711, opened 2020-11-03
EOS-14875: Generate manifest file as part of boxing.

83bhp:


cli/src/common_utils/lr-serial-number
```diff
@@ -29,10 +29,10 @@ function lr_serial_number {
     else
         if [[ -f /opt/seagate/lr-serial-number ]]; then
             echo "Lyve Rack r1 Serial Number already set."
-            exit 31
+            exit 0
```
ypise, line 32: The non-zero exit here is on purpose.
The caller can decide to neglect it.
The primary purpose of this script is to set the value.

ypise (changes requested):

PR #712, opened 2020-11-03
[EOS-14884] SW Update fixes

andkononykhin2:
- fixes regression in get_result routine
- disables the logic that fails sw update in case the same update release is being asked to be configured: for now it would just warns and removes already configured one

### DIFF
--- a/api/python/provisioner/commands/__init__.py
+++ b/api/python/provisioner/commands/__init__.py
@@ -671,13 +671,21 @@ class SetSWUpdateRepo(Set):
 
             # there is no the same release repo is already active
             if self._is_repo_enabled(f'sw_update_{release}'):
-                raise SWUpdateRepoSourceError(
-                    str(repo.source),
-                    (
-                        f"SW update repository for the release "
-                        f"'{release}' has been already enabled"
-                    )
+                err_msg = (
+                    "SW update repository for the release "
+                    f"'{release}' has been already enabled"
                 )
+                logger.warning(err_msg)
+
+                # TODO IMPROVE later raise and error
+                if False:
+                    raise SWUpdateRepoSourceError(
+                        str(repo.source),
+                        (
+                            f"SW update repository for the release "
+                            f"'{release}' has been already enabled"
+                        )
+                    )
 
             # TODO IMPROVE
             #   - new version is higher then currently installed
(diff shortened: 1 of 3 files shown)


api/python/provisioner/commands/__init__.py
```diff
@@ -671,13 +671,21 @@ def dynamic_validation(self, params: inputs.SWUpdateRepo, targets: str):  # noqa
...
-                        f"'{release}' has been already enabled"
-                    )
+                err_msg = (
+                    "SW update repository for the release "
+                    f"'{release}' has been already enabled"
                 )
+                logger.warning(err_msg)
+
+                # TODO IMPROVE later raise and error
+                if False:
```
ypise, line 681: This is a smart way of TODO :)
But why not just leave the actual code commented?

ypise (approved):

api/python/provisioner/commands/__init__.py
dmitrydb: Why do we need this logic if we decided to allow the re-uploading of the ISO version? My suggestion, we need to decide what functionality we support. Should we support uploading of SW update of the same version or not? Because in the future, it is possible to get a situation when we will have two possible bugs which contradict each other: one for "we need to allow to re-upload the existing one SW update" and the second one for "We should reject re-uploading of ISO with the existing release".

dmitrydb (commented):
Please, consider my few comments for minor fixes

api/python/provisioner/commands/__init__.py
andkononykhin2: For sure we need that logic but it would require a week of so fo integration with CSM team since they should adjust their logic accordingly.

api/python/provisioner/errors.py
```diff
@@ -265,7 +267,8 @@ def __init__(self, reason: ProvisionerError, rollback_error=None):
 
     def __str__(self):
         return (
-            'SSL Cert update failed: {!r}'.format(self)
+            'SSL Cert update failed: reason={}, rollback_error={}'
```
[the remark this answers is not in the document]
andkononykhin2: legacy code, for sure f-strings are better but we hadn't used them for at the beginning. it makes sense to create a separate PR to adjust all such places, not a target of the current PR

api/python/provisioner/salt.py
```diff
@@ -1310,7 +1310,8 @@ def prvsnr_job_result(cls, jid):
                     fun=job.function,
                     fun_args=job.arguments
                 )
-                client_res = SaltClientResult(job.result, cmd_args)
+                cmd_args_view = cmd_args._as_dict()
```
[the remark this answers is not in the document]
andkononykhin2: sure, better to do in a separate PR since the current one is just to fix regressions

andkononykhin2:
@ypise @dmitrydb I am agree with all you suggestions but that PR is just to fix the current issues with SW update and provide CSM some time to adjust their logic to our updated one

andkononykhin2:
https://jts.seagate.com/browse/EOS-14972
https://jts.seagate.com/browse/EOS-14970
were created as a follow-up to the current discussion

SomwanshiAnjali (approved):

PR #713, opened 2020-11-04
EOS-14811: Avoid duplication of cluster_id during auto_deploy

SomwanshiAnjali:

### DIFF
--- a/api/python/provisioner/config.py
+++ b/api/python/provisioner/config.py
@@ -253,7 +253,7 @@ def profile_base_dir(
     setup_name: Optional[str] = 'default'
 ):
     if location is None:
-        location = Path.cwd() / PROFILE_DIR_NAME
+        location = Path('/root') / PROFILE_DIR_NAME
     else:
         location = location.resolve()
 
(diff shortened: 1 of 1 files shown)


api/python/provisioner/config.py
```diff
@@ -229,7 +229,7 @@ def profile_base_dir(
     setup_name: Optional[str] = 'default'
 ):
     if location is None:
-        location = Path.cwd() / PROFILE_DIR_NAME
+        location = Path('/root') / PROFILE_DIR_NAME
```
andkononykhin2, line 232: @SomwanshiAnjali it will block all deployments for non root users. Please use home directory instead: it will respect both cases.

andkononykhin2 (changes requested):

api/python/provisioner/config.py
ypise: User environment variable HOME.

andkononykhin2: I beleive  pathlib can provide a better alternative: https://docs.python.org/3.6/library/pathlib.html#pathlib.Path.home

SomwanshiAnjali: Addressed

andkononykhin2 (approved):

ypise (approved):

PR #720, opened 2020-11-05
EOS-14923: Remove default URLs from commons pillar

83bhp:

### DIFF
--- a/srv/components/defaults.yaml
+++ b/srv/components/defaults.yaml
@@ -38,14 +38,6 @@ tmp_dir: /tmp
 
   {% set base_bundle_url = target_build %}
 
-  # FIXME too complicated command
-  {% if ("RedHat" in grains['os']) and (salt['cmd.shell']('subscription-manager list | grep -m1 -A4 -Pe "Product Name:.*Red Hat Enterprise Linux Server"|grep -Pe "Status:.*Subscribed"')) %}
-    {% set system_repo = '/'.join([base_bundle_url, 'rhel7.7']) %}
-  {% else %}
-    # FIXME EOS-12508 not considered yet ?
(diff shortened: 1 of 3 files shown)


srv/components/defaults.yaml
```diff
@@ -95,7 +69,6 @@ tmp_dir: /tmp
 
 
 # finally give env variables a chance
-{% set commons_repo = salt["environ.get"]('CORTX_COMMONS', commons_repo) %}
```
ypise, line 98: Need this value.

ypise (changes requested):

ypise (approved):

srv/components/defaults.yaml
```diff
@@ -54,25 +46,7 @@ tmp_dir: /tmp
...
-        {"id": "cortx_platform_updates", "url": os_updates_repo}
-      ]
-  %}
-
-  {% if "RedHat" in grains['os'] %}
-    {% set commons_repo = pillar['commons']['cortx_commons']['RedHat'] %}
-  {% else %}
-    {% set commons_repo = pillar['commons']['cortx_commons']['CentOS'] %}
-  {% endif %}
+ {% set common_repo = '/'.join([target_build, '3rd_party']) %}
```
andkononykhin2, line 49: That should never happen since {target_build}/3rd_party means that it is a bundle distribution. And epel and saltstack are part of it.
If we want to drop the support for satellite urls then better just to remove that `# LEGACY REPOS STRUCTURE` block

SomwanshiAnjali (approved):

srv/components/defaults.yaml
ypise: @andkononykhin2 
This needs rework agreed, but for now we need to make minimum changes to code
Let's discuss.

PR #728, opened 2020-11-07
[EOS-14307] Wrapper for cortx-py-utils checks and replace_node checks

mazinamdar:


api/python/provisioner/commands/check.py
```diff
@@ -609,6 +609,106 @@ def _cluster_status(*, args: str) -> CheckEntry:
...
+    def _storage_lvms(*, args: str) -> CheckEntry:
+        """Storage lvms check."""
+        res: CheckEntry = CheckEntry(cfg.Checks.STORAGE_LVMS.value)
+        try:
+            from cortx.utils.kvstore.pillar import PillarStorage
+            from cortx.utils.validator.v_storage import StorageV
+        except ImportError:
+            res.set_fail(checked_target=cfg.ALL_MINIONS,
+                         comment="Package cortx-py-utils not installed")
+            return res
```
andkononykhin2, line 622: newlines might help to make the code slightly better

api/python/provisioner/commands/check.py
```diff
@@ -609,6 +609,106 @@ def _cluster_status(*, args: str) -> CheckEntry:
 
         return res
 
+    @staticmethod
+    def _storage_lvms(*, args: str) -> CheckEntry:
+        """Storage lvms check."""
+        res: CheckEntry = CheckEntry(cfg.Checks.STORAGE_LVMS.value)
+        try:
+            from cortx.utils.kvstore.pillar import PillarStorage
+            from cortx.utils.validator.v_storage import StorageV
```
andkononykhin2, line 618: that might be done at the top of the file:
- better practice
- less duplication

api/python/provisioner/commands/check.py
```diff
@@ -609,6 +609,106 @@ def _cluster_status(*, args: str) -> CheckEntry:
...
+        except ImportError:
+            res.set_fail(checked_target=cfg.ALL_MINIONS,
+                         comment="Package cortx-py-utils not installed")
+            return res
+        try:
+            nodes = PillarStorage().get("cluster:node_list")
```
andkononykhin2, line 624: you can do the same using provisioner API, please do not mess the code

api/python/provisioner/commands/check.py
```diff
@@ -623,7 +723,14 @@ def run(self, check_name: cfg.Checks = None,
         res: CheckResult = CheckResult()
 
         if check_name is None or check_name == cfg.Checks.ALL.value:
-            for check_name in cfg.CHECKS:
+            check_lists = cfg.SWUPDATE_CHECKS
+        else:
+            try:
+                check_lists = getattr(cfg, check_name.strip().upper())
+            except AttributeError:
+                check_lists = None
```
andkononykhin2, line 731: new lines might help here as well

api/python/provisioner/config.py
```diff
@@ -378,10 +378,32 @@ class Checks(Enum):
...
+    Checks.BMC_ACCESSIBILITY.value,
+    Checks.COMMUNICABILITY.value,
+    Checks.CLUSTER_STATUS.value,
+    Checks.LOGS_ARE_GOOD.value,
+    Checks.PASSWORDLESS_SSH_ACCESS.value,
+    Checks.STORAGE_LVMS.value,
+    Checks.STORAGE_LUNS.value,
+    Checks.MGMT_VIP.value,
+    Checks.HOSTNAMES.value,
+    Checks.POST_CHECKS.value,
```
andkononykhin2, line 402: I don't understand the design:
- granular checks are combined with group checks
- swupdate (as a group check) is not listed though
- sw update checks are triggered for ALL value but list for ALL != list for sw update

api/python/provisioner/commands/check.py
```diff
@@ -623,7 +723,14 @@ def run(self, check_name: cfg.Checks = None,
         res: CheckResult = CheckResult()
 
         if check_name is None or check_name == cfg.Checks.ALL.value:
-            for check_name in cfg.CHECKS:
+            check_lists = cfg.SWUPDATE_CHECKS
```
andkononykhin2, line 726: list of ALL != list of sw update, how we can request to do all checks then?

andkononykhin2 (changes requested):

api/python/provisioner/commands/check.py
dmitrydb: I think it was done to avoid import errors in the provisioner if `cortx.utils` is not installed on the system.
So if we can't for that reason(or any other reason) place these imports at the top of the file, we can consider using a dynamic import approach and load `cortx.utils` on the fly when it will be needed.

api/python/provisioner/commands/check.py
```diff
@@ -609,6 +609,106 @@ def _cluster_status(*, args: str) -> CheckEntry:
...
+
+            for node in nodes:
+                check_ent: CheckEntry = CheckEntry(cfg.Checks.HOSTNAMES.value)
+                try:
+                    hostname = PillarStorage().get(f"cluster:{node}:hostname")
+                    NetworkV().validate('connectivity', [hostname])
+                except Exception as exc:
+                    check_ent.set_fail(checked_target=local_minion_id(),
+                                       comment=str(exc))
+                check_ent.set_passed(checked_target=hostname)
```
dmitrydb, line 704: I think this line should be placed in `else`-part of **try/except** block. Otherwise, we will always trigger `set_passed`

api/python/provisioner/commands/check.py
```diff
@@ -609,6 +609,106 @@ def _cluster_status(*, args: str) -> CheckEntry:
...
+                res.append(check_ent)
+
+        except Exception as exc:
+            check_salt.set_fail(checked_target=local_minion_id(),
```
dmitrydb, line 708: We create `check_salt` validation entry but don't add it to the `res` list.

api/python/provisioner/commands/check.py
dmitrydb: Also, it was designed to allow the user running check by its name. For example,
```
provisioner check network
```
So we can allow running all checks by group name (sw_update, post_checks, replacenode_checks). But we need to track that names of granular checks will not intersect with group check names.

api/python/provisioner/commands/check.py
```diff
@@ -623,7 +723,14 @@ def run(self, check_name: cfg.Checks = None,
         res: CheckResult = CheckResult()
 
         if check_name is None or check_name == cfg.Checks.ALL.value:
-            for check_name in cfg.CHECKS:
+            check_lists = cfg.SWUPDATE_CHECKS
+        else:
+            try:
+                check_lists = getattr(cfg, check_name.strip().upper())
```
dmitrydb, line 729: In my opinion, it is better to refer to list checks by its names rather than try to obtain them from `config` module at the top level

api/python/provisioner/config.py
dmitrydb: We can also consider the creation of different classes/types for granular checks and group checks. This will allow us to use `type/isinstance` python API, for example.

dmitrydb (changes requested):
Added some comments and my thoughts

api/python/provisioner/config.py
```diff
@@ -378,10 +378,32 @@ class Checks(Enum):
...
+    STORAGE_LUNS = "storage_luns"
+    MGMT_VIP = "mgmt_vip"
+    HOSTNAMES = "hostnames"
+    POST_CHECKS = "post_checks"
+    REPLACENODE_CHECKS = "replacenode_checks"
 
 
 # Set of supported validations/checks
 CHECKS = {
+    Checks.NETWORK.value,
```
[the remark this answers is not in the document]
mazinamdar: comment is addressed

api/python/provisioner/commands/check.py
mazinamdar: comment is addressed

api/python/provisioner/commands/check.py
mazinamdar: used global variable to handle error and raised exception if actual check is called

api/python/provisioner/commands/check.py
mazinamdar: comment is addressed

api/python/provisioner/commands/check.py
mazinamdar: comment is addressed

api/python/provisioner/config.py
mazinamdar: Simplified design by making 2 different classes for checks:

``` class Checks(Enum):

    """Enumeration for available checks/validations"""

    NETWORK = "network"
    CONNECTIVITY = "connectivity"
    BMC_ACCESSIBILITY = "bmc_accessibility"
    COMMUNICABILITY = "communicability"
    CLUSTER_STATUS = "cluster_status"
    LOGS_ARE_GOOD = "logs_are_good"
    PASSWORDLESS_SSH_ACCESS = "passwordless_ssh_access"
    STORAGE_LVMS = "storage_lvms"
    STORAGE_LUNS = "storage_luns"
    MGMT_VIP = "mgmt_vip"
    HOSTNAMES = "hostnames"


class GroupChecks(Enum):

    """Enum for group checks"""
    ALL = "all"
    POST_CHECKS = "post_checks"
    REPLACENODE_CHECKS = "replacenode_checks"
    SWUPDATE_CHECKS = "swupdate_checks"
```

api/python/provisioner/commands/check.py
mazinamdar: now below command will run all checks available in CHECKS list

``` provisioner check all ```

File: api/python/provisioner/commands/check.py (hunk 7 above)
Comment (mazinamdar) [reply api/python/provisioner/commands/check.py]: comment is addressed

File: api/python/provisioner/commands/check.py (hunk 8 above)
Comment (mazinamdar) [reply api/python/provisioner/commands/check.py:714]: comment is addressed

File: api/python/provisioner/commands/check.py (hunk 9 above)
Comment (mazinamdar) [reply api/python/provisioner/commands/check.py]: Simplified design by making 2 different classes for checks:
Getting directly list of checks from config using getattr

    """Enumeration for available checks/validations"""

    NETWORK = "network"
    CONNECTIVITY = "connectivity"
    BMC_ACCESSIBILITY = "bmc_accessibility"
    COMMUNICABILITY = "communicability"
    CLUSTER_STATUS = "cluster_status"
    LOGS_ARE_GOOD = "logs_are_good"
    PASSWORDLESS_SSH_ACCESS = "passwordless_ssh_access"
    STORAGE_LVMS = "storage_lvms"
    STORAGE_LUNS = "storage_luns"
    MGMT_VIP = "mgmt_vip"
    HOSTNAMES = "hostnames"


class GroupChecks(Enum):

    """Enum for group checks"""
    ALL = "all"
    POST_CHECKS = "post_checks"
    REPLACENODE_CHECKS = "replacenode_checks"
    SWUPDATE_CHECKS = "swupdate_checks"

File: api/python/provisioner/commands/check.py (hunk 2 above)
Comment (andkononykhin2) [reply api/python/provisioner/commands/check.py]: `import_error` is too generic variable name for that python module (not only cortx utils related routine is presented here).
I would suggest to use something more specific one.

File: api/python/provisioner/commands/check.py [hunk 11]
```diff
@@ -613,15 +620,15 @@ def _cluster_status(*, args: str) -> CheckEntry:
     def _storage_lvms(*, args: str) -> CheckEntry:
         """Storage lvms check."""
         res: CheckEntry = CheckEntry(cfg.Checks.STORAGE_LVMS.value)
-        try:
-            from cortx.utils.kvstore.pillar import PillarStorage
-            from cortx.utils.validator.v_storage import StorageV
-        except ImportError:
+
+        global import_error
```
andkononykhin2, line 624: not needed

api/python/provisioner/commands/check.py
mazinamdar: used "cortx_py_utils_import_error" as variable name

api/python/provisioner/commands/check.py
mazinamdar: removed, thanks

andkononykhin2 (approved):

dmitrydb (approved):

api/python/provisioner/commands/__init__.py
```diff
@@ -764,7 +764,7 @@ def run(self, targets):  # noqa: C901 FIXME
 
             checker = Check()
             try:
-                check_res = checker.run()
+                check_res = checker.run("swupdate_checks")
```
dmitrydb, line 767: Please, consider to use the constant for "swupdate_checks" string

dmitrydb (changes requested):

api/python/provisioner/commands/__init__.py
mazinamdar: comment is addressed

dmitrydb (approved):

PR #730, opened 2020-11-09
[EOS-15007] cortxub user configuration as a service user

andkononykhin2:
The logic:

- user is created at setup provisioner stage and the generated password is put into the pillar
  - on the second node the account would be disabled (expired)
- during the deployment stage it should be encrypted
- at boxing time the credentials are dumped as before to log file so one can extract them for the docs
  - (optionally) we may remove the password from pillar here if needed (one line change but I am not sure that we need that)
- at unboxing time the acoount on the secondary node is activated (expiration is set to 'never')

### DIFF
--- a/cli/src/factory_ops/boxing/init
+++ b/cli/src/factory_ops/boxing/init
@@ -109,14 +109,14 @@ else
 fi
 
 echo "INFO: Resolving service user credentials" | tee -a ${LOG_FILE}
-# FIXME these values might be encrypted
 service_user_name=$(salt-call pillar.get system:service_user:name --output=newline_values_only)
-service_user_secret=$(salt-call pillar.get system:service_user:password --output=newline_values_only)
+service_user_secret_encrypted=$(salt-call pillar.get system:service_user:password --output=newline_values_only)
+service_user_secret=$(salt-call lyveutil.decrypt system "$service_user_secret_encrypted" --output=newline_values_only)
 
 if [[ -z "$service_user_name" || -z "$service_user_secret" ]]; then
     msg="ERROR: credentials for a service user are not found"
     echo "$msg" | tee -a ${LOG_FILE}
-    die "$msg"
+    exit 1
 fi
 
 # Shutdown HA
--- a/cli/src/factory_ops/boxing/init
+++ b/cli/src/factory_ops/boxing/init
@@ -108,6 +108,17 @@ else
     cortxcli support_bundle generate manifest -c "manifest" 2>&1 | tee -a ${LOG_FILE}
 fi
 
+echo "INFO: Resolving service user credentials" | tee -a ${LOG_FILE}
+# FIXME these values might be encrypted
+service_user_name=$(salt-call pillar.get system:service_user:name --output=newline_values_only)
+service_user_secret=$(salt-call pillar.get system:service_user:password --output=newline_values_only)
+
+if [[ -z "$service_user_name" || -z "$service_user_secret" ]]; then
+    msg="ERROR: credentials for a service user are not found"
+    echo "$msg" | tee -a ${LOG_FILE}
+    die "$msg"
+fi
+
 # Shutdown HA
 if ! command -v pcs; then
     echo "ERROR: Command 'pcs' not found"
@@ -143,7 +154,7 @@ configure_service_port
 #Remove gluster fs mounts
 remove_gfs_mounts
 # Create user for unboxing
-create_unboxing_user
+dump_unboxing_data "$service_user_name" "$service_user_secret"
 
 # Create flag file
 boxing_flag $box_flag_file
--- a/cli/src/factory_ops/unboxing/init
+++ b/cli/src/factory_ops/unboxing/init
@@ -370,13 +370,13 @@ lock_unboxing_user
 # FIXME the value might be encrypted
 service_user=$(salt srvnode-2 pillar.get system:service_user:name --output=newline_values_only)
 if [[ -n "$service_user" ]]; then
-    echo -n "Activating service user $service_user on the secondary node................." | tee -a ${LOG_FILE}
+    echo -n "Activating service user $service_user on the Server B node................." | tee -a ${LOG_FILE}
     salt "srvnode-2" state.single user.present name="$service_user" expire=-1 ${salt_opts}
     sleep 5     # Mindfulness break
     echo "Ok." | tee -a ${LOG_FILE}
 else
     # backward compatibility
-    echo -n "WARNING: Skipping service user activation on the secondary node, the user is not set" | tee -a ${LOG_FILE}
+    echo -n "WARNING: Skipping service user activation on the Server B node, the user is not set" | tee -a ${LOG_FILE}
 fi
 
 # Unboxing SUCCESS
--- a/cli/src/factory_ops/unboxing/init
+++ b/cli/src/factory_ops/unboxing/init
@@ -367,7 +367,6 @@ echo "Ok." | tee -a ${LOG_FILE}
 # lock unboxing user
 lock_unboxing_user
 
-# FIXME the value might be encrypted
 service_user=$(salt srvnode-2 pillar.get system:service_user:name --output=newline_values_only)
 if [[ -n "$service_user" ]]; then
     echo -n "Activating service user $service_user on the Server B node................." | tee -a ${LOG_FILE}
--- a/cli/src/factory_ops/unboxing/init
+++ b/cli/src/factory_ops/unboxing/init
@@ -367,6 +367,18 @@ echo "Ok." | tee -a ${LOG_FILE}
 # lock unboxing user
 lock_unboxing_user
 
+# FIXME the value might be encrypted
+service_user=$(salt srvnode-2 pillar.get system:service_user:name --output=newline_values_only)
+if [[ -n "$service_user" ]]; then
+    echo -n "Activating service user $service_user on the secondary node................." | tee -a ${LOG_FILE}
+    salt "srvnode-2" state.single user.present name="$service_user" expire=-1 ${salt_opts}
+    sleep 5     # Mindfulness break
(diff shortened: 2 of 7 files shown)


ipoddubnyy (approved):
approved

cli/src/factory_ops/boxing/init
```diff
@@ -95,6 +95,17 @@ fi
 # Prepare shutdown
 ensure_healthy_cluster
 
+echo "INFO: Resolving service user credentials" | tee -a ${LOG_FILE}
+# FIXME these values might be encrypted
+service_user_name=$(salt-call pillar.get system:service_user:name --output=newline_values_only)
```
83bhp, line 100: Please use --local option to salt-call commands.

cli/src/factory_ops/boxing/init
```diff
@@ -95,6 +95,17 @@ fi
...
+service_user_secret=$(salt-call pillar.get system:service_user:password --output=newline_values_only)
```
83bhp, line 101: same here, use --local option

cli/src/factory_ops/boxing/init
```diff
@@ -95,6 +95,17 @@ fi
...
+
+if [[ -z "$service_user_name" || -z "$service_user_secret" ]]; then
+    msg="ERROR: credentials for a service user are not found"
+    echo "$msg" | tee -a ${LOG_FILE}
+    die "$msg"
```
83bhp, line 106: I don't see definition of "die"

pillar/components/system.sls
```diff
@@ -19,3 +19,9 @@ system:
   ntp:
     time_server: time.seagate.com
     timezone: UTC             #To be replaced with input
+  service_user:
+    name: cortxub
+    password:
```
83bhp, line 24: are you deliberately not using "secret" to avoid password encryption?

83bhp: As per my understanding only fields with "secret" are encrypted in lyveutil.encrypt module, @ypise can confirm.

srv/components/system/config/service_user.sls
```diff
@@ -0,0 +1,48 @@
+#
```
83bhp, line 1: Where are we calling this state from?

cli/src/factory_ops/unboxing/init
```diff
@@ -367,6 +367,18 @@ echo "Ok." | tee -a ${LOG_FILE}
 # lock unboxing user
 lock_unboxing_user
 
+# FIXME the value might be encrypted
+service_user=$(salt srvnode-2 pillar.get system:service_user:name --output=newline_values_only)
+if [[ -n "$service_user" ]]; then
+    echo -n "Activating service user $service_user on the secondary node................." | tee -a ${LOG_FILE}
+    salt "srvnode-2" state.single user.present name="$service_user" expire=-1 ${salt_opts}
```
83bhp, line 374: why are we not doing this on node-1?

cli/src/factory_ops/unboxing/init
```diff
@@ -367,6 +367,18 @@ echo "Ok." | tee -a ${LOG_FILE}
 # lock unboxing user
 lock_unboxing_user
 
+# FIXME the value might be encrypted
+service_user=$(salt srvnode-2 pillar.get system:service_user:name --output=newline_values_only)
+if [[ -n "$service_user" ]]; then
+    echo -n "Activating service user $service_user on the secondary node................." | tee -a ${LOG_FILE}
```
83bhp, line 373: Replace "secondary node" with "Server B"

cli/src/factory_ops/unboxing/init
```diff
@@ -367,6 +367,18 @@ echo "Ok." | tee -a ${LOG_FILE}
...
+    salt "srvnode-2" state.single user.present name="$service_user" expire=-1 ${salt_opts}
+    sleep 5     # Mindfulness break
+    echo "Ok." | tee -a ${LOG_FILE}
+else
+    # backward compatibility
+    echo -n "WARNING: Skipping service user activation on the secondary node, the user is not set" | tee -a ${LOG_FILE}
```
83bhp, line 379: Replace "secondary node" with "Server B"

cli/src/factory_ops/boxing/init
83bhp, line 101: Shouldn't we decrypt the password before using it?

83bhp (dismissed):

cli/src/factory_ops/boxing/init
andkononykhin2: how about that one ? https://github.com/Seagate/cortx-prvsnr/blob/e6d15babe156505f810b912ccd0918112cc42788/cli/src/factory_ops/boxing/init#L54

pillar/components/system.sls
andkononykhin2: >are you deliberately not using "secret" to avoid password encryption?
That block of data is related to usual system user params. I guess it's ok to in encryption logic. You may check that: https://github.com/Seagate/cortx-prvsnr/blob/e6d15babe156505f810b912ccd0918112cc42788/srv/_modules/pillar_ops.py#L53

srv/components/system/config/service_user.sls
andkononykhin2: Good catch. I will add the trigger.

cli/src/factory_ops/unboxing/init
andkononykhin2: It is activated there already by [design](https://jts.seagate.com/browse/EOS-15007?focusedCommentId=1976996&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1976996)

cli/src/factory_ops/unboxing/init
```diff
@@ -367,6 +367,18 @@ echo "Ok." | tee -a ${LOG_FILE}
 # lock unboxing user
 lock_unboxing_user
 
+# FIXME the value might be encrypted
+service_user=$(salt srvnode-2 pillar.get system:service_user:name --output=newline_values_only)
```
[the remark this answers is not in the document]
andkononykhin2: I copied from [here](https://github.com/Seagate/cortx-prvsnr/blob/e6d15babe156505f810b912ccd0918112cc42788/cli/src/factory_ops/boxing/prov_tasks#L76). What is the difference?

cli/src/factory_ops/boxing/init
andkononykhin2: I copied from [here](https://github.com/Seagate/cortx-prvsnr/blob/e6d15babe156505f810b912ccd0918112cc42788/cli/src/factory_ops/boxing/prov_tasks#L76). What is the difference?

cli/src/factory_ops/boxing/init
andkononykhin2: I copied from [here](https://github.com/Seagate/cortx-prvsnr/blob/e6d15babe156505f810b912ccd0918112cc42788/cli/src/factory_ops/boxing/prov_tasks#L76). What is the difference?

andkononykhin2: Yes, it's not done yet. I will address.

cli/src/factory_ops/unboxing/init
andkononykhin2: done

cli/src/factory_ops/unboxing/init
andkononykhin2: done

srv/components/system/config/service_user.sls
andkononykhin2: [done](https://github.com/Seagate/cortx-prvsnr/blob/5eb884a11fc622c405fd44a396242bf6e43e16a8/srv/components/system/config/init.sls#L23)

cli/src/factory_ops/unboxing/init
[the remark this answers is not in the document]
83bhp: yes, thanks for pointing, those needs to be updated too (will take care in my next patch).. current implementation isn't harmful as such, it's just that --local ensures that the command works even if salt-master/minion isn't available (We realized it only yesterday), it's a good practice to follow.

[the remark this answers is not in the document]
83bhp: Ok, agree.. I didn't realize it's the pillar.get salt command. salt_opts will redirect the output to unboxing.log file.

cli/src/factory_ops/boxing/init
83bhp: yes, not needed for pillar.get commands. Ignore.

pillar/components/system.sls
83bhp: I see, thanks!

cli/src/factory_ops/unboxing/init
83bhp: ok!

cli/src/factory_ops/boxing/init
andkononykhin2: done

cli/src/factory_ops/unboxing/init
```diff
@@ -367,6 +367,17 @@ echo "Ok." | tee -a ${LOG_FILE}
 # lock unboxing user
 lock_unboxing_user
 
+service_user=$(salt srvnode-2 pillar.get system:service_user:name --output=newline_values_only)
```
ypise, line 370: @andkononykhin2 
Why is `salt srvnode-2` being used instead of `salt-call`?

andkononykhin2: we performing that on node-1 but need to know the configuration of node-2, it's more accurate to refer to accordant node here instead of relaying on the same pillar for both nodes (that might become false in future).
Also the code above uses the salt `salt` calls a lot so I think it's safe to do the same here

83bhp (approved):

PR #733, opened 2020-11-10
[EOS-14140] light pillar set/get using salt-call --local wrapper

andkononykhin2:
- provisioner `pillar_get` a `pillar_set` comands now supports `--local` option that will switch the mode to use `salt-call --local`
- also one more pillar root was added into account `/srv/glusterfs/volume_prvsnr_data/srv/pillar`
- it means that local mode would work for both read and write pillar requests even if salt and gluster services are not running:
  - set commands will update values in `/var/lib/seagate/cortx/provisioner/local/srv/pillar` (it is not shared via gluster)
  - get commands will take into account both gluster brick `/srv/glusterfs/volume_prvsnr_data/srv/pillar` and `/var/lib/seagate/cortx/provisioner/local/srv/pillar`

### DIFF
--- a/api/python/provisioner/config.py
+++ b/api/python/provisioner/config.py
@@ -48,15 +48,32 @@ PRVSNR_DATA_SHARED_DIR = Path('/var/lib/seagate/cortx/provisioner/shared')
 PRVSNR_DATA_LOCAL_DIR = Path('/var/lib/seagate/cortx/provisioner/local')
 
 PRVSNR_USER_SALT_DIR = PRVSNR_DATA_SHARED_DIR / 'srv'
-# PRVSNR_USER_LOCAL_SALT_DIR = PRVSNR_DATA_LOCAL_DIR / 'srv'
+PRVSNR_USER_LOCAL_SALT_DIR = PRVSNR_DATA_LOCAL_DIR / 'srv'
 PRVSNR_FACTORY_PROFILE_DIR = PRVSNR_DATA_SHARED_DIR / 'factory_profile'
 
 # reflects salt-master file_roots configuration
 PRVSNR_USER_FILEROOT_DIR = PRVSNR_USER_SALT_DIR / 'salt'
-# PRVSNR_USER_LOCAL_FILEROOT_DIR = PRVSNR_USER_LOCAL_SALT_DIR / 'salt'
+PRVSNR_USER_LOCAL_FILEROOT_DIR = PRVSNR_USER_LOCAL_SALT_DIR / 'salt'
 # reflects pillar/top.sls
 PRVSNR_USER_PILLAR_DIR = PRVSNR_USER_SALT_DIR / 'pillar'
-# PRVSNR_USER_LOCAL_PILLAR_DIR = PRVSNR_USER_LOCAL_SALT_DIR / 'pillar'
+PRVSNR_USER_LOCAL_PILLAR_DIR = PRVSNR_USER_LOCAL_SALT_DIR / 'pillar'
+
+# Notes:
+# 1. how salt's pillar roots organized:
+#   - salt searches for files in each root
+#   - if a file (with the same path, path is relative to pillar root)
+#     has been found in one of the previous roots - it is ignored
+#   - all found files are sorted alphabetically
+#   - then merge logic happens: later files win
+#     (e.g. file zzz.sls wins aaa.sls)
+# 2. why we nee prefixes:
+#   - we don't want the same pillar file paths in different roots
+#   - we need them to be sorted according to priorities we have in
+#     salt master/minion config files
+#     (e.g. default ones have less priority than user ones,
+#           shared user pillar has less priority than local one)
+PRVSNR_USER_PILLAR_PREFIX = 'uu_'
+PRVSNR_USER_LOCAL_PILLAR_PREFIX = 'zzz_'
 
 PRVSNR_PILLAR_CONFIG_INI = str(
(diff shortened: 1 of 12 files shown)


api/python/provisioner/salt.py
```diff
@@ -410,6 +425,36 @@ def _get_state_fails(self, ret: Dict):
...
+            }
+        }
+        super().__attrs_post_init__()
+
+    def _parse_raw_dict(self):
+        ret = self.raw['local']['ret']
+
+        if (
+            self.cmd_args_view['fun'].startswith('state.')
+            and (type(ret) is dict)
```
dmitrydb, line 447: Based on PEP8:
https://www.flake8rules.com/rules/W503.html
The best practice it is place binary operator `and` before line break:
```
if (
self.cmd_args_view['fun'].startswith('state.') and
(type(ret) is dict)
}
```

I also prefer to do the additional indent for the second and following lines with conditions
```
        if (self.cmd_args_view['fun'].startswith('state.') and 
               (type(ret) is dict)):
```
From my point of view, such way is more readable and widely used in other projects. But it is not required by PEP8.

dmitrydb (changes requested):
Added some minor comments

api/python/provisioner/paths.py
```diff
@@ -0,0 +1,66 @@
...
+default_pillar = PillarPath(config.PRVSNR_PILLAR_DIR)
+
+user_shared_pillar = PillarPath(
+    config.PRVSNR_USER_PILLAR_DIR,
+    config.PRVSNR_USER_PILLAR_PREFIX
+)
+
+user_local_pillar = PillarPath(
+    config.PRVSNR_USER_LOCAL_PILLAR_DIR,
+    config.PRVSNR_USER_LOCAL_PILLAR_PREFIX
```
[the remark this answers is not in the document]
andkononykhin2: updated

api/python/provisioner/salt.py
andkononykhin2: Sorry, neither I nor pep8 can agree with you here. Please check the link you provided (it's not a pep8 but how flake8 interprets it which is a different thing and looks outdated) and the [link](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) which is referred there more carefully:
```
In Python code, it is permissible to break before or after a binary operator, as long as the convention is consistent locally. For new code Knuth's style is suggested.
```
and
```
Donald Knuth explains the traditional rule in his Computers and Typesetting series: "Although formulas within a paragraph always break after binary operations and relations, displayed formulas always break before binary operations"
```
We don't have a good habit in our code for the moment but generally I believe the option of "before operator" is presented much and preffered.

That PR is not intended to make a style clean up for the whole repo and is quite aligned with the current code base. If you want to change something globally please raise a discussion inside a team and then a separate PR might be created.

dmitrydb: I agree that it is discussable. So my main message here is even bad code guidelines are better when all follow them. About pep8, yes I investigated it before my original post. You can also see that even pep8 mesh different ways:
https://www.python.org/dev/peps/pep-0008/#code-lay-out
```
# No extra indentation.
if (this_is_one_thing and
    that_is_another_thing):
    do_something()

# Add a comment, which will provide some distinction in editors
# supporting syntax highlighting.
if (this_is_one_thing and
    that_is_another_thing):
    # Since both conditions are true, we can frobnicate.
    do_something()

# Add some extra indentation on the conditional continuation line.
if (this_is_one_thing
        and that_is_another_thing):
    do_something()
```
Also need to remember that flake8 was widely used in Seagate to check PEP8 guidelines before codacy. And flake8 treats them as warnings which we in CSM/Provisioner fixed.
But as you mention, the new approach for line breaks appeared as the best practice in the last few years. So as result we need to choose one method and use it always.

andkononykhin2: >Also need to remember that flake8 was widely used in Seagate to check PEP8 guidelines before codacy. And flake8 treats them as warnings which we in CSM/Provisioner fixed.

I had checked the PR using flake8 before requested for a review and I didn't see any warnings, not sure why are you mentioning codacy here.

So to summarize:
- I don't see here any distraction with neither recommended way (pep8) nor existent codebase
- flake8 check is passed
- broad changes are out of the PR's scope

dmitrydb (approved):

api/python/provisioner/config.py
```diff
@@ -48,15 +48,32 @@
...
+#   - then merge logic happens: later files win
+#     (e.g. file zzz.sls wins aaa.sls)
+# 2. why we nee prefixes:
+#   - we don't want the same pillar file paths in different roots
+#   - we need them to be sorted according to priorities we have in
+#     salt master/minion config files
+#     (e.g. default ones have less priority than user ones,
+#           shared user pillar has less priority than local one)
+PRVSNR_USER_PILLAR_PREFIX = 'uu_'
+PRVSNR_USER_LOCAL_PILLAR_PREFIX = 'zzz_'
```
ypise, line 76: @andkononykhin2 
Could we prefix it with `local_` or `lcl` instead for better understanding?

andkononykhin2: please check the comment above: `zzz` here is to push these files to the end of merging list (sorted alphabetically) which salt builds during search in pillar roots

ypise: Got it.

ypise (approved):

PR #751, opened 2020-11-19
EOS-15156: Add check to remove only remote build url

SomwanshiAnjali:

### DIFF
--- a/cli/src/cortx_prep.sh
+++ b/cli/src/cortx_prep.sh
@@ -44,7 +44,7 @@ function install_prvsnr() {
         mount -t iso9660 /opt/isos/${cortx_iso} /tmp/iso_mount/cortx 2>&1 | tee -a ${LOG_FILE}
         mount -t iso9660 /opt/isos/${os_iso} /tmp/iso_mount/cortx-os 2>&1 | tee -a ${LOG_FILE}
 
-        echo "INFO: Backing up exisitng repositories" 2>&1 | tee -a ${LOG_FILE}
+        echo "INFO: Backing up exisitng repositories" 2>&1 | tee -a "${LOG_FILE}"
         mv /etc/yum.repos.d /etc/yum.repos.d.bak
 
         echo "INFO: Creating bootstrap.repo" 2>&1 | tee -a ${LOG_FILE}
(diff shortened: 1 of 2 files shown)


cli/src/cortx_prep.sh
```diff
@@ -44,8 +44,11 @@ function install_prvsnr() {
         mount -t iso9660 /opt/isos/${cortx_iso} /tmp/iso_mount/cortx 2>&1 | tee -a ${LOG_FILE}
         mount -t iso9660 /opt/isos/${os_iso} /tmp/iso_mount/cortx-os 2>&1 | tee -a ${LOG_FILE}
 
+        echo "INFO: Backing up exisitng repositories" 2>&1 | tee -a ${LOG_FILE}
+        mv /etc/yum.repos.d /etc/yum.repos.d.bak
```
andkononykhin2, line 48: @SomwanshiAnjali  It's not only backing up but also a cleanup. Why have we decided to clean existent repos here?

SomwanshiAnjali: @andkononykhin2  
This change was asked by @ypise, Some systemd packages are getting installed as dependencies from the existent repos with unexpected versions hence we are removing them
@ypise  could you please explain more if I have missed anything

ypise: @andkononykhin2 this is particularly for lab environment where we might get additional repos like redhat.repo (even on CentOS).
To ensure we have complete control during deployment and avoid any unexpected versions being installed, I proposed cleaning up the directory.

cli/src/cortx_prep.sh
```diff
@@ -44,8 +44,11 @@ function install_prvsnr() {
         mount -t iso9660 /opt/isos/${cortx_iso} /tmp/iso_mount/cortx 2>&1 | tee -a ${LOG_FILE}
         mount -t iso9660 /opt/isos/${os_iso} /tmp/iso_mount/cortx-os 2>&1 | tee -a ${LOG_FILE}
 
+        echo "INFO: Backing up exisitng repositories" 2>&1 | tee -a ${LOG_FILE}
```
ypise, line 47: Update this line to:
`echo "INFO: Backing up exisitng repositories" 2>&1 | tee -a "${LOG_FILE}"`

To fix Codacy error.

ypise (approved):

cli/src/cortx_prep.sh
SomwanshiAnjali: Fixed

PR #754, opened 2020-11-20
EOS-15160: start haproxy after csm config in unboxing

SomwanshiAnjali:

### DIFF
--- a/cli/src/factory_ops/unboxing/init
+++ b/cli/src/factory_ops/unboxing/init
@@ -310,7 +310,7 @@ sleep 2
 
 echo "Bringing HCTL cluster out of maintenance mode" 2>&1 | tee -a ${LOG_FILE}
 hctl node unmaintenance --all 2>&1 | tee -a ${LOG_FILE}
-sleep 15
+sleep 120
 
 # Update SSPL init
 echo -n "Configuring Cortx RAS services on server A.................................." | tee -a ${LOG_FILE}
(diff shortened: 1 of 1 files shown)


cli/src/factory_ops/unboxing/init
```diff
@@ -307,7 +307,7 @@ sleep 2
 
 echo "Bringing HCTL cluster out of maintenance mode" 2>&1 | tee -a ${LOG_FILE}
 hctl node unmaintenance --all 2>&1 | tee -a ${LOG_FILE}
-sleep 15
+sleep 120
```
andkononykhin2, line 311: shouldn't we check that cluster is ready and healthy as we do that in [sw update logic](https://github.com/Seagate/cortx-prvsnr/blob/54f331dfc053eb83db74f3dea7115ca7b2614899/api/python/provisioner/commands/__init__.py#L813-L831), in particular we call:
`ensure_healthy_cluster` from [utility_scripts.sh](https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/cli/src/common_utils/utility_scripts.sh#L50)

SomwanshiAnjali: @andkononykhin2  yes we should check but I was not aware of this function
can we do it in next PR as improvement?

andkononykhin2: sure, please add a TODO then since `sleep 120` will likely won't work for all cases

andkononykhin2 (approved):

ypise (approved):

PR #755, opened 2020-11-21
EOS-15197: Improve profile dir reference picker logic

ypise:

### DIFF
--- a/api/python/provisioner/commands/replace_node.py
+++ b/api/python/provisioner/commands/replace_node.py
@@ -16,6 +16,7 @@
 #
 
 import logging
+import os
 
 from .. import (
     config,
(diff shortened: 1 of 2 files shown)


api/python/provisioner/commands/replace_node.py
```diff
@@ -123,7 +124,7 @@ def run(self, **kwargs):
             run_subprocess_cmd(['unalias', 'cp'])
 
         factory_profile = config.PRVSNR_FACTORY_PROFILE_DIR
-        if not factory_profile.is_dir():
+        if (not factory_profile.is_dir()) or (len(os.listdir())):
```
andkononykhin2, line 127: @ypise [os.listdir](https://docs.python.org/3.6/library/os.html#os.listdir) by default lists the current directory. I don't think that it is what we need here. Also it is not good to mix os and pathlib. I guess something like the following would be better:
```
    if not (factory_profile.is_dir() and list(factory_profile.glob('*'))):
```

andkononykhin2 (changes requested):

api/python/provisioner/commands/replace_node.py
ypise: Addressed.

andkononykhin2 (approved):

PR #762, opened 2020-11-24
EOS-14135: add unboxing pre-flight checks

sanjognaik:
- Added below pre-checks in unboxing
    - IP addresses for public data interfaces are set
    - BMC accessibility from both servers
    - Server-A and Server-B can access (ping) Controller-A and
      Controller-B on the enclosure

Signed-off-by: Sanjog Vikram Naik <sanjog.naik@seagate.com>

### DIFF
--- a/cli/src/factory_ops/unboxing/system_check
+++ b/cli/src/factory_ops/unboxing/system_check
@@ -19,15 +19,15 @@
 
 
 # Functions in this file address following:
-#   1. Check boxing flag file and proceed only if boxing.<timestamp> file is present 
-#       /opt/seagate/cortx/provisioner/generated_configs/boxing.<timestamp> 
-#   2. Check if IP is available on management network interface on node. Wait till one is obtained. 
-#   3. Check if hostname is available on node. Wait till one is obtained. 
-#   4. Ping across nodes over direct network to check connectivity and fetch peer hostname. Wait till success. 
-#       From node-1 
-#       $ ping -c1 -I<public_data_ip> srvnode-2 
-#       From node-2 
-#       $ ping -c1 -I<public_data_ip> srvnode-1 
+#   1. Check boxing flag file and proceed only if boxing.<timestamp> file is present
+#       /opt/seagate/cortx/provisioner/generated_configs/boxing.<timestamp>
+#   2. Check if IP is available on management network interface on node. Wait till one is obtained.
+#   3. Check if hostname is available on node. Wait till one is obtained.
+#   4. Ping across nodes over direct network to check connectivity and fetch peer hostname. Wait till success.
+#       From node-1
+#       $ ping -c1 -I<public_data_ip> srvnode-2
+#       From node-2
+#       $ ping -c1 -I<public_data_ip> srvnode-1
 # NOTE: pick value of public_data_ip from cluster.sls
 set -euE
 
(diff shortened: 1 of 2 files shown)


cli/src/factory_ops/unboxing/system_check
```diff
@@ -144,3 +144,45 @@ function check_mgmt_ip {
...
+        echo "Ok." | tee -a ${LOG_FILE}
+    else
+        echo -n "ERROR: Command '${prvsnr_chk_cmd}' Failed. " | tee -a ${LOG_FILE}
+        echo ${fail_msg} | tee -a ${LOG_FILE}
+        exit 1
+    fi
+}
+
+function check_public_data_interfaces {
+    echo -n "Checking IP addresses for public data interfaces are set........" | tee -a ${LOG_FILE}
```
83bhp, line 176: Checking **if** IP .....

83bhp (approved):
Approving with minor comment, rest looks good.

cli/src/factory_ops/unboxing/system_check
sanjognaik: Done

cli/src/factory_ops/unboxing/system_check
```diff
@@ -146,3 +146,45 @@ function check_mgmt_ip {
...
+#       - The provisioner repo is installed.
+#
+#   Args:
+#       check: check to be performed. The checks are listed in Checks class
+#              in api/python/provisioner/config.py
+function prvsnr_check {
+    # Execute 'provisioner check' cli command with mentioned check
+    local prvsnr_chk_cmd="provisioner check $1"
+    # Perform the check and get the output. Check for any failures.
+    local fail_msg=$(${prvsnr_chk_cmd} | grep "fail")
```
andkononykhin2, line 164: @sanjognaik do you think grep'ed result would be enough to get the context of the failure?

andkononykhin2: Also I think it's a bit dangerous to grep such quite generic word (`fail`) in case we have logging into console. I wold suggest to consider to turn off console logging during that check:

- either explicitely using `--noconsole`
- or switching to machine output mode (e.g. using `--out json`)

andkononykhin2 (approved):
@sanjognaik Generally I am ok with the change. But please think about my comments. Makes sense to play with mentioned options and decided if they can harden the solution.

cli/src/factory_ops/unboxing/system_check
sanjognaik: As suggested, updated failure string to "validation failed" instead of just "fail".

andkononykhin2: @sanjognaik I see your changes and they make the logic much better. Thanks. But I've explored the Check logic deeper and see that new variant has few drawbacks (sorry, I might had pushed your thoughts in a wrong way):
- you left [PASSED](https://github.com/Seagate/cortx-prvsnr/blob/8782a812099c37f10a479fb98433e4824bd00396/api/python/provisioner/config.py#L465-L466) for a success result so it's not consistent
- but actually PASSED/FAILED are very good and we shouldn't adjust that just to make grepping simpler

As I understand for the moment direct call to Check command won't fail (it always exits with 0 code) even if some checks have failed so we can't rely on exit code here which is not good (@dmitrydb  @mazinamdar  @divya-vijay FYI).

So, for the moment I think it's better to revert you last commit 8782a81 and merge the PR as-is.

For the future improvements:
- rely on exit code rather than output
  - if the latter is not 0 then treat as a validation failure and provide the output to get a user the details
- grep `ValidationError` which is [coming](https://github.com/Seagate/cortx-prvsnr/blob/922b15c09dc74baf354fd6effa93136f34419701/api/python/provisioner/errors.py#L135) so you can differentiate real validation failure and any other kind of failures 

@dmitrydb  @mazinamdar  @divya-vijay I believe we should consider to make possible raising ValidationError right inside `Check.run`, and I guess it would be better to enable that by default. In scope of  sw update, deloyment etc. that fail might be postponed or corrected.

sanjognaik: Done

PR #764, opened 2020-11-24
EOS-14136: DecisionMaker class for factory deployment

divya-vijay:
1. Added both pre-deploy and post-deploy checks. Works as expected (positive and exception scenarios). 
2. For both validations, raising Exception and breaking the flow right there.

### DIFF
--- a/api/python/provisioner/commands/auto_deploy.py
+++ b/api/python/provisioner/commands/auto_deploy.py
@@ -36,7 +36,8 @@ from .setup_provisioner import (
 )
 from .check import (
     Check,
-    DeploymentDecisionMaker
+    PreChecksDecisionMaker,
+    PostChecksDecisionMaker
 )
 from . import deploy_dual
 
@@ -54,7 +55,6 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
 
     def deployment_validations(self, deploy_check):
         checker = Check()
-        decision_maker = DeploymentDecisionMaker()
 
         try:
             check_res = checker.run(deploy_check)
@@ -62,7 +62,11 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
             raise ValueError("Error During Deployment "
                             f"{deploy_check} Validations: {str(exc)}")
         else:
-            decision_maker.make_decision(check_result=check_res)
+            if "pre" in deploy_check:
+                PreChecksDecisionMaker().make_decision(check_result=check_res)
+
+            else:
+                PostChecksDecisionMaker().make_decision(check_result=check_res)
 
     def run(self, nodes, **kwargs):
         setup_provisioner_args = {
@@ -97,9 +101,7 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
             )
 
         logger.info("Deployment Pre-Flight Validations")
-
-        self.deployment_validations(GroupChecks.PRE_CHECKS.value)
-        logger.info("DEPLOYMENT PRE-FLIGHT CHECKS DONE")
+        self.deployment_validations(GroupChecks.DEPLOY_PRE_CHECKS.value)
 
         logger.info("Deploy")
         deploy_dual.DeployDual(setup_ctx=setup_ctx).run(
@@ -107,8 +109,6 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
         )
 
         logger.info("Post-Deployment Validations")
-
-        self.deployment_validations(GroupChecks.POST_CHECKS.value)
-        logger.info("DEPLOYMENT POST-CHECKS DONE")
+        self.deployment_validations(GroupChecks.DEPLOY_POST_CHECKS.value)
 
         logger.info("Done")
--- a/api/python/provisioner/commands/auto_deploy.py
+++ b/api/python/provisioner/commands/auto_deploy.py
@@ -18,10 +18,12 @@
 from typing import Type
 import logging
 
-from .. import (
-    config,
-    inputs
+from ..config import (
+    PRVSNR_PILLAR_CONFIG_INI,
+    GroupChecks
 )
+from .. import inputs
+
 from ..vendor import attr
 
 from . import (
@@ -31,6 +33,10 @@ from .setup_provisioner import (
     RunArgsSetupProvisionerGeneric,
     SetupProvisioner,
     SetupCmdBase
+)
+from .check import (
+    Check,
+    DeploymentDecisionMaker
 )
 from . import deploy_dual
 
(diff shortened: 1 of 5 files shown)


api/python/provisioner/commands/check.py
```diff
@@ -269,6 +269,42 @@ def make_decision(self, check_result: CheckResult):
         :return:
         """
 
+class DeploymentDecisionMaker(DecisionMaker):
+
+    """Class analyses `CheckResult` structure and will decide
+       to continue or to stop Factory Deployment routine
+    """
+
+    def _check_critical_errors(self, check_result: CheckResult):
```
andkononykhin2, line 278: @divya-vijay  why do we need that copy-paste here?
@dmitrydb As long as the method is not used  I think we either need to move it to the base class or remove at all. I believe it (`ust as example that `DecisionMaker` child classes can threat some
        errors as critical to raise appropriate Exceptio`) is a kind of clear thing and don't need to be emphasized in a such a way.

api/python/provisioner/commands/check.py
```diff
@@ -269,6 +269,42 @@ def make_decision(self, check_result: CheckResult):
...
+    def make_decision(self, check_result: CheckResult):
+        """
+        Make a decision for Factory Deployment
+        based on `CheckResult` analysis
+
+        :param CheckResult check_result: instance with all checks needed for
+                                         to make a decision
+        :return:
+        """
+        if check_result.is_failed:
```
andkononykhin2, line 300: looks like that logic is becoming quite useful in multiple subclasses, makes sense to move the common part to the case class and leave only differences in the child classes

api/python/provisioner/commands/check.py
```diff
@@ -269,6 +269,42 @@ def make_decision(self, check_result: CheckResult):
...
+            failed = "; ".join(str(check)
+                               for check in check_result.get_failed())
+            raise ValueError("ONE OR MORE FACTORY DEPLOYMENT "
```
andkononykhin2, line 303: please use more specific error here (create if needed), e.g. add some `class ValidationError(ProvisionerError, ValueError)` to errors.py

api/python/provisioner/commands/auto_deploy.py
```diff
@@ -77,9 +96,19 @@ def run(self, nodes, **kwargs):
                 ), targets=setup_ctx.run_args.primary.minion_id
             )
 
+        logger.info("Deployment Pre-Flight Validations")
+
+        self.deployment_validations(GroupChecks.PRE_CHECKS.value)
```
andkononykhin2, line 101: why we use something named `PRE_CHECKS`: is it deployment specific or makes sense to check for any routine (swupdate, unboxing)

api/python/provisioner/commands/auto_deploy.py
```diff
@@ -77,9 +96,19 @@ def run(self, nodes, **kwargs):
...
+        logger.info("DEPLOYMENT PRE-FLIGHT CHECKS DONE")
+
         logger.info("Deploy")
         deploy_dual.DeployDual(setup_ctx=setup_ctx).run(
             **deploy_args
         )
 
+        logger.info("Post-Deployment Validations")
+
+        self.deployment_validations(GroupChecks.POST_CHECKS.value)
```
andkononykhin2, line 111: the same question as for `PRE_CHECKS`

api/python/provisioner/commands/auto_deploy.py
```diff
@@ -77,9 +96,19 @@ def run(self, nodes, **kwargs):
                 ), targets=setup_ctx.run_args.primary.minion_id
             )
 
+        logger.info("Deployment Pre-Flight Validations")
+
+        self.deployment_validations(GroupChecks.PRE_CHECKS.value)
+        logger.info("DEPLOYMENT PRE-FLIGHT CHECKS DONE")
```
andkononykhin2, line 102: I don't think we need that log or we should add Done for all other logged staged. But it would be an unnecessary noise in logs.

api/python/provisioner/commands/auto_deploy.py
```diff
@@ -77,9 +96,19 @@ def run(self, nodes, **kwargs):
...
+
         logger.info("Deploy")
         deploy_dual.DeployDual(setup_ctx=setup_ctx).run(
             **deploy_args
         )
 
+        logger.info("Post-Deployment Validations")
+
+        self.deployment_validations(GroupChecks.POST_CHECKS.value)
+        logger.info("DEPLOYMENT POST-CHECKS DONE")
```
andkononykhin2, line 112: I don't think we need that log

andkononykhin2 (commented):
@divya-vijay Thank you. I dont' see any blocker issue in the PR. But I posted few concerns regarding code organization and namings. Please check.

api/python/provisioner/commands/auto_deploy.py
divya-vijay: Initially, when started out with Deployment checks, we named them pre_checks and post_checks. 
Now updated. Comment addressed.

api/python/provisioner/commands/auto_deploy.py
divya-vijay: Comment addressed.

api/python/provisioner/commands/auto_deploy.py
divya-vijay: Comment addressed.

api/python/provisioner/commands/auto_deploy.py
divya-vijay: Just to make sure and test the right set of checks is completed. Deleted now.

api/python/provisioner/commands/check.py
divya-vijay: Comment addressed.

api/python/provisioner/commands/check.py
divya-vijay: We're looking at making this implementation better by handling critical checks in DM class. 
Since for GA, we had decided to let the user run CLI for post-checks, this was not needed. Now, working on handling the checks automatically, so this might be needed next.

mazinamdar (approved):
Looks good to me

api/python/provisioner/errors.py
```diff
@@ -132,6 +132,16 @@ def __str__(self):
         return repr(self.reason)
 
 
+class ValidationError(ProvisionerError, ValueError):
```
andkononykhin2, line 135: I think my previous suggestion wasn't good enough: we don't need ValueError here since generally an error might not be related to wrong values. Please keep only ProvisionerError as a base class. Thanks

andkononykhin2 (changes requested):
@divya-vijay  Please check one missed place: I guess you pasted something wronlgy and it breaks the syntax.

Also one general thing: I see you re-define functions in child classes for the only difference in output strings. I feel it's an overhead. You may postpone further changes for that but I suggest you at least to think about better design here.

api/python/provisioner/errors.py
divya-vijay: Comment addressed.

divya-vijay:
> @divya-vijay Please check one missed place: I guess you pasted something wronlgy and it breaks the syntax.
> 
> Also one general thing: I see you re-define functions in child classes for the only difference in output strings. I feel it's an overhead. You may postpone further changes for that but I suggest you at least to think about better design here.

@andkononykhin the `else` was some rebase mishap. Removed it now. 
For Design:  For now, we have one DecisionMaking class for all PreChecks and PostChecks each. 
In case of error in validations, PreChecks will raise Exceptions, while PostChecks will log Warning (the output strings in both the `make_decision` functions are different). This is common for all the modules.

api/python/provisioner/commands/check.py
```diff
@@ -1073,6 +1114,13 @@ def run(self, check_name: str = "",
         else:
             raise ValueError(f'Check "{check_name}" is not supported')
 
+        # Check for supported check name
```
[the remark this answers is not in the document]
divya-vijay: yes, during rebase. Comment addressed now.

andkononykhin2 (approved):

PR #766, opened 2020-11-25
EOS-15292: Updating vips failed during Unboxing in DHCP environment

ypise:


83bhp (approved):
If removing srvnode-1/2 from "hostname: srvnode-1/2" is NOT absolutely required I would suggest to have them there as it's harmless as of now. Rest all looks good.

PR #769, opened 2020-11-27
Improvise dhclient command and add HW addr in management interface

ypise:

### DIFF
--- a/srv/components/system/network/prepare.sls
+++ b/srv/components/system/network/prepare.sls
@@ -15,14 +15,14 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-{% set nwmanager_is_running = salt['service.status']('NetworkManager') %}
+{#% set nwmanager_is_running = salt['service.status']('NetworkManager') %#}
 
-{% if nwmanager_is_running %}
+{#% if nwmanager_is_running %#}
 
-Stop and disable NetworkManager service:
-  service.dead:
-    - name: NetworkManager
-    - enable: False
+# Stop and disable NetworkManager service:
+#   service.dead:
+#     - name: NetworkManager
+#     - enable: False
 
 # Remove NetworkManager package:
 #   pkg.purged:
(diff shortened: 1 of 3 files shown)


srv/components/system/network/prepare.sls
```diff
@@ -43,18 +43,20 @@ Stop and disable NetworkManager service:
 # If not killed explicitly, it causes network restart to fail: COSTOR-439
 Kill dhclient:
   cmd.run:
-    - name: pkill -SIGTERM dhclient
+    # - name: pkill -SIGTERM dhclient
+    - name: dhclient -x
```
andkononykhin2, line 47: @ypise  as long as we will start dhclient with particular iface don't you think we need to specify iface here as well (not sure how self kill relates with that actually) ?

ypise: We probably don't need an interface here, but I'm wondering what happens if there are one than one processes running.

ypise: Agreed, we need to only change the value for one interface, but it doesn't work.
So we need to go back to pkill.
I'll prefer to disable the kill and restart logic as it doesn't make sense anymore.

srv/components/system/network/prepare.sls
```diff
@@ -41,28 +41,30 @@ Stop and disable NetworkManager service:
...
+# network_manager_is_not_running:
+#   test.nop: []
 
-{% endif %}
+{#% endif %#}
 
-# Dummy placeholder for network.prepare:
-#   test.show_notification:
-#     - text: "A yaml file with comments results in minion non-zero exit"
+Dummy placeholder for network.prepare:
```
83bhp, line 68: Instead we can skip running prepare state from init.

83bhp (approved):
approving with minor comment, rest looks ok.

srv/components/system/network/prepare.sls
ypise: This is necessary as there are multiple calls internally and there's possibility of direct call somewhere.
But, I've commented it in init.sls as well

PR #770, opened 2020-11-27
adds dns validation step after dhclient restart

andkononykhin2:
- adds limited awaiting for DNS to set up after dhclient restart

### DIFF
--- a/srv/components/system/network/files/dns_check.sh
+++ b/srv/components/system/network/files/dns_check.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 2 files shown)


srv/components/system/network/prepare.sls
```diff
@@ -19,6 +19,8 @@
 
 {% if nwmanager_is_running %}
 
+    {% set hostname_short = salt['cmd.run']('hostname -s') %}
```
ypise, line 22: Should be `cmd.shell` if we are capturing the value.

srv/components/system/network/files/dns_check.sh
```diff
@@ -0,0 +1,45 @@
...
+#
+
+
+set -eu
+
+domain="${1:-$(hostname)}"
+
+echo "Checking DNS resolution for domain '$domain'"
+
+ns_servers=$(dig +short NS "$domain" | paste -sd " " -)
```
ypise, line 26: Should we check if dig is installed?

srv/components/system/network/prepare.sls
andkononykhin2: cmd.run for worked the same

srv/components/system/network/files/dns_check.sh
andkononykhin2: bind-utils provides it and we do install them [here](https://github.com/Seagate/cortx-prvsnr/blob/72d5c1e2a8375de6f8548791a4123770ac17bba8/srv/components/system/install.sls#L23)

andkononykhin2: resolve that as `getent` now is used instead

ypise (approved):

PR #772, opened 2020-11-27
EOS-15322: Add improvements for provisioner check API: get by name, a…

dmitrydb:
Add some improvements for provisioner checks API:
- get checks by name
- api to support critical validation checks
- common implementation of `DecisionMaker` class


divya-vijay:
Hello @dmitrydb . I see the PR base branch is set to `pre-cortx-1.0` and not `pre-cortx-1.0-validation` . Any particular reason? Can you please change it so we can see your changes and merge accordingly.

api/python/provisioner/commands/check.py
```diff
@@ -137,6 +137,37 @@ def is_set(self) -> bool:
         """
         return self._verdict is not None
 
+    @property
+    def is_critical(self) -> bool:
+        """
+        Return True if check result is marked as critical and False otherwise
+        """
+        return bool(self._critical)
```
andkononykhin2, line 145: I guess that conversion is required since `None` is possible. But it means that `None` would be presented as non critical. Don't you think that we can simplify that and use `False` as default instead.

api/python/provisioner/commands/check.py
```diff
@@ -137,6 +137,37 @@ def is_set(self) -> bool:
...
+
+    @property
+    def check_name(self):
+        """
+
+        :return:
+        """
+        return self._check_name
+
+    def set_critical(self) -> None:
```
andkononykhin2, line 155: why do we need two separate methods here, can't we just add an argument and use only one?

api/python/provisioner/commands/check.py
```diff
@@ -220,13 +266,55 @@ def get_passed(self) -> list:
         """
         return [check for check in self._check_entries if check.is_passed]
 
-    def get_failed(self) -> list:
+    def get_failed(self, critical: bool = None) -> list:
```
andkononykhin2, line 269: we don't follow typing strictly but for the sake of accuracy and just if you are not aware: `Optional[bool]` I believe would be the best choice here

api/python/provisioner/commands/check.py
```diff
@@ -220,13 +266,55 @@ def get_passed(self) -> list:
...
+        return only critical checks that failed. If critical is set to False,
+        return only non-critical checks that failed. If critical is set to
+        None, return all failed checks.
+
+        :param bool critical: parameter to filter returning check results
 
+        :return: list with failed checks satisfying the given condition
         """
-        return [check for check in self._check_entries if check.is_failed]
+        if critical is None:
```
andkononykhin2, line 280: I think we can keep the previous version but instead of `self._check_entries` use result of `get_checks`

andkononykhin2 (commented):
@dmitrydb I see you are making some good design changes. I posted few comments. I might haven't got completely your idea but the one general concern:
- I see you are adding `critical` attribute for the check so it would have two different  characteristic 
  - check failed/passed
  - check critical/non critical
- but in in some places (e.g. in SWUpdate) you filtering by `critical` attr only so I guess passed critical might be selected and I guess it's not that needed

api/python/provisioner/commands/check.py
dmitrydb: Yes, I also think it would be better here and for `get_checks` method too.

api/python/provisioner/commands/check.py
dmitrydb: Yes, I agree. Fixed that

api/python/provisioner/commands/check.py
dmitrydb: Yes, I thought about that. The reason why I decided to keep `None` as one of the possible values is to mark that this property wasn't set by the user. This information is not used yet, but maybe it is good to store it and use it for unit tests in the future.
Also, as this property is optional for use, I treat it the same as a non-critical value if it would have been set to `False` value.

api/python/provisioner/commands/check.py
dmitrydb: I agree, it will take a less lines of code. Done

dmitrydb:
Hi @andkononykhin2, Thank you for the comments
>but in in some places (e.g. in SWUpdate) you filtering by critical attr only so I guess passed critical might be selected and I guess it's not that needed

Yes, I improved that. We need to select there all failed checks (critical and non-critical).

api/python/provisioner/commands/check.py
andkononykhin2: ok, makes sense about `is_set` api then

api/python/provisioner/commands/check.py
```diff
@@ -249,38 +326,47 @@ def to_dict(self) -> dict:
...
-    @abstractmethod
     def make_decision(self, check_result: CheckResult):
         """
         Make a decision based on given check_result
 
         :return:
         """
+        if check_result.is_failed:
+            # Get non-critical errors first
+            warnings = check_result.get_checks(critical=False)
```
andkononykhin2, line 352: @dmitrydb  The same concern here as for SWUpdate before: do we need all non critical checks here or we need all failed ones?

dmitrydb: Here we need to fetch all non-critical failed checks. I fixed that. 
I left only the `get_checks` method with different parameters to specify the returning set of checks by characteristic like critical, failed.

ypise (approved):

PR #775, opened 2020-11-30
changing multipath as a new check and adding Dynamic imports

divya-vijay:

### DIFF
--- a/api/python/provisioner/commands/auto_deploy.py
+++ b/api/python/provisioner/commands/auto_deploy.py
@@ -17,6 +17,7 @@
 
 from typing import Type
 import logging
+from importlib import import_module
 
 from ..config import (
     PRVSNR_PILLAR_CONFIG_INI,
@@ -34,11 +35,6 @@ from .setup_provisioner import (
     SetupProvisioner,
     SetupCmdBase
 )
-from .check import (
-    Check,
-    PreChecksDecisionMaker,
-    PostChecksDecisionMaker
-)
 from . import deploy_dual
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


api/python/provisioner/commands/auto_deploy.py
```diff
@@ -55,19 +51,24 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
 
     @staticmethod
     def deployment_validations(deploy_check):
-        checker = Check()
+        check_import = import_module(
```
andkononykhin2, line 54: I guess it won't work since all commands are exported at the very beginning  when api spec yaml is parsed: https://github.com/Seagate/cortx-prvsnr/blob/2dd21bdae0fcc26471cf56398df59ccf0b701ac1/api/python/provisioner/commands/__init__.py#L1563-L1569.
I guess you may try `importlib.reload` instead.

andkononykhin2 (changes requested):
@divya-vijay please check my concern
@dmitrydb  could you verify that the rest logic (specific to Check) is ok as well
Thank you

api/python/provisioner/commands/auto_deploy.py
divya-vijay: Thanks @andkononykhin . 
`import_module()` did not reload and accept the new imports, `reload()` worked. Comment addressed.

andkononykhin2 (approved):

api/python/provisioner/commands/auto_deploy.py
dmitrydb: Why can't we do `reload` inside of `provisioner.commands.checks`? 
Also, we use the global variable `cortx_py_utils_import_error` in `checks` module, so we can validate if imports from `cortx.utils` was successful after `reload` in `deployment_validations` method.

api/python/provisioner/commands/auto_deploy.py
```diff
@@ -55,19 +52,23 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
 
     @staticmethod
     def deployment_validations(deploy_check):
-        checker = Check()
+        check_import = importlib.reload(check)
+
+        check_cmd = getattr(check_import, 'Check')
```
dmitrydb, line 57: I think, we can just use here:
```
check = importlib.reload(check)

check_cmd = check.Check
```
Or even more change ` precheck_dm().make_decision(check_result=check_res)` to
```python
check.PreChecksDecisionMaker().make_decision(check_result=check_res)
```

dmitrydb (changes requested):
Hi @divya-vijay, left minor comments.

api/python/provisioner/commands/auto_deploy.py
dmitrydb: There is also this option
```python
check = importlib.reload(check)

PreChecksDecisionMaker = check.PreChecksDecisionMaker
```
and use it as ordinary
```python
PreChecksDecisionMaker().make_decision(check_result=check_res)
```

divya-vijay: Comment addressed. Thanks.

dmitrydb (approved):
Looks good for me

mazinamdar (approved):

PR #778, opened 2020-12-02
Updated BMC check and fix flake8 issues

mazinamdar:

### DIFF
--- a/api/python/provisioner/commands/auto_deploy.py
+++ b/api/python/provisioner/commands/auto_deploy.py
@@ -61,12 +61,15 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
             raise ValueError("Error During Deployment "
                              f"{deploy_check} Validations: {str(exc)}")
         else:
-            if "pre" in deploy_check:
+            if deploy_check == GroupChecks.DEPLOY_PRE_CHECKS.value:
                 check_import.PreChecksDecisionMaker().make_decision(
                     check_result=check_res)
-            else:
+            elif deploy_check == GroupChecks.DEPLOY_POST_CHECKS.value:
                 check_import.PostChecksDecisionMaker().make_decision(
                     check_result=check_res)
+            else:
+                raise ValueError(
+                    f'Group Check "{deploy_check}" is not supported')
 
     def run(self, nodes, **kwargs):
         setup_provisioner_args = {
--- a/api/python/provisioner/commands/auto_deploy.py
+++ b/api/python/provisioner/commands/auto_deploy.py
@@ -49,7 +49,6 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
         deploy_dual.run_args_type
     ]
 
-
     @staticmethod
     def deployment_validations(deploy_check):
         check_import = importlib.reload(check)
@@ -60,13 +59,14 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
             check_res = check_cmd().run(deploy_check)
         except Exception as exc:
             raise ValueError("Error During Deployment "
-                            f"{deploy_check} Validations: {str(exc)}")
+                             f"{deploy_check} Validations: {str(exc)}")
         else:
             if "pre" in deploy_check:
-                check_import.PreChecksDecisionMaker().make_decision(check_result=check_res)
-
+                check_import.PreChecksDecisionMaker().make_decision(
+                    check_result=check_res)
             else:
-                check_import.PostChecksDecisionMaker().make_decision(check_result=check_res)
+                check_import.PostChecksDecisionMaker().make_decision(
+                    check_result=check_res)
 
     def run(self, nodes, **kwargs):
         setup_provisioner_args = {
--- a/api/python/provisioner/commands/check.py
+++ b/api/python/provisioner/commands/check.py
@@ -452,26 +452,6 @@ class Check(CommandParserFillerMixin):
 
         return res
 
-    def _bmc_accessibility(self, *, args: str) -> CheckEntry:
-        """
-        Check BMC accessibility
-
-        :param args: bmc accessibility check specific parameters and arguments
-        :return:
-        """
-        return self._validate_bmc_check(self.ACCESSIBLE)
-
-    def _bmc_stonith(self, *, args: str) -> Union[CheckEntry,
-                                                  List[CheckEntry]]:
-        """
-        Check BMC Stonith Configuration
-
-        :param args: bmc stonith configuration check specific
-                     parameters and arguments
-        :return:
-        """
-        return self._validate_bmc_check(self.STONITH)
-
     @staticmethod
     def _connectivity(*, servers: Union[list, tuple, set] = PENDING_SERVERS,
                       args: str) -> List[CheckEntry]:
@@ -1052,6 +1032,26 @@ class Check(CommandParserFillerMixin):
             res.append(check_ret)
 
         return res
+
+    def _bmc_accessibility(self, *, args: str) -> CheckEntry:
+        """
+        Check BMC accessibility
+
+        :param args: bmc accessibility check specific parameters and arguments
+        :return:
+        """
+        return self._validate_bmc_check(self.ACCESSIBLE)
+
+    def _bmc_stonith(self, *, args: str) -> Union[CheckEntry,
+                                                  List[CheckEntry]]:
+        """
+        Check BMC Stonith Configuration
+
+        :param args: bmc stonith configuration check specific
+                     parameters and arguments
+        :return:
+        """
+        return self._validate_bmc_check(self.STONITH)
 
     def _validate_bmc_check(self, check: str):
 
--- a/api/python/provisioner/commands/check.py
+++ b/api/python/provisioner/commands/check.py
@@ -1079,8 +1079,10 @@ class Check(CommandParserFillerMixin):
             return check_entry
 
         res: List[CheckEntry] = list()
+        #  TODO use of salt python API with function_run
         _DECRYPT_PSWD_CMD = ('salt-call lyveutil.decrypt cluster {val}'
                              ' --output=newline_values_only')
+        minion_id = local_minion_id()
 
         for node in nodes:
             check_res: CheckEntry = CheckEntry(bmc_check)
@@ -1093,7 +1095,6 @@ class Check(CommandParserFillerMixin):
 
                 # Decrypt the secret
                 cmd = _DECRYPT_PSWD_CMD.format(val=secret)
-                minion_id = local_minion_id()
                 bmc_passwd = cmd_run(cmd, targets=minion_id)[minion_id]
 
                 # Check the stonith configuration
--- a/api/python/provisioner/commands/check.py
+++ b/api/python/provisioner/commands/check.py
@@ -385,6 +385,9 @@ class Check(CommandParserFillerMixin):
 
     _PRV_METHOD_MOD = "_"  # private method modificator
 
+    STONITH = "stonith"
+    ACCESSIBLE = "accessible"
+
     @staticmethod
     def _network(*, args: str) -> List[CheckEntry]:
         """
@@ -449,18 +452,17 @@ class Check(CommandParserFillerMixin):
 
         return res
 
-    @staticmethod
-    def _bmc_accessibility(*, args: str) -> CheckEntry:
+    def _bmc_accessibility(self, *, args: str) -> CheckEntry:
         """
         Check BMC accessibility
 
         :param args: bmc accessibility check specific parameters and arguments
         :return:
         """
-        return Check._validate_bmc_check('accessible')
+        return self._validate_bmc_check(self.ACCESSIBLE)
 
-    @staticmethod
-    def _bmc_stonith(*, args: str) -> Union[CheckEntry, List[CheckEntry]]:
+    def _bmc_stonith(self, *, args: str) -> Union[CheckEntry,
+                                                  List[CheckEntry]]:
         """
         Check BMC Stonith Configuration
 
@@ -468,7 +470,7 @@ class Check(CommandParserFillerMixin):
                      parameters and arguments
         :return:
         """
-        return Check._validate_bmc_check('stonith')
+        return self._validate_bmc_check(self.STONITH)
 
     @staticmethod
     def _connectivity(*, servers: Union[list, tuple, set] = PENDING_SERVERS,
@@ -1051,13 +1053,14 @@ class Check(CommandParserFillerMixin):
 
         return res
 
-    @staticmethod
-    def _validate_bmc_check(check: str):
+    def _validate_bmc_check(self, check: str):
 
-        if check == "stonith":
+        if check == self.STONITH:
             bmc_check = cfg.Checks.BMC_STONITH.value
-        else:  # accessible
+        elif check == self.ACCESSIBLE:
             bmc_check = cfg.Checks.BMC_ACCESSIBILITY.value
+        else:
+            raise ValueError(f'BMC Check "{check}" is not supported')
 
         check_entry: CheckEntry = CheckEntry(bmc_check)
         # Check for import errors
--- a/api/python/provisioner/commands/check.py
+++ b/api/python/provisioner/commands/check.py
@@ -275,9 +275,10 @@ class DecisionMaker(ABC):
 
         if check_result.is_failed:
             failed_checks = "; ".join(str(check)
-                                    for check in check_result.get_failed())
+                                      for check in check_result.get_failed())
 
         return failed_checks
+
 
 class PreChecksDecisionMaker(DecisionMaker):
 
@@ -294,6 +295,7 @@ class PreChecksDecisionMaker(DecisionMaker):
                                   f"Checks have failed: {failed}")
 
         logger.info("All Pre-Flight Checks have Passed")
+
 
 class PostChecksDecisionMaker(DecisionMaker):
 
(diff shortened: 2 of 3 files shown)


sanjognaik (approved):
Looks good to me.

api/python/provisioner/commands/auto_deploy.py
```diff
@@ -60,13 +59,14 @@ def deployment_validations(deploy_check):
             check_res = check_cmd().run(deploy_check)
         except Exception as exc:
             raise ValueError("Error During Deployment "
-                            f"{deploy_check} Validations: {str(exc)}")
+                             f"{deploy_check} Validations: {str(exc)}")
         else:
             if "pre" in deploy_check:
```
dmitrydb, line 64: I think it is better to check with `deploy_check.startswith("pre-")`. It may be more accurate.

api/python/provisioner/commands/check.py
```diff
@@ -455,79 +457,18 @@ def _bmc_accessibility(*, args: str) -> CheckEntry:
...
-            # Check if BMC is accessible for these nodes
-            BmcV().validate('accessible', nodes)
-        except Exception as exc:
-            res.set_fail(checked_target=cfg.ALL_MINIONS,
-                         comment=str(exc))
-        else:
-            res.set_passed(checked_target=cfg.ALL_MINIONS)
-
-        return res
+        return Check._validate_bmc_check('accessible')
```
dmitrydb, line 460: It is better to make `_bmc_accesibility` as ordinary class method (remove `@staticmethod`). So it will allow to refer to `_validate_bmc_check` using `self.` and be independent of class renaming:
```python
return self._validate_bmc_check('accessible')
```
Also, please, use constants for `accessible` and `stonith` strings.

api/python/provisioner/commands/check.py
```diff
@@ -455,79 +457,18 @@ def _bmc_accessibility(*, args: str) -> CheckEntry:
...
-                                [node, bmc_ip, bmc_user, bmc_passwd])
-            except Exception as exc:
-                check_stonith.set_fail(checked_target=node, comment=str(exc))
-            else:
-                check_stonith.set_passed(checked_target=node)
-
-            res.append(check_stonith)
-
-        return res
+        return Check._validate_bmc_check('stonith')
```
dmitrydb, line 471: Same here. Make `_bmc_stonith` as bound class method by removing `@staticmethod` to refer to class method using `self.`:
```python
 return self._validate_bmc_check('stonith')
```

api/python/provisioner/commands/check.py
```diff
@@ -1104,6 +1051,59 @@ def _controller_ip(*, args: str) -> Union[CheckEntry, List[CheckEntry]]:
 
         return res
 
+    @staticmethod
+    def _validate_bmc_check(check: str):
+
+        if check == "stonith":
```
dmitrydb, line 1057: Use constants here for `stonith` and `accessible`. Also I would recommend you to use `if-elif-else` here, and raise an exception if the user passed something different from these two variants:
```python
if check == "stonith":
     bmc_check = cfg.Checks.BMC_STONITH.value
elif check == "accessible":
     bmc_check = cfg.Checks.BMC_ACCESSIBILITY.value
else:
    raise Exception("Unsupported check value")
```

dmitrydb (changes requested):
Thank you, @mazinamdar. It is a good improvement for validation API. Please, take a look at my minor comments.

api/python/provisioner/commands/auto_deploy.py
mazinamdar: DEPLOY_PRE_CHECKS = "deploy_pre_checks"    DEPLOY_POST_CHECKS = "deploy_post_checks"
We have pre and post in middle so we cannot use startswith here

api/python/provisioner/commands/check.py
mazinamdar: comment is addressed

api/python/provisioner/commands/check.py
mazinamdar: comment is addressed

api/python/provisioner/commands/check.py
mazinamdar: comment is addressed

api/python/provisioner/commands/auto_deploy.py
dmitrydb: Ok, then we can update
we can update
```python
if "pre" in deploy_check:
```
by the 
```python
if deploy_check in (GroupChecks.DEPLOY_PRE_CHECKS, DEPLOY_PRE_CHECKS.UNBOXING_PRE_CHECKS)
```

dmitrydb (changes requested):

api/python/provisioner/commands/auto_deploy.py

dmitrydb (approved):
All comments are addressed. Approve

api/python/provisioner/commands/auto_deploy.py
```diff
@@ -61,12 +61,15 @@ def deployment_validations(deploy_check):
...
+            if deploy_check == GroupChecks.DEPLOY_PRE_CHECKS.value:
                 check_import.PreChecksDecisionMaker().make_decision(
                     check_result=check_res)
-            else:
+            elif deploy_check == GroupChecks.DEPLOY_POST_CHECKS.value:
                 check_import.PostChecksDecisionMaker().make_decision(
                     check_result=check_res)
+            else:
+                raise ValueError(
+                    f'Group Check "{deploy_check}" is not supported')
```
divya-vijay, line 72: Overall, looks good to me. 
Although, we're calling the pre-check or post-check before and after Deploy step explicitly with the same GroupCheck value. 
`self.deployment_validations(GroupChecks.DEPLOY_PRE_CHECKS.value)`

So, validating them again here and raising an `else` block for exception may be unnecessary. Just my thought @mazinamdar @dmitrydb

api/python/provisioner/commands/check.py
```diff
@@ -1104,6 +1053,60 @@ def _controller_ip(*, args: str) -> Union[CheckEntry, List[CheckEntry]]:
...
+
+            try:
+                # Get BMC IP, user and secret.
+                bmc_ip = Check._get_pillar_data(f"cluster/{node}/bmc/ip")
+                bmc_user = Check._get_pillar_data(f"cluster/{node}/bmc/user")
+                secret = Check._get_pillar_data(f"cluster/{node}/bmc/secret")
+
+                # Decrypt the secret
+                cmd = _DECRYPT_PSWD_CMD.format(val=secret)
+                minion_id = local_minion_id()
```
andkononykhin2, line 1096: makes sense to move outside the loop since it should be always the same

andkononykhin2 (commented):
@mazinamdar Looks good. Please just check few minor comments. Thanks

api/python/provisioner/commands/check.py
```diff
@@ -385,6 +385,9 @@ class Check(CommandParserFillerMixin):
 
     _PRV_METHOD_MOD = "_"  # private method modificator
 
+    STONITH = "stonith"
+    ACCESSIBLE = "accessible"
+
```
divya-vijay, line 390: We can even move this to `config.py` and address as `cfg.STONITH` , so we can handle all validation-related config values in a single place.

api/python/provisioner/commands/auto_deploy.py
dmitrydb: Yes, at the first glance, it seems like unnecessary. But I always can do this:
```python
from .auto_deploy import AutoDeploy

AutoDeploy.deployment_validations("Hello")
```
And without this changes it will execute `PostChecksDecisionMaker` validation.
So as result, Mazhar's changes helps to keep method according to designed flow.

api/python/provisioner/commands/check.py
dmitrydb: I think it is unnecessary:

1. These values are only needed by the entire logic of the `Check` class and I don't see how we can reuse them outside of `check.py`
2. Also, we have already two different methods that provide bmc checks without these parameters: `_bmc_stonith` and `_bmc_accessibility`.

So these parameters help us to reuse `_validate_bmc_check` code without duplication. But if we really need them outside of check.py, I agree, we need to move it to `config.py`

api/python/provisioner/commands/check.py
```diff
@@ -1104,6 +1053,60 @@ def _controller_ip(*, args: str) -> Union[CheckEntry, List[CheckEntry]]:
 
         return res
 
+    def _validate_bmc_check(self, check: str):
```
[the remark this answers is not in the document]
mazinamdar: As per @dmitrydb suggestion I have not used staticmethod.
 
It is better to make _bmc_accesibility as ordinary class method (remove @staticmethod). So it will allow to refer to _validate_bmc_check using self. and be independent of class renaming:

`return self._validate_bmc_check('accessible')`

api/python/provisioner/commands/check.py
```diff
@@ -1104,6 +1053,60 @@ def _controller_ip(*, args: str) -> Union[CheckEntry, List[CheckEntry]]:
...
+        try:
+            # Get node list
+            nodes = Check._get_pillar_data("cluster/node_list")
+        except Exception as exc:
+            check_entry.set_fail(checked_target=cfg.ALL_MINIONS,
+                                 comment=str(exc))
+            return check_entry
+
+        res: List[CheckEntry] = list()
+        _DECRYPT_PSWD_CMD = ('salt-call lyveutil.decrypt cluster {val}'
```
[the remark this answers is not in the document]
mazinamdar: comment is addressed

api/python/provisioner/commands/check.py
mazinamdar: comment is addressed

api/python/provisioner/commands/check.py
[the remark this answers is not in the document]
dmitrydb: I think @andkononykhin2 means something different. You can do this: keep this two method as non-static:
```
    def _bmc_accessibility(self, *, args: str) -> CheckEntry:
        """
        Check BMC accessibility
        :param args: bmc accessibility check specific parameters and arguments
        :return:
        """
        return self._validate_bmc_check(self.ACCESSIBLE)

    def _bmc_stonith(self, *, args: str) -> Union[CheckEntry,
                                                  List[CheckEntry]]:
        """
        Check BMC Stonith Configuration
        :param args: bmc stonith configuration check specific
                     parameters and arguments
        :return:
        """
        return self._validate_bmc_check(self.STONITH)
```
But you can make static this method
```
   @staticmethod
    def _validate_bmc_check(check: str):

        if check == self.STONITH:
            bmc_check = cfg.Checks.BMC_STONITH.value
        elif check == self.ACCESSIBLE:
            bmc_check = cfg.Checks.BMC_ACCESSIBILITY.value
        else:
            raise ValueError(f'BMC Check "{check}" is not supported')
```
There is no contradiction here.

@andkononykhin2 correct me if I am wrong

api/python/provisioner/commands/check.py
mazinamdar: I agree with @dmitrydb , These are just constants we used in Check class actual check names are already in config so I think no need to move them to config.

`BMC_ACCESSIBILITY = "bmc_accessibility"
 BMC_STONITH = "bmc_stonith"
`

divya-vijay: I agree with the scope not being used. This was more about handling all validation related-parameters in `config.py`.

api/python/provisioner/commands/auto_deploy.py
divya-vijay: Sure, but any sample name in the check_name will be handled in Exception in `check.run()` .
But, okay, let us have the `else` block. Thanks

```
        try:
            check_res = check_cmd().run(deploy_check)
        except Exception as exc:
            raise ValueError("Error During Deployment "
                            f"{deploy_check} Validations: {str(exc)}")
```

dmitrydb: @divya-vijay I think we don't understand each other. Let me just clarify what I did mean.
My previous example is not the best choice to demonstrate my point. 
So before this `else` block appeared, I could call
```python
from .auto_deploy import AutoDeploy

AutoDeploy.deployment_validations("swupdate_checks")
```
As it is the correct and supported value from `GroupChecks`, there will be no Exceptions here:
```
        try:
            check_res = check_cmd().run(deploy_check)
        except Exception as exc:
            raise ValueError("Error During Deployment "
                            f"{deploy_check} Validations: {str(exc)}")
```

But after that we had the following block (before new and latest Mazhar changes)
```python
            if "pre" in deploy_check:
                check_import.PreChecksDecisionMaker().make_decision(check_result=check_res)
            else:
                check_import.PostChecksDecisionMaker().make_decision(check_result=check_res)
```
in case of `swupdate_checks` parameter `else` block will be executed.
So question to you @divya-vijay. Do you think that this is the correct behavior? If you think that was the plan, I don't mind reverting changes.

dmitrydb: Based on this conversation, I would recommend to implement it as:
```python

_DECISION_MAKERS = {
    "deploy_pre_checks": "PreChecksDecisionMaker",
    "deploy_post_checks": "PostChecksDecisionMaker",
}

@staticmethod

    if deploy_check not in self._DECISON_MAKERS:
        raise ValueError(f'Group Check "{deploy_check}" is not supported')

    check_import = importlib.reload(check)

    check_cmd = check_import.Check

    try:
        check_res = check_cmd().run(deploy_check)
    except Exception as exc:
        raise ValueError("Error During Deployment "
                         f"{deploy_check} Validations: {str(exc)}")


    getattr(check_import, deploy_check)().make_decision(check_result=check_res)
```

divya-vijay: @dmitrydb No, I understood what you meant. 
What I meant was, we have a similar implementation in `replace_node.py` .
`def repl_node_validations(repl_check):`
We initially decided to make this function common so all the modules can use from a single place, but later decided to have specific it as per each implementation. In a common implementation, validating the incoming param is required. 
But now, the `check` parameter that is passed to this method would already be specific to it and validated. 
Which is why I thought re-validating them with the same `value == GroupChecks.DEPLOY_PRE_CHECKS.value` might not be necessary. 
As long as the check name is valid, we only need to verify if they belong to either pre- or post- category. 

In case of your example, "swupdate_checks" does not have "pre" in it, so, post-checks block will be executed and raise warning (which is how we're handling checks for SW now). 

But I understood your implementation suggestion here. So, we can have the same for now as per Mazhar's PR, and based on more modules validations in future, we can add any more improvements to it.

api/python/provisioner/commands/check.py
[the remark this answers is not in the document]
andkononykhin2: If you look around that function you will see only methods marked as static. So I wonder why it was placed here, looks non-consistent. So options:

-  if you want to keep it non-static, rearrange methods of that class
- or you can make it static (I didn't get why you had chosen something different here and are keeping static methods for all other checks)

For the latter you need to refer your checks using class name (like `Check.STONITH`).
One more option here is classmethod, so you could refer to it as `cls.STONITH` but it would be also non-consistent.

>So it will allow to refer to _validate_bmc_check using self. and be independent of class renaming

`self` can be used for staticmethods instead of classname, so no issues here

```
Python 3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class AAA:
...   @staticmethod
...   def test_static():
...     print('STATIC')
...   def test(self):
...     self.test_static()
... 
>>> AAA().test()
STATIC
```

andkononykhin2 (approved):

api/python/provisioner/commands/auto_deploy.py
dmitrydb: >What I meant was, we have a similar implementation in `replace_node.py`.
`def repl_node_validations(repl_check):`

I have the same concern regarding to `repl_node_validations(repl_check)`. Why do we need to pass `repl_check` parameter to this function? I think this method can be presented as:
```python
    @staticmethod
    def repl_node_validations():
        checker = Check()
        try:
            check_res = checker.run(GroupChecks.REPLACENODE_CHECKS.value)
        except Exception as exc:
            raise errors.ValidationError("Error During Replace Nodes "
                                                       f"replacenode_checks Validations: {str(exc)}")
        else:
            PreChecksDecisionMaker().make_decision(check_result=check_res)
```
In this implementation, you don't need to care about parameter validations and anything else because this method does just one thing: runs the desired group of checks for replace_node verification and executes the `PreChecks` decision-maker.

>We initially decided to make this function common so all the modules can use from a single place, but later decided to have it specific as per each implementation. In a common implementation, validating the incoming param is required.
But now, the check parameter that is passed to this method would already be specific to it and validated.
Which is why I thought re-validating them with the same value == GroupChecks.DEPLOY_PRE_CHECKS.value might not be necessary.

I can't agree with this assumption that for common implementation we need to validate incoming parameters, but for specific implementations, we don't. If you have no parameters validation for a specific implementation, so it means that you have just a common implementation. Because it can be executed with all parameters that `check_res = check_cmd().run(deploy_check)` accepts including single checks (**"network", "hostnames", "mgmt_vip"**, etc.). And for all those cases we will run the `PostChecksDecisionMaker` decision-maker class.  So my main concern here if you mark class method to be specific, so it should have specific behavior and should not allow executing cases which I mentioned even if we are thinking that this method will be invoked always correctly and with correct input parameters.

>In case of your example, "swupdate_checks" does not have "pre" in it, so, post-checks block will be executed and raise warning (which is how we're handling checks for SW now).

Yes, exactly. For `swupdate_checks` we will run `PostChecksDecisionMaker` class which works the same as `SWUpdateDecisionMaker`.
But again, consider this case
```python
from .auto_deploy import AutoDeploy

AutoDeploy.deployment_validations("replacenode_checks")
```
In `def repl_node_validations(repl_check):` we execute in this case `PreChecksDecisionMaker().make_decision(check_result=check_res)`. 
But for `def deployment_validations(deploy_check):` we execute `check_import.PostChecksDecisionMaker().make_decision(check_result=check_res)`. 

`PreChecksDecisionMaker` raises exceptions for fail results, `PostChecksDecisionMaker` shows warning message to user. So for me it looks like a conflict here. So this is why I requested Mazhar to add `else` block here.

To conclude our long and productive discussion, we have three options here:
* We can revert changes for `AutoDeploy.deployment_validations`
* We can keep the current variant
* We can change to use this variant: https://github.com/Seagate/cortx-prvsnr/pull/778#discussion_r534212931

So, @mazinamdar, it is up to you which option to choose. 

I think this variant: https://github.com/Seagate/cortx-prvsnr/pull/778#discussion_r534212931
is more readable and simple for the user's understanding.

divya-vijay: Thank you, Dmitry! Yes, I agree with the conclusion options here. Based on time as a factor, we can choose either of second or third option.

PR #781, opened 2020-12-02
@83bhp EOS-15360: Unboxing: cleanup pcs resources upon cluster health check validation failure

83bhp:


ypise:
I would prefer it for 1.0.1

ypise (approved):
Looks good to merge.

cli/src/common_utils/utility_scripts.sh
```diff
@@ -142,7 +146,12 @@ function ensure_healthy_cluster {
 
                 if [[ ${attempt} -ge 20 ]]; then
                     _lerror "Giving up after ${attempt} attempts."
-                    exit 20
+                    if [[ "$_exit_on_err" == true ]]; then
```
mssawant, line 149: So we never invoke this function with `_exit_on_err == true`? because in that case `ret_ensure_healthy_cluster` will remain 0.

cli/src/factory_ops/unboxing/init
```diff
@@ -304,7 +304,16 @@ echo "Bringing HCTL cluster out of maintenance mode" 2>&1 | tee -a ${LOG_FILE}
 hctl node unmaintenance --all 2>&1 | tee -a ${LOG_FILE}
 
 echo "Ensure cluster is in healthy state" 2>&1 | tee -a ${LOG_FILE}
-ensure_healthy_cluster
+
+ensure_healthy_cluster "false" "false"
```
ypise, line 308: Can we convert all exits in `ensure_healthy_cluster` to return to avoid execution halt with exit and reduce the lines of code?

cli/src/factory_ops/unboxing/init
```diff
@@ -304,7 +304,16 @@ echo "Bringing HCTL cluster out of maintenance mode" 2>&1 | tee -a ${LOG_FILE}
...
+if [[ $ret_ensure_healthy_cluster -ne 0 ]]; then
+    echo "Checking if cleaning the cluster resources helps" | tee -a ${LOG_FILE}
+    echo  -n "Cleaning failed resource history......" | tee -a ${LOG_FILE}
```
mssawant, line 311: Please log `pcs status` output to the logs before `pcs resource cleanup`.

cli/src/common_utils/utility_scripts.sh
```diff
@@ -142,7 +146,12 @@ function ensure_healthy_cluster {
...
+                        exit 20
```
ypise, line 150: Here rather than exit can we reset counter and run `pcs resource cleanup --all`?

cli/src/common_utils/utility_scripts.sh
```diff
@@ -153,7 +162,12 @@ function ensure_healthy_cluster {
             if [[ ${attempt} -ge 10 ]]; then
                 _lerror "One or more nodes in cluster is/are Offline."
                 _lerror "Giving up after 10 attempts."
-                exit 10
+                if [[ "$_exit_on_err" == true ]]; then
```
ypise, line 165: If the cluster is not online after 10 attempts, shouldn't we be worried?

cli/src/factory_ops/unboxing/init
mssawant, line 308: I think this should run in a loop, we are just trying it once. I think we should run this loop until `$ret_ensure_healthy_cluster -ne 0 `. So instead of if loop it should be a while loop.

mssawant, line 308: Please add a comment about the parameters to ensure_healthy_cluster. Ideally a function should not exit the script but just return and the invoker or the callee must decide to exit or not.

mssawant: I think let's add a jira ticket number https://jts.seagate.com/browse/EOS-15360 in the comment as this is still a workaround.

cli/src/common_utils/utility_scripts.sh
mssawant: @ypise, in that case I think update will be affected, lets not do anything in this function, let the callee take the decision to exit or to continue.

83bhp:
Closing this PR as the changes in validation code may have impact on other functionality like sw update.
I have opened other PR with minimal impact: https://github.com/Seagate/cortx-prvsnr/pull/783
Please review that (#783).

cli/src/factory_ops/unboxing/init
```diff
@@ -304,7 +304,16 @@ echo "Bringing HCTL cluster out of maintenance mode" 2>&1 | tee -a ${LOG_FILE}
...
+    pcs resource cleanup >> ${LOG_FILE}
```
ujjwalpl, line 312: pcs resource cleanup 2>&1 >> ${LOG_FILE}

PR #789, opened 2020-12-10
EOS-15569 display logs in better format in file

mazinamdar:

Direct appending dictionary to string not displaying logs in proper format so created custom function to convert dict to string.


api/python/provisioner/errors.py
```diff
@@ -16,9 +16,37 @@
...
+
+    def dict_to_str(self):
+
+        if isinstance(self._obj, dict):
+            res = ''
+
+            for key, value in self._obj.items():
+                if isinstance(value, dict):
+                    value = json.dumps(value, indent=4)
+                res = res + f"{key}: {value},\n"
```
andkononykhin2, line 36: I guess there would be no indentation for wrapped parts of a record, like that:
```
record1 ....
<json data starts for the line beginning>
record2 ...
```

andkononykhin2 (commented):
@mazinamdar I think that change may make logs very noisy since all long salt's results will be forcibly wrapped. Also sometimes better to have only one-line record so one would be able to use text tools (like grep) to filter the log.

So, I believe the following  options should be considered as a better way:
- limit such changes to only important ones (like errors)
- limit to logging handlers (consider all or some of the following):
  - use only for console handler
  - use a separate handler (thus a separate logging destination) or/and make that post-processing optional
- use a separate tool (cli command) to make a post-processing

api/python/provisioner/errors.py
```diff
@@ -16,9 +16,37 @@
 #
 
 import subprocess
+import json
 from typing import Dict, Union, Any
 
 
+class CustomFormat():
```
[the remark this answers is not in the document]
mazinamdar: comment is addressed

api/python/provisioner/errors.py
```diff
@@ -16,9 +16,37 @@
...
+
+    def __init__(self, obj):
+        self._obj = obj
+
+    def dict_to_str(self):
+
+        if isinstance(self._obj, dict):
+            res = ''
+
+            for key, value in self._obj.items():
```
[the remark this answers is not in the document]
mazinamdar: `>>> x = {'hi' : 'My name \n Mazhar'}
>>> print(f"hello {x}")
hello {'hi': 'My name \n Mazhar'}

>>> a = ''
>>> for i, j in x.items():
...  a = a + f"{i}:{j}\n"
...
>>> print(a)
hi:My name
 Mazhar
`

Direct appending dict to string is not displaying values in proper format if we manually read each value then only it is giving proper display format.

api/python/provisioner/salt.py
```diff
@@ -588,7 +588,8 @@ def run(
             try:
                 logger.debug(
                     "Function '{}' on '{}' resulted in {}"
-                    .format(fun, targets, res.results)
+                    .format(fun, targets,
+                            CustomFormat(res.results).dict_to_str())
```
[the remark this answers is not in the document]
mazinamdar: removed , for  now covered only for salt errors

api/python/provisioner/errors.py
mazinamdar: I have given indentation for key and value but in case of value, it will be displayed as it is if they do not have tab after newline then result will be displayed it as it is.

api/python/provisioner/errors.py
```diff
@@ -80,9 +108,14 @@ def __init__(
         self.reason = reason
 
     def __str__(self):
+        logger.error(
```
andkononykhin2, line 111: @mazinamdar that is too far, please don't do that in such a way: `__str__` should just return string.

andkononykhin2: As an option:
- leave that `__str__` as is
- call logging for the error explicitly in the place where it is raised (we don't have a single place to deal with that for now) with a wrapped output so two log lines would appear: wrapped one first, legacy one-lined - second.

It's a very dummy/dirty option as well but better than current one.

Better option would be:
- keep that one-line but log only valuable data (e.g. do not log the states that are not failed)

andkononykhin2 (changes requested):

api/python/provisioner/errors.py
mazinamdar: For now added error log where we are raising exceptions. In case of failure I think it is better to display complete information to user so that it can help to debug the issue.

andkononykhin2 (approved):

PR #802, opened 2020-12-23
EOS-13069: sspl files path change due to disable pyinstaller

satish-seagate:


ypise (approved):

satish-seagate:
Need approval from @cdeshmukh after that only we can go ahead for merge. SSPL code was not merged yet which needs parallel goes with this change. (https://github.com/Seagate/cortx-monitor/pull/259)

FYI - @ypise @83bhp

PR #816, opened 2020-12-28
EOS-16027: Restart all services once CSM and SSPL config's are updated

SomwanshiAnjali:

### DIFF
--- a/cli/src/factory_ops/unboxing/init
+++ b/cli/src/factory_ops/unboxing/init
@@ -335,7 +335,7 @@ provisioner pillar_set commons/repo/lustre \"\" --logfile --logfile-filename ${L
 echo "Removing the repos with Seagate URL from Server A" | tee -a ${LOG_FILE}
 seagate_refs_cleanup
 
-echo "Restarting the cluster resources.........." | tee -a ${LOG_FILE}
+echo "Restarting the cluster resources through switching maintenance mode........" | tee -a ${LOG_FILE}
 echo "INFO: Bringing HCTL cluster in maintenance mode" 2>&1 | tee -a ${LOG_FILE}
 hctl node maintenance --all --timeout-sec=600 2>&1 | tee -a ${LOG_FILE}
 echo "INFO: Bringing HCTL cluster out of maintenance mode" 2>&1 | tee -a ${LOG_FILE}
--- a/cli/src/factory_ops/unboxing/init
+++ b/cli/src/factory_ops/unboxing/init
@@ -336,33 +336,9 @@ echo "Removing the repos with Seagate URL from Server A" | tee -a ${LOG_FILE}
 seagate_refs_cleanup
 
 echo "Restarting the cluster resources.........." | tee -a ${LOG_FILE}
-echo -n "ClusterIP.............................." | tee -a ${LOG_FILE}
-pcs resource refresh ClusterIP:0 >> ${LOG_FILE}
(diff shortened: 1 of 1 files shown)


cli/src/factory_ops/unboxing/init
```diff
@@ -336,33 +336,12 @@ echo "Removing the repos with Seagate URL from Server A" | tee -a ${LOG_FILE}
 seagate_refs_cleanup
 
 echo "Restarting the cluster resources.........." | tee -a ${LOG_FILE}
```
nitin-seagate, line 338: Change this sentence preferably mentioning about cluster restart (through switching maintenance mode)

cli/src/factory_ops/unboxing/init
```diff
@@ -336,33 +336,12 @@ echo "Removing the repos with Seagate URL from Server A" | tee -a ${LOG_FILE}
...
-echo -n "ClusterIP.............................." | tee -a ${LOG_FILE}
```
nitin-seagate, line 339: Instead of deleting, guess we should check feasibility of moving this block to well-defined function lets say **restart_cluster_resources** (Just for backward-compatibility support even though we will not be calling this function).

This comment need not be addressed immediately in-case we need urgent fix (but should be planned for LR2 later).

nitin-seagate (approved):
Please ensure multiple successful test-runs before merging.

cli/src/factory_ops/unboxing/init
```diff
@@ -335,34 +335,13 @@ provisioner pillar_set commons/repo/lustre \"\" --logfile --logfile-filename ${L
...
-echo -n "UDS...................................." | tee -a ${LOG_FILE}
-pcs resource refresh uds >> ${LOG_FILE}
-sleep 5
-echo "Ok." | tee -a ${LOG_FILE}
-
-echo -n "HAProxy................................" | tee -a ${LOG_FILE}
-pcs resource refresh haproxy-c1 >> ${LOG_FILE}
-pcs resource refresh haproxy-c2 >> ${LOG_FILE}
-echo "Ok." | tee -a ${LOG_FILE}
+echo "Restarting the cluster resources through switching maintenance mode........" | tee -a ${LOG_FILE}
```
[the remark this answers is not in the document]
83bhp: also remove "......." as there is no -e to echo.

cli/src/factory_ops/unboxing/init
```diff
@@ -335,34 +335,13 @@ provisioner pillar_set commons/repo/lustre \"\" --logfile --logfile-filename ${L
...
+echo "INFO: Bringing HCTL cluster in maintenance mode" 2>&1 | tee -a ${LOG_FILE}
+hctl node maintenance --all --timeout-sec=600 2>&1 | tee -a ${LOG_FILE}
+echo "INFO: Bringing HCTL cluster out of maintenance mode" 2>&1 | tee -a ${LOG_FILE}
+hctl node unmaintenance --all 2>&1 | tee -a ${LOG_FILE}
 sleep 5
+echo "OK." | tee -a ${LOG_FILE}
```
83bhp, line 344: As there's no -e to echo (on line# 338) this "OK" would look out of place in the output, so please remove this line.

83bhp (changes requested):

83bhp (approved):
Looks good to me!

PR #823, opened 2021-01-07
EOS-15458: Provisioner overwrites the hosts file during deployment of system component

ypise:

### DIFF
--- a/cli/src/factory_ops/boxing/prov_tasks
+++ b/cli/src/factory_ops/boxing/prov_tasks
@@ -504,10 +504,16 @@ function cleanup_hosts {
     remote_mgmt_ip_srvnode_1=$(ssh_over_pvt_data ${pvt_ip_a} "ip -4 address show dev ${interface_mgmt_a} | grep inet | head -1 | awk '{print \$2}' | awk -F '/' '{print \$1}'")
     remote_mgmt_ip_srvnode_2=$(ssh_over_pvt_data ${pvt_ip_b} "ip -4 address show dev ${interface_mgmt_b} | grep inet | head -1 | awk '{print \$2}' | awk -F '/' '{print \$1}'")
 
-    _linfo "Removing management IPs from /etc/hosts"
+    echo "Removing management IPs from /etc/hosts"
     sed -i "s/${remote_mgmt_ip_srvnode_1}.*//g" /etc/hosts
     sed -i "s/${remote_mgmt_ip_srvnode_2}.*//g" /etc/hosts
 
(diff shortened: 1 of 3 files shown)


cli/src/factory_ops/boxing/prov_tasks
```diff
@@ -497,3 +497,17 @@ function remove_gfs_mounts {
...
+
+    remote_mgmt_ip_srvnode_1=$(ssh_over_pvt_data ${pvt_ip_a} "ip -4 address show dev ${interface_mgmt_a} | grep inet | head -1 | awk '{print \$2}' | awk -F '/' '{print \$1}'")
+    remote_mgmt_ip_srvnode_2=$(ssh_over_pvt_data ${pvt_ip_b} "ip -4 address show dev ${interface_mgmt_b} | grep inet | head -1 | awk '{print \$2}' | awk -F '/' '{print \$1}'")
+
+    _linfo "Removing management IPs from /etc/hosts"
+    sed -i "s/${remote_mgmt_ip_srvnode_1}.*//g" /etc/hosts
+    sed -i "s/${remote_mgmt_ip_srvnode_2}.*//g" /etc/hosts
+
+    _linfo "Copying updated /etc/hosts to Server node B"
+    scp -o "StrictHostKeyChecking=no" -i /root/.ssh/id_rsa_prvsnr /etc/hosts "${pvt_ip_b}":/etc/hosts
```
ujjwalpl, line 512: scp ... || true or error ? What is scp fails for some reason ?

83bhp (approved):

cli/src/factory_ops/boxing/prov_tasks
83bhp: The script has "set -ue" at the start, so the scp failure would result in the boxing failure as without /etc/hosts the boxing could not proceed.

ypise: If scp fails, it means there is something wrong with network or the SSH configuration.
Will add a error message to display the failure better.

ujjwalpl: If something is wrong with network, wont it be caught early enough ?

ypise: There ideally should not be anything wrong with network. Also, the `/etc/hosts` file should be present.
Thus, ideally scp should not change. But, if it does, we want to capture the error and let user know that one of these 2 causes resulted in command failure.

ujjwalpl (approved):

PR #824, opened 2021-01-07
EOS-15324: Use ssh to trigger shutdown

ypise:


cli/src/factory_ops/boxing/shutdown
```diff
@@ -63,13 +66,17 @@ function shutdown_node {
...
     elif [[ "$_node" == "node2" ]]; then
         echo "INFO: Shutting down srvnode-2 in 1 min." | tee -a ${LOG_FILE}
-        salt srvnode-2 system.shutdown 1 --async
+        # salt srvnode-2 system.shutdown 1 --async
+        ssh_over_pvt_data "${pvt_ip_b}" "shutdown -P +1 'Shutting down srvnode-2 in 1 min'" || true
     else
         echo "INFO: Shutting down both nodes in 1 min." | tee -a ${LOG_FILE}
-        salt "*" system.shutdown 1 --async --no-color --out-file=${LOG_FILE} --out-file-append
+        # salt "*" system.shutdown 1 --async --no-color --out-file=${LOG_FILE} --out-file-append
+        ssh_over_pvt_data "${pvt_ip_b}" "shutdown -P +1 'Shutting down srvnode-2 in 1 min'" || true
```
ujjwalpl, line 78: Why are we trying to shutdown node 2 in else case ? This is change of behavior.

83bhp (approved):

cli/src/factory_ops/boxing/shutdown
83bhp: in else both the nodes are shutdown, since the script is running from primary node the secondary node needs to be shutdown before shutting down the node where the script is invoked from. If you see the original code, the primary is shutdown after the secondary on the very next line (#79)

ujjwalpl (approved):

ujjwalpl (approved):

PR #826, opened 2021-01-08
[EOS-16313] Reload site package before validation API call

mazinamdar:

We install cortx-py-util package during deployment so that we are not able to import that package on the fly after installing after reloading site package cortx-py-util package is getting imported


api/python/provisioner/commands/auto_deploy.py
```diff
@@ -51,6 +52,11 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
 
     @staticmethod
     def deployment_validations(deploy_check):
+
+        # even if cortx-py-util package is installed during runtime
+        # it failed to import, reload of site pkg helped here
+        importlib.reload(site)
```
ypise, line 58: This happens due to chicken and egg problem.
We are attempting to load cortx-py-utils before installing cortx-utils rpm.
I believe we need to include any cortx-py-utils imports in try-except as well to avoid such situations.

ypise (approved):

PR #828, opened 2021-01-11
EOS-16308: R2-Config.ini validation

divya-vijay:
1. Code changes for config.ini validation for latest Cluster config design
Sample ini file tested on 2-node VM: https://github.com/Seagate/cortx-prvsnr/wiki/Cluster-Config#reference-configini-file
2. Param details and logic can be fine-tuned with design updates.
3. To be tested on 3-node, 6-node and HW setups.


api/python/provisioner/commands/validate_setup.py
```diff
@@ -134,12 +134,27 @@ def __attrs_post_init__(self):
         if len(missing_params) > 0:
             raise ValueError(f"Mandatory param missing {missing_params}")
 
+@attr.s(auto_attribs=True)
+class StorageNodeParamsValidation:
+    hostname: str = StorageNodeParams.hostname
```
divya-vijay, line 139: This is only for testing and a placeholder for future node-specific storage enclosure params.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -95,118 +99,30 @@ def __attrs_post_init__(self):
...
 @attr.s(auto_attribs=True)
 class ConfigureSetup(CommandParserFillerMixin):
     input_type: Type[inputs.NoParams] = inputs.NoParams
     _run_args_type = RunArgsConfigureSetup
 
     validate_map = {"cluster": NetworkParamsValidation,
                     "node": NodeParamsValidation,
-                    "storage_enclosure": StorageEnclosureParamsValidation}
+                    "controller": StorageNodeParamsValidation,
+                    "srv_default": ServerDefaultParamsValidation,
```
andkononykhin2, line 110: @divya-vijay what is a difference between `server` and `node`?

api/python/provisioner/commands/configure_setup.py
```diff
@@ -95,118 +99,30 @@ def __attrs_post_init__(self):
...
+                    "storage_enclosure_default": StorageEnclosureDefaultParamsValidation}
```
andkononykhin2, line 111: @divya-vijay what is a difference between `controller` and `storage enclosure`?

api/python/provisioner/commands/configure_setup.py
```diff
@@ -255,12 +172,32 @@ def run(self, path, number_of_nodes):  # noqa: C901
...
         for section in content:
             input_type = section
             pillar_type = section
-            if 'srvnode' in section:
-                input_type = 'node'
-                pillar_type = f'cluster/{section}'
+            if 'default' in section:
+                #for section_name in ['srv_default', 'storage_enclosure_default']:
+                if 'srvnode' in section:
+                    input_type = 'srv_default'
```
andkononykhin2, line 184: @divya-vijay  could you use some variable for `srv_default` and `storage_enclosure_default` (better -- constant like in upper case) here and in lines 175-177 (and in any other referenced places) to avoid hard coding? The same for `default` as an expected prefix, `cluster`, `storage_enclosure` and any other strings that make special meaning for our business logic. Thank you

api/python/provisioner/inputs.py
```diff
@@ -466,16 +466,84 @@ class Release(ParamGroupInputBase):
     target_build: str = ReleaseParams.target_build
 
 
-class StorageEnclosureParams():
-    _param_group = 'storage_enclosure'
+class ServerDefaultParams():
```
andkononykhin2, line 469: @divya-vijay I feel some misunderstanding here: such kind of class helps to avoid copy paste for attributes definitions. 

So the idea is:
- you have some classes `...Params` that just keep attrs definitions and groups them (e.g. network, controller ...)
- you have many validation classes that may include (patially or fully) attrs defined in the classes above

It means:
- we don't need to define attrs multiple time
- in latter classes we may combine attrs from different groups

So here makes sense:
- decide which word to use for node: `server` or `node` (let's say `node`)
- remove `Default` suffix and use `NodeParams` class to define attrs that describes some node
- existent class `NodeNetworkParams` defines set of node params for network, so `NodeParams` may inherit from it and define only non-network params

api/python/provisioner/inputs.py
```diff
@@ -492,23 +560,40 @@ class StorageEnclosureParams():
 
 
 @attr.s(auto_attribs=True)
-class StorageEnclosure(ParamGroupInputBase):
-    controller_a_ip: str = StorageEnclosureParams.primary_mc_ip
-    controller_b_ip: str = StorageEnclosureParams.secondary_mc_ip
-    controller_user: str = StorageEnclosureParams.controller_user
-    controller_secret: str = StorageEnclosureParams.controller_secret
+class StorageEnclosureDefault(ParamGroupInputBase):
```
andkononykhin2, line 563: @divya-vijay the same idea here: please adjust according to the comment above for node attributes

andkononykhin2 (changes requested):

api/python/provisioner/inputs.py
```diff
@@ -466,16 +466,84 @@ class Release(ParamGroupInputBase):
...
+        _param_group, descr="list of dns servers as json"
+    )
+    bmc_user: str = ParamGroupInputBase._attr_ib(
+        _param_group, descr="node BMC User"
+    )
+    bmc_secret: str = ParamGroupInputBase._attr_ib(
+        _param_group, descr="node BMC password"
+    )
+    network_mgmt_interfaces: str = ParamGroupInputBase._attr_ib(
+        _param_group, descr=" Network Mgmt Interfaces"
```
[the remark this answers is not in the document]
divya-vijay: not really, some existing params had, and some didn't. Removed space to make it uniform throughout.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -222,7 +138,8 @@ def _parse_input(self, input):
                 value = [f'\"{x.strip()}\"' for x in input[key].split(",")]
                 value = ','.join(value)
                 input[key] = f'[{value}]'
-            elif 'mgmt_nw.iface' in key:
+            #elif 'network.mgmt.interfaces' in key:
+            elif 'interfaces' or 'device' or 'devices' in key:
```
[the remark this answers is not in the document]
divya-vijay: yep, missed while pushing it. Addressed now.

api/python/provisioner/commands/configure_setup.py
divya-vijay: Just to bring out a difference between `default` params and `node` params. Removed `Default` from class definition for all.

api/python/provisioner/commands/configure_setup.py
divya-vijay: Same as above. Just to show a difference. Sticking with 'node' for all Node-related sections and 'storage-enclosure' for all Storage-Enclosure sections.

api/python/provisioner/commands/configure_setup.py
divya-vijay: Addressed with CONSTANTS for all params in config file..

api/python/provisioner/inputs.py
divya-vijay: Thanks @andkononykhin for the suggestions. 
Addressed the same for now. And, based on updates/naming convention changes in config design, have planned to finalise then. 

1. For now, since `NodeNetworkParams` had all `srvnode-default` section params, so I have added few more related params to the same class and handled it. 
2. `NodeParams` will consist only of params specific to each node. 
3. `Default` is removed from both Node and StorageEnclosure class definitions.

api/python/provisioner/inputs.py
divya-vijay: `StorageEnclosureParams` addressed all params, so removed `Default` from name. 
Just created a class `StorageNodeParams` as placeholder for node-specific storage params in future.

mazinamdar:
@divya-vijay  is there any difference in configure_setup and validate_setup in provisioner dir and commands dir?

divya-vijay:
> @divya-vijay is there any difference in configure_setup and validate_setup in provisioner dir and commands dir?

@mazinamdar No, some merge conflict as there are many changes since. updated in the latest commit.

api/python/provisioner/commands/validate_setup.py
```diff
@@ -0,0 +1,201 @@
...
+            logger.error(
+                f"{missing_params} cannot be left empty or blank.. "
+                "These are mandatory to configure the setup."
+            )
+            raise ValueError(f"Mandatory param missing {missing_params}")
+
+
+@attr.s(auto_attribs=True)
+class StorageNodeParamsValidation:
+    hostname: str = StorageEnclosureParams.hostname
```
mazinamdar, line 107: I don't think hostname should come under storage section. can you please verify.

api/python/provisioner/inputs.py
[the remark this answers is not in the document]
andkononykhin2: thank you, yes, looks like a legacy

api/python/provisioner/commands/validate_setup.py
divya-vijay: @mazinamdar this is just a placeholder, like I had mentioned. 
could be any param, I preferred `hostname` just as a safe one. We can edit later once we have more params to mention in inidividual storage sections, till then I did not want to leave them empty. 
Sample config.ini I used for testing is here : 
https://github.com/Seagate/cortx-prvsnr/wiki/Cluster-Config

mazinamdar: can it be same as default storage section ? if I want to change controller ip's for only particular node ?

api/python/provisioner/commands/validate_setup.py
```diff
@@ -0,0 +1,201 @@
...
+            )
+            raise ValueError(f"Mandatory param missing {missing_params}")
+
+
+@attr.s(auto_attribs=True)
+class NodeParamsValidation:
+    hostname: str = NodeParams.hostname
+    is_primary: str = NodeParams.is_primary
+    roles: str = NodeParams.roles
+    bmc_ip: str = NodeParams.bmc_ip
```
mazinamdar, line 173: bmc_ip should be optional as for HW we get this directly no need provide them in config.ini

api/python/provisioner/inputs.py
```diff
@@ -523,43 +530,81 @@ class NodeNetworkParams():
...
+    network_mgmt_public_ip_addr: str = NodeNetworkParams.network_mgmt_public_ip_addr
+    pvt_ip_addr: str = NodeNetworkParams.pvt_ip_addr
+
+
+class NodeParams():
+    _param_group = 'node'
+    hostname: str = ParamGroupInputBase._attr_ib(
+        _param_group, descr="node hostname"
+    )
+    roles: str = ParamGroupInputBase._attr_ib(
```
mazinamdar, line 587: roles should be type List.

api/python/provisioner/inputs.py
```diff
@@ -523,43 +530,81 @@ class NodeNetworkParams():
...
+
+
+@attr.s(auto_attribs=True)
+class NodeNetwork(ParamGroupInputBase):
+    cluster_id: str = NodeNetworkParams.cluster_id
+    bmc_user: str = NodeNetworkParams.bmc_user
+    bmc_secret: str = NodeNetworkParams.bmc_secret
+    network_mgmt_gateway: str = NodeNetworkParams.network_mgmt_gateway
+    network_mgmt_netmask: str = NodeNetworkParams.network_mgmt_netmask
+    network_mgmt_interfaces: str = NodeNetworkParams.network_mgmt_interfaces
```
mazinamdar, line 573: should be type List

api/python/provisioner/commands/validate_setup.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python
```
andkononykhin2, line 1: that file doesn't include any command so not a good place to locate it inside `commands`

api/python/provisioner/inputs.py
```diff
@@ -467,54 +467,61 @@ class Release(ParamGroupInputBase):
...
 
 class NodeNetworkParams():
     _param_group = 'node'
-    hostname: str = ParamGroupInputBase._attr_ib(
-        _param_group, descr="node hostname"
-    )
-    roles: List = ParamGroupInputBase._attr_ib(
-        _param_group, descr="List of roles assigned to the node"
+    type: str = ParamGroupInputBase._attr_ib(
+        _param_group, descr="Common Server configuration"
```
andkononykhin2, line 518: shouldn't we use a word `type` in the description in some way, for now description doesn't provide any understanding of the parameter

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,145 +101,51 @@ def __attrs_post_init__(self):
...
 class ConfigureSetup(CommandParserFillerMixin):
     input_type: Type[inputs.NoParams] = inputs.NoParams
     _run_args_type = RunArgsConfigureSetup
 
-    validate_map = {
-        "cluster": NetworkParamsValidation,
-        "node": NodeParamsValidation,
-        "storage": StorageEnclosureParamsValidation
-    }
+    validate_map = {f'{CLUSTER}': NetworkParamsValidation,
```
andkononykhin2, line 109: sorry, but I don't understand the idea, here we have

- cluster - we validating network params
- node - validating node params (looks ok here)
- node default - validating node network params
- storage-enclosure - validating storage node params
- storage-enclosure-default - validating storage params  (is storage node the same as just storage enclosure?)

For me it would be much more logical if we do:
- cluster - validating cluster params
- node - node params
- node default - the same set of params
- storage - storage params
- stroage-default - the same set of params

Once everything is validated we construct the result configuration as:
- cluster level params
- node params as merge of default plus specific per node (the latter overrides the former)
- storage params ... similar to the above for nodes

@ypise could you clarify the initial idea?

andkononykhin2 (changes requested):
@divya-vijay Thank you for the updates: looks much better now! But I still feel lost for some design parts. Plus some misses in syntax/functionality. Please check my comments.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,145 +101,51 @@ def __attrs_post_init__(self):
...
+                    f'{NODE}': NodeParamsValidation,
+                    f'{STORAGE_ENCLOSURE}': StorageNodeParamsValidation,
+                    f'{NODE_DEFAULT}': NodeNetworkParamsValidation,
+                    f'{STORAGE_DEFAULT}': StorageEnclosureParamsValidation}
 
-    def _parse_params(self, input):
+    @staticmethod
+    def _parse_params(input):
```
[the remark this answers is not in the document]
divya-vijay: @andkononykhin thanks, yes, I have updated this after codacy conflict.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,145 +101,51 @@ def __attrs_post_init__(self):
...
                 params[val[-1]] = input[key]
         return params
 
     def _validate_params(self, input_type, content):
         params = self._parse_params(content)
         self.validate_map[input_type](**params)
 
-    def _parse_input(self, input):
+    @staticmethod
+    def _parse_input(input):
```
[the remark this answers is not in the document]
divya-vijay: Updated in latest commit.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,145 +101,51 @@ def __attrs_post_init__(self):
...
-                'private_interfaces', 'public_interfaces',
-                'gateway', 'netmask', 'public_ip', 'private_ip'
-            ]:
+            conf_values = [
+                'ip', 'user', 'secret', 'ipaddr', 'interfaces', 'gateway',
+                'netmask', 'public_ip_addr', 'type',
+                'id', 'roles', 'data_devices', 'metadata_device']
+
+            if len(val) > 2 and val[-1] in conf_values:
+                params[f'{val[-3]}_{val[-2]}_{val[-1]}'] = input[key]
```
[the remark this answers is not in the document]
mazinamdar: I think we don't need above code block.

` if len(val) > 2 and val[-1] in conf_values:
                params[f'{val[-3]}_{val[-2]}_{val[-1]}'] = input[key]`

logic is to  append only last 2 elements 
so that we can uniquely identify each element.
example:
 network.data.netmask=
network.data.gateway=
network.mgmt.netmask=

we need only below data which is params for validation class.
data_netmast
data.gateway
mgmt.netmask

@divya-vijay  can you please check if code works fine after removing above block.

cortx-admin:

api/python/provisioner/commands/configure_setup.py
ypise: Yes I agree with:
```
For me it would be much more logical if we do:
cluster - validating cluster params
node - node params
node default - the same set of params
storage - storage params
stroage-default - the same set of params
```

`storage_enclosure` is now being called `storage` as it could be an attached disk as well.

divya-vijay: @andkononykhin @ypise Addressed. 
Params are updated to fit into these 3 categories as per current config.ini format. Can be changed later if there are updates to this format. Thank you for the suggestions.

api/python/provisioner/commands/validate_setup.py
divya-vijay: added this as part of pretty format of logs in `utf` format, removed now.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -241,36 +158,60 @@ def _parse_input(self, input):
...
-
+                input_type = (f'{NODE_DEFAULT}' if 'default' in section
+                              else f'{NODE}')
+            elif 'storage' in section:
+                input_type = (f'{STORAGE_DEFAULT}' if 'default' in section
+                              else f'{STORAGE_ENCLOSURE}')
+
+            pillar_map_type = pillar_map[input_type]
+            pillar_type = f'{pillar_map_type}/{section}'
+            count = count - 1
```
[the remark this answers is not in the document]
divya-vijay: Parsing logic updated now. Thanks

api/python/provisioner/commands/configure_setup.py
```diff
@@ -241,36 +158,60 @@ def _parse_input(self, input):
...
-    def _parse_pillar_key(self, key):
+    @staticmethod
+    def _parse_pillar_key(key):
         pillar_key = deepcopy(key)
         return pillar_key.replace(".", "/")
 
     def run(self, path, number_of_nodes):  # noqa: C901
 
         if not Path(path).is_file():
+            logger.error(
```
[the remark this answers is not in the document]
divya-vijay: Done!

api/python/provisioner/commands/configure_setup.py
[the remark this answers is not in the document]
divya-vijay: In latest naming changes of `config.ini`, the length of few variables was more than 2. 
Updated parsing logic and removed this block now. Thanks !

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,145 +101,51 @@ def __attrs_post_init__(self):
...
             if input[key] and "," in input[key]:
                 value = [f'\"{x.strip()}\"' for x in input[key].split(",")]
                 value = ','.join(value)
                 input[key] = f'[{value}]'
-            elif 'mgmt.interfaces' in key:
-                # special case single value as array
-                # Need to fix this array having single value
-                input[key] = f'[\"{input[key]}\"]'
-            elif 'roles' in key:
+            elif ('interfaces' or 'device' or 'devices' or 'roles') in key:
```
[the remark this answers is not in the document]
divya-vijay: @andkononykhin created a list of keys to check, so more can be added if necessary, based on design changes. Thanks.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,145 +101,51 @@ def __attrs_post_init__(self):
...
-                'gateway', 'netmask', 'public_ip', 'private_ip'
-            ]:
+            conf_values = [
+                'ip', 'user', 'secret', 'ipaddr', 'interfaces', 'gateway',
+                'netmask', 'public_ip_addr', 'type',
+                'id', 'roles', 'data_devices', 'metadata_device']
+
+            if len(val) > 2 and val[-1] in conf_values:
+                params[f'{val[-3]}_{val[-2]}_{val[-1]}'] = input[key]
+                logger.debug(f"Params generated: {params}")
```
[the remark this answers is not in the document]
divya-vijay: Done.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -241,36 +158,60 @@ def _parse_input(self, input):
...
         count = int(number_of_nodes)
 
+        pillar_map = {f'{NODE_DEFAULT}': f'{CLUSTER}',
+                      f'{NODE}': f'{CLUSTER}',
+                      f'{STORAGE_ENCLOSURE}': f'{STORAGE_ENCLOSURE}',
+                      f'{STORAGE_DEFAULT}': f'{STORAGE_ENCLOSURE}'}
+
         for section in content:
             input_type = section
             pillar_type = section
```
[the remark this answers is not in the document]
divya-vijay: Parsing logic updated now. Thanks.

api/python/provisioner/inputs.py
```diff
@@ -476,63 +476,107 @@ class Release(ParamGroupInputBase):
...
 
-class NodeNetworkParams():
+class NodeParams():
     _param_group = 'node'
     hostname: str = ParamGroupInputBase._attr_ib(
         _param_group, descr="node hostname"
     )
     roles: List = ParamGroupInputBase._attr_ib(
-        _param_group, descr="List of roles assigned to the node"
+        _param_group, descr="role of the node"
```
andkononykhin, line 548: @divya-vijay  why not list of roles?

api/python/provisioner/inputs.py
andkononykhin: I meant that `Common Server configuration` looks too generic. I believe it should be something like `type of smth`

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,199 +98,166 @@ def __attrs_post_init__(self):
...
 class ConfigureSetup(CommandParserFillerMixin):
     input_type: Type[inputs.NoParams] = inputs.NoParams
     _run_args_type = RunArgsConfigureSetup
 
-    validate_map = {
-        "cluster": NetworkParamsValidation,
-        "node": NodeParamsValidation,
-        "storage": StorageEnclosureParamsValidation
-    }
+    validate_map = {f'{CLUSTER}': ClusterParamsValidation,
```
andkononykhin, line 106: could you add a simple comment about what are the keys (section name of the ini) and values (validation class).

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,199 +98,166 @@ def __attrs_post_init__(self):
...
+        for pillar_key in content[section]:
+
+            key = f'{pillar_type}/{self._parse_pillar_key(pillar_key)}'
+            value_to_set = f"{content[section][pillar_key]}"
+
+            run_subprocess_cmd([
+                   "provisioner", "pillar_set",
+                   key, value_to_set])
+
+        if content.get('cluster', None):
```
andkononykhin, line 182: that block of code would be run each time that function is called, looks not clean enough

api/python/provisioner/commands/configure_setup.py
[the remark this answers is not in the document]
andkononykhin: I meant that we can use validation as part of `path` attr definition in `RunArgsConfigureSetup`, so more accurate to say "Move to 'path' definition"

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,199 +98,166 @@ def __attrs_post_init__(self):
...
-        pillar_type = None
-        node_list = []
-        count = int(number_of_nodes)
-
-        for section in content:
-            input_type = section
-            pillar_type = section
+        logger.info("Updating salt data..")
+        content = {section: dict(config.items(section))
+                   for section in config.sections()}  # noqa: E501
```
andkononykhin, line 210: as long as you wrapped that long line `noqa` suppression comment is no more needed

andkononykhin (commented):
@divya-vijay Major concerns are resolved. Thank you! I added just few minor comments. Please take a look.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,199 +98,166 @@ def __attrs_post_init__(self):
...
 
-        if content.get('cluster', None):
-            if content.get('cluster').get('cluster_ip', None):
-                run_subprocess_cmd([
-                       "provisioner", "pillar_set",
-                       "s3clients/ip",
-                       f"{content.get('cluster').get('cluster_ip')}"])
+            # Contents of 'default' sections will be applied to each node
+            if input_type == f'{NODE_DEFAULT}':
+                for each_node in parsed_nodes['server_list']:
```
andkononykhin, line 247: I have one more concern here: I guess currently you can guarantee that update for defaults goes first for all nodes. We need that since specific non-default values should overwrite it.

api/python/provisioner/commands/configure_setup.py
```diff
@@ -90,199 +98,166 @@ def __attrs_post_init__(self):
...
+                    pillar_type = f'{pillar_map_type}/{each_node}'
+                    self._update_pillar_data(content, section, pillar_type)
 
-        if count > 0:
-            raise ValueError(f"Node information for {count} node missing")
+            elif input_type == f'{STORAGE_DEFAULT}':
```
andkononykhin, line 251: The same concern as for node: update with defaults should go first before any specific set of values. Please ensure that it is always true.

api/python/provisioner/commands/validate_setup.py
```diff
@@ -0,0 +1,181 @@
...
+    STORAGE_DEFAULT
+)
+
+from ..values import UNCHANGED
+
+logger = logging.getLogger(__name__)
+
+
+@attr.s(auto_attribs=True)
+class ValidateSetup:
```
andkononykhin, line 38: It's not a command

I believe we don't need that module inside `commands`.  And that class with the only static method looks unnecessary.
So I suggest:
- move the method (the logic of) `_parse_nodes` to `ConfigureSetup`
- move the module outside `commands` since there would be actually definitions of inputs for configuration file

andkononykhin (commented):
@divya-vijay I've noticed few more places. Could you check as well? Thanks

api/python/provisioner/commands/validate_setup.py
divya-vijay: This was originally to validate node sections and count - which I later split to PR 910. 
We'll move some common modules to outside of `commands/` to `utils/` which will be addressed in a separate ticket - https://jts.seagate.com/browse/EOS-18518 .

andkononykhin2: As I see that module is a new one and came from inputs.py which is outside the  `commands` directory. And EOS-18518 is about restructuring **inside** commands. Currently I think we don't have any common modules inside commands.

divya-vijay: Agreed! So, for now, I've moved it out of `commands/`. Will handle other alterations in future scope. Thanks

api/python/provisioner/commands/configure_setup.py
divya-vijay: For now, I've added a code block to handle 'defaults` separately. Once the config design is finalised and the team started using the updated `ini` file, we can remove and update this block accordingly. 
If it encounters the individual node again, it will simply continue and not re-update the same pillar data. 
I hope this might temporarily ensure the contents of `default` sections are distributed across nodes. 
Thanks.

api/python/provisioner/commands/configure_setup.py
divya-vijay: Yes, updated!

api/python/provisioner/commands/configure_setup.py
divya-vijay: Done. Moved it out of iteration.

api/python/provisioner/commands/configure_setup.py
divya-vijay: Added a comment for user to understand the param scope.

api/python/provisioner/inputs.py
divya-vijay: Typo. Updated now. Thanks

api/python/provisioner/commands/validate_setup.py
```diff
@@ -0,0 +1,181 @@
...
+
+    @staticmethod
+    def _parse_nodes(content):
+        """
+        Parses the content of nodes.
+
+        The node names are parsed based on config-sections.
+
+        """
+        section_list = []
```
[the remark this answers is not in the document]
divya-vijay: Okay. Sorted the keys in the beginning and removed from loop.

api/python/provisioner/commands/configure_setup.py
divya-vijay: Hope this block ensure that. Once the setup is up and running, will test on multi-node setups and update test scenarios accordingly.

api/python/provisioner/inputs.py
divya-vijay: Yes, removed this param from nodes altogether. It was not necessary here.

api/python/provisioner/commands/configure_setup.py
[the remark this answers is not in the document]
divya-vijay: Sure, will handle this as part of EOS-18518. Guess we have a list of small things like these to handle. 
Thanks.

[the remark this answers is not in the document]
andkononykhin2: We don't have such a list. But for that case we actually don't need that comment.

api/python/provisioner/validate.py
```diff
@@ -0,0 +1,180 @@
...
+        Parses the content of nodes.
+
+        The node names are parsed based on config-sections.
+
+        """
+        section_list = sorted(list(content))
+        output = {}
+
+        # Gives section names of only nodes : srvnode[1-12]
+        servers = [srv for srv in section_list if "srvnode" in srv
```
andkononykhin2, line 48: @divya-vijay That logic became more more clearer now. Thanks. But there are still two problems here:
1. you iterate over sections 3 times instead of just 1. Please use just single `for` loop with multiple conditionals inside to filter enclosures, servers and clusters
2. once collected per section type you don't use that enough actually and again [iterate over section list repeating the same conditionals](--build-orig-single-iso cortx-2.0.0-791-single.iso): `'srvnode' in`, `'storage' in`, `'default' not in` etc. Instead please:
    - consider to filter default for voth servers and storages right here and assign them to a separate key inside `output` (e.g. `output[NODE_DEFAULT] = servers_defaults` and `output[STORAGE_DEFAULT] = enclosures_defaults` )
    - use these constants for other keys: `NODE` - for server, `STORAGE` - for enclosures, `CLUSTER` - for clusters (btw do we consider multiple clusters actually @ypise)
    - update logic of validation accordingly:
      - you would be able to apply defaults first without any additional filtering there
      - you would be able to iterate over servers and storages separately with any additional filtering as well

andkononykhin2 (changes requested):

divya-vijay:
Implementation is present in pre-cortx and some validations that have changed will be handled in separate PRs.

divya-vijay:
This patch was addressed as part of PRs #1060 and #1071 and merged. Hence, this PR is no more valid.

PR #830, opened 2021-01-12
EOS-16495: VM: Single Node CORTX Stack Provisioning (cortx_setup)

ypise:


sradhanand-pati (approved):

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1024,6 +1024,36 @@ def _prepare_factory_setup_pillar(
...
+        
+        cluster_sls_sample_path = (
+            config.PRVSNR_PILLAR_DIR / 'samples/dualnode.cluster.sls'
+        )
+        if len(run_args.nodes) == 1:
+            cluster_sls_sample_path = (
+                config.PRVSNR_PILLAR_DIR / 'samples/singlenode.cluster.sls'
+            )
+        
+        pillar_dir = config.PRVSNR_DATA_ROOT_DIR / "shared/srv/pillar/groups/all"
```
andkononykhin2, line 1042: `PRVSNR_DATA_SHARED_DIR` would be good to improve that a bit

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1267,6 +1297,10 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
                     )
                 })
 
+            self._prepare_cluster_pillar(
```
andkononykhin2, line 1300: please do not (consider any target nodes changes here, that stage of the logic is only about preparation (e.g. prepare pillar of salt-ssh and not on target nodes). In current version even in case when setup (jump) server is one of the target nodes it will not affect another node. For remote case it would be worse - it would try to put the pillar into `/var/lib` of that not related system.

 I would suggest you to go another way:
- let core provisioner part be installed on nodes
- do something like `file.managed` with `makedirs=True` copying one of the samples (single or dual) into calculated pillar locations, as you do inside that fun:
```
        pillar_dir = config.PRVSNR_DATA_SHARED_DIR / "srv/pillar/groups/all"
        cluster_sls_path = add_pillar_merge_prefix(
            pillar_dir / 'cluster.sls'
        )
```

In other words you may consider to create a state file like [here](https://github.com/Seagate/cortx-prvsnr/blob/6da5c201a94676cc88d0008df142c78d49eab91d/api/python/provisioner/srv/salt/provisioner/configure_salt_master.sls#L22-L27) where:
- `name` might be passed via pillar (either online or may be put inside pillar file profile as for release in `_prepare_release_pillar` function).
- `source` might be put under jinja depending on nodes number

andkononykhin2 (changes requested):

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1790,12 +1792,29 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
...
+                        'cluster_pillar_path':
+                            PRVSNR_PILLAR_DIR / "/groups/all/cluster.sls",
+                        'nodes': len(run_args.nodes)
+                    }
+                }
+            }
+        }
+        ssh_client.cmd_run(
+            f"salt-call state.apply components.provisioner.config",
+            targets=ALL_MINIONS,
```
andkononykhin2, line 1815: targeting al minions for that pillar destination in case of glusterfs sharing is overhead and even may relate to an issue.

I would suggest to make that more granular:
- split states appliance into:
  - ones that should be applied for all minions
  - ones for all masters only
  - ones that updates pillar:
     - in case of clustered (ha) we need to run that only on one node
     - in case of non-ha - all nodes

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1790,12 +1792,29 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
...
             ssh_client.cmd_run(
                 f"salt-call state.apply {state}",
                 targets=master_targets
             )
+        
+        # Seperation of variable to make flake8 happy
+        provisioner_pillar = {
+            'pillar': {
+                'provisioner': {
+                    'cluster': {
```
andkononykhin2, line 1805: generally I think we would need to re-think usage of inline pillar: bad that it sticks our logic much to api so states become more and more not self-contained.
So, that change looks pretty good for now. But will likely be re-thought later.

srv/components/provisioner/salt_master/config.sls
```diff
@@ -20,6 +20,22 @@ salt_master_config_updated:
...
+Update salt-master.service:
+  file.managed:
+    - name: /usr/lib/systemd/system/salt-master.service
+    - source: salt://components/provisioner/salt_master/files/salt-master.service
+
+Reload updated services:
+  cmd.run:
+    - name: systemctl daemon-reload
+    - onchanges:
+      - file: Update salt-master.service
```
andkononykhin2, line 38: I believe the same relation is required for the last state to restart salt-master in the background since in general it may impact the way the service is started.

srv/components/provisioner/files/cluster_sls.template
```diff
@@ -20,65 +20,50 @@ cluster:
   mgmt_vip:                           # Management VIP for CSM
   search_domains:                     # Do not update
   dns_servers:                        # Do not update
-  type: dual                          # single/dual/generic/3_node
```
andkononykhin2, line 23: why do we need old two files and that combined one?

srv/components/provisioner/salt_master/config.sls
```diff
@@ -20,6 +20,22 @@ salt_master_config_updated:
     - name: /etc/salt/master
     - source: salt://components/provisioner/salt_master/files/master
 
+# Always start glusterfshsaredstorage before salt-master
+Update glusterfssharedstorage.service:
```
andkononykhin2, line 24: it might be not the best place for that (in salt's config). Also in case of any changes systemd daemon should be also reloaded.

andkononykhin2 (changes requested):

ypise:
Tested on a VM to be working as expected.
Merging for further integrations with #837

PR #833, opened 2021-01-15
EOS-15325 (sms20): Updating glusterfs service relation with salt-master

ypise:
Signed-off-by: sradhanandpati sradhanand.pati@seagate.com

### DIFF
--- a/cli/src/replace_node/deploy-replacement
+++ b/cli/src/replace_node/deploy-replacement
@@ -162,6 +162,7 @@ function system_states {
         # "system.storage.multipath.install"
         # "system.storage.multipath.config"
         "system.network"
+        "system.network.mgmt.service_port"
         "system.network.data.public"
         "system.network.data.direct"
         "misc_pkgs.rsyslog"
--- a/srv/components/provisioner/config.sls
+++ b/srv/components/provisioner/config.sls
@@ -29,15 +29,15 @@ provisioner_rsyslog_conf_updated:
       - service: Start rsyslog
 
 # Always start glusterfshsaredstorage before salt-master
-Update salt-master.service:
-  file.managed:
-    - name: /usr/lib/systemd/system/salt-master.service
-    - source: salt://components/provisioner/salt_master/files/salt-master.service
-
 Update glusterfssharedstorage.service:
   file.managed:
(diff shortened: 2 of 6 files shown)


cli/src/replace_node/deploy-replacement
```diff
@@ -162,6 +162,7 @@ function system_states {
         # "system.storage.multipath.install"
         # "system.storage.multipath.config"
         "system.network"
+        "system.network.mgmt.service_port"
```
andkononykhin2, line 165: @ypise  looks like not related thing, do we have a confidence here?

andkononykhin2 (approved):

cli/src/replace_node/deploy-replacement
ypise: Yes, it a required feature.
We are not performing boxing-unboxing after a node replacement. Hence, we need this.

It's also present: https://github.com/Seagate/cortx-prvsnr/blob/f1151526312d2689b7f625b8b4b507ebebd13eac/cli/src/factory_ops/boxing/prov_tasks#L115
```
function configure_service_port {
    salt "srvnode-1" state.apply components.system.network.mgmt.service_port ${salt_opts} || (
        echo "ERROR: Failed to configure service port on srvnode-1." | tee -a ${LOG_FILE}
        exit 25
    )
}
```

srv/components/provisioner/config.sls
```diff
@@ -28,8 +28,13 @@ provisioner_rsyslog_conf_updated:
     - watch_in:
       - service: Start rsyslog
 
-# Always start glusterfshsaredstorage before salt-minion
-Update glusterfshsaredstorage.service:
+# Always start glusterfshsaredstorage before salt-master
```
nitin-seagate, line 31: Sequence-wise should we write GlusterFS info before this Salt-Master section

nitin-seagate (approved):
LGTM

ujjwalpl (approved):

srv/components/provisioner/config.sls
ypise: @nitin-seagate here were are just copying service files from template location to target location.
No service is being restarted.

However, I've sequenced the states appropriately.

83bhp (approved):

PR #839, opened 2021-01-19
EOS-16704: FW update over sftp- handle error cases.

83bhp:
Handled following error cases:
1. controller goes down in middle of the fw update.
2. sftp command exits with non zero return code.


nitin-seagate (approved):
Looks good to me...

ypise (approved):

PR #840, opened 2021-01-20
EOS-16661: Create script for provisioner code coverage automation

SomwanshiAnjali:


cli/src/code_coverage/prvsnr-coverage.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/python3
+
+import argparse
+import subprocess
+import sys
+import os.path
+import shlex
```
ypise, line 7: It's a good practice to arrange the imports alphabetically.

cli/src/code_coverage/prvsnr-coverage.py
```diff
@@ -0,0 +1,56 @@
...
+        --cov test/ --cov-report=xml"
+    )
+    if os.path.isfile(f'coverage.xml'):
+        print("Coverage report generated successfully!!")
+    else:
+        print("Coverage report generation failed!!")
+        sys.exit(1)
+    print("Done")
+
+def parse_args():
```
ypise, line 44: Is this a private method?
It would be better to have it as:
`_parse_args()`

cli/src/code_coverage/prvsnr-coverage.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/python3
+
+import argparse
+import subprocess
+import sys
+import os.path
+import shlex
+
+def _run_command(cmd):
```
ypise, line 9: Prefer calling it to `_run` or `_execute`.

cli/src/code_coverage/prvsnr-coverage.py
```diff
@@ -0,0 +1,56 @@
...
+    try:
+        res = subprocess.run(shlex.split(cmd), shell=False)
```
ypise, line 11: Also, would we need `sudo` ever?
I recently discovered subprocess is potentially risky.
Add a check for allowed commands only to be executed, so that no one can attempt:
`rm -rf`

Similar has been attempted in:
https://github.com/Seagate/cortx-prvsnr/blob/150ca3047a25cebbccc79e8cfdbb62f72e3873c3/srv/_modules/setup_conf.py#L72

cli/src/code_coverage/prvsnr-coverage.py
```diff
@@ -0,0 +1,56 @@
...
+        if res.returncode:
+            print(f"failed to run command {cmd} ")
+            sys.exit(1)
```
ypise, line 14: `subprocess.run` can handle exit non-zero with and exception.
Use parameter `check=True`
https://docs.python.org/3/library/subprocess.html#subprocess.run

ypise (changes requested):

cli/src/code_coverage/prvsnr-coverage.py
SomwanshiAnjali: Addressed

cli/src/code_coverage/prvsnr-coverage.py
SomwanshiAnjali: Addressed

cli/src/code_coverage/prvsnr-coverage.py
SomwanshiAnjali: Addressed

cli/src/code_coverage/prvsnr-coverage.py
SomwanshiAnjali: Addressed

cli/src/code_coverage/prvsnr-coverage.py
SomwanshiAnjali: Addressed

ypise (approved):

PR #843, opened 2021-01-21
EOS-16560: Pillar data translation to JSON

divya-vijay:
A straightforward API which converts pillar data to JSON and dumps it into a file.


api/python/provisioner/commands/__init__.py
```diff
@@ -431,6 +438,34 @@ def run(self, *args, targets=ALL_MINIONS, dry_run=False, **kwargs):
...
+    ):
+        return cls(input_type=getattr(inputs, input_type))
+
+
+    def run(
+        self, *args, targets: str = ALL_MINIONS, local: bool = False,
+        **kwargs
+    ):
+        res = PillarGet.run(self)
+        os.makedirs(f"{PRVSNR_CORTX_CONFIG_DIR}", exist_ok=True)
```
andkononykhin2, line 458: please use [pathlib.mkdir](https://docs.python.org/3.6/library/pathlib.html#pathlib.Path.mkdir) here instead, please note that PRVSNR_CORTX_CONFIG_DIR is already a Path object

api/python/provisioner/commands/__init__.py
```diff
@@ -431,6 +438,34 @@ def run(self, *args, targets=ALL_MINIONS, dry_run=False, **kwargs):
...
+
+        resp_loaded = json.loads(json.dumps(res).replace('null', '""'))
+        for key in resp_loaded.keys():
+            resp_loaded[key] = ({k:v
+                 for k, v in resp_loaded[key].items() if not k == "mine_functions"})
```
andkononykhin2, line 463: `not k == "mine_functions"` will work but `k != ...` would look more friendly

api/python/provisioner/commands/__init__.py
```diff
@@ -431,6 +438,34 @@ def run(self, *args, targets=ALL_MINIONS, dry_run=False, **kwargs):
...
+    def run(
+        self, *args, targets: str = ALL_MINIONS, local: bool = False,
+        **kwargs
+    ):
+        res = PillarGet.run(self)
+        os.makedirs(f"{PRVSNR_CORTX_CONFIG_DIR}", exist_ok=True)
+
+        resp_loaded = json.loads(json.dumps(res).replace('null', '""'))
+        for key in resp_loaded.keys():
+            resp_loaded[key] = ({k:v
```
andkononykhin2, line 462: you may drop need here additional parenthesis here

api/python/provisioner/config.py
```diff
@@ -38,6 +38,9 @@
 #  - then rename base.py to config.py
 #  - remove PRVSNR_ prefix
 
+PRVSNR_CORTX_CONFIG_DIR = Path('/opt/seagate/cortx_configs')
+PRVSNR_CORTX_CONFIG_FILE = PRVSNR_CORTX_CONFIG_DIR / 'provisioner_cluster.json'
```
andkononykhin2, line 42: I believe here we need to either use something like `PRVSNR_CLUSTER_CONFIG_FILE` or `cortx_cluster.json`. In current it seemed like we introduce some provisioner config to be treated as config for the whole CORTX. Also makes sense to think about some suffixes for both directory and files (e.g. `mini`)

@nitin-seagate @ypise do you have any comments here?

api/python/provisioner/commands/__init__.py
```diff
@@ -431,6 +438,34 @@ def run(self, *args, targets=ALL_MINIONS, dry_run=False, **kwargs):
...
+                 for k, v in resp_loaded[key].items() if not k == "mine_functions"})
+        resp_dumped = json.dumps(resp_loaded, indent=4, sort_keys=True)
+        resp_unload = open(f'{PRVSNR_CORTX_CONFIG_FILE}', "w").write(resp_dumped)
```
andkononykhin2, line 465: that should be closed as well, please use pathlib (PRVSNR_CORTX_CONFIG_FILE is already a Path object) instead, e.g.:
```
PRVSNR_CORTX_CONFIG_FILE.write_text(resp_dumped)
```

api/python/provisioner/commands/__init__.py
```diff
@@ -112,6 +114,11 @@ class RunArgsPillarGet(RunArgsBase):
     local: bool = RunArgs.local
 
 
+@attr.s(auto_attribs=True)
+class RunArgsPillarDump(RunArgsBase):
+    local: bool = RunArgs.local
```
andkononykhin2, line 119: As stated in EOS-16560 we need also an optional argument for path to be generated, default value is `/opt/seagate/cortx_configs/provisioner_cluster.json`

andkononykhin2 (changes requested):

api/python/provisioner/commands/__init__.py
ypise: unwanted_keys = ("mine_functions", "provisioner", "glusterfs")
```
for key in resp_loaded.keys():
    if key in unwanted_keys:
        del resp_loaded[key]
```

api/python/provisioner/config.py
ypise: PRVSNR_CORTX_CONFIG_DIR --> CORTX_CONFIG_DIR
PRVSNR_CORTX_CONFIG_FILE --> CONFSTORE_CLUSTER_CONFIG

api/python/provisioner/commands/__init__.py
```diff
@@ -431,6 +438,34 @@ def run(self, *args, targets=ALL_MINIONS, dry_run=False, **kwargs):
...
+        cls, input_type: str = 'PillarKeysList'
+    ):
+        return cls(input_type=getattr(inputs, input_type))
+
+
+    def run(
+        self, *args, targets: str = ALL_MINIONS, local: bool = False,
+        **kwargs
+    ):
+        res = PillarGet.run(self)
```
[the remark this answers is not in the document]
ypise: res -> full_pillar_load

api/python/provisioner/commands/__init__.py
```diff
@@ -431,6 +438,34 @@ def run(self, *args, targets=ALL_MINIONS, dry_run=False, **kwargs):
...
+        os.makedirs(f"{PRVSNR_CORTX_CONFIG_DIR}", exist_ok=True)
+
+        resp_loaded = json.loads(json.dumps(res).replace('null', '""'))
```
ypise, line 460: res is dict()
don't make any operations here. This line is not required.

api/python/provisioner/commands/__init__.py
```diff
@@ -431,6 +438,34 @@ def run(self, *args, targets=ALL_MINIONS, dry_run=False, **kwargs):
...
+        for key in resp_loaded.keys():
```
ypise, line 461:
```
for key in res.keys():
```

ypise (changes requested):

api/python/provisioner/commands/__init__.py
divya-vijay: Addressed. Thanks

api/python/provisioner/config.py
divya-vijay: @ypise Addressed now.

api/python/provisioner/commands/__init__.py
divya-vijay: @ypise Addressed to make it more user-friendly. Thanks.

api/python/provisioner/commands/__init__.py
divya-vijay: Yes, Addressed now.

api/python/provisioner/commands/__init__.py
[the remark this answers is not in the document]
divya-vijay: @ypise @andkononykhin Addressed. Thanks

api/python/provisioner/commands/__init__.py
```diff
@@ -431,6 +438,34 @@ def run(self, *args, targets=ALL_MINIONS, dry_run=False, **kwargs):
 
         return self._run(params, targets)
 
+@attr.s(auto_attribs=True)
```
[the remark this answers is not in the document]
divya-vijay: Created a separate module in `commands/` and added.

api/python/provisioner/commands/__init__.py
divya-vijay: @ypise @andkononykhin @nitin-seagate 

This is how the converted data would looks like (a sample block) 

`
                "cluster_id": "a918a173-c8e4-4231-ac11-d3472d265291",
                "hostname": "ssc-vm-c-1210.colo.seagate.com",
                "is_primary": **"True"**,
                "network": {
                    "data": {
                        "public_ip_addr": "172.20.30.101"
                    },
                    "data_nw": {
                        "gateway": **"None"**,
                        "iface": [
                            "eth1",
                            "eth2"
                        ],
                        "netmask": "255.255.0.0",
                        "public_ip_addr": "None",
                        "pvt_ip_addr": "",
                        "roaming_ip": "127.0.0.1"
                    }
`

api/python/provisioner/api_spec.yaml
```diff
@@ -23,6 +23,9 @@ grains_get:
 pillar_get:
   type: PillarGet
   input_type: PillarKeysList
+pillar_translate:
```
ypise, line 26: Let's call this pillar_export.

api/python/provisioner/commands/pillar_translate.py
```diff
@@ -0,0 +1,84 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
ypise, line 2: Rename this file to pillar_export.py

api/python/provisioner/commands/pillar_translate.py
```diff
@@ -0,0 +1,84 @@
...
+        return str(obj)
+    elif isinstance(obj, int):
+        return str(obj)
+    elif isinstance(obj, Mapping):
+        return {k: convert_to_str(v, repl) for k, v in obj.items()}
+    return obj
+
+
+@attr.s(auto_attribs=True)
+class PillarTranslate(PillarGet):
```
ypise, line 63: Rename this class to PillarExport.

api/python/provisioner/commands/pillar_translate.py
```diff
@@ -0,0 +1,84 @@
...
+    _run_args_type = RunArgsPillarTranslate
```
ypise, line 64: Rename to RunArgsPillarExport

api/python/provisioner/api_spec.yaml
```diff
@@ -23,6 +23,9 @@ grains_get:
...
+  type: PillarTranslate
```
ypise, line 27: PillarExport

api/python/provisioner/commands/pillar_translate.py
```diff
@@ -0,0 +1,84 @@
...
+
+        unwanted_keys = ["mine_functions", "provisioner", "glusterfs"]
+
+        for key in full_pillar_load.keys():
+            for filter_key in list(full_pillar_load[key].keys()):
+                if filter_key in unwanted_keys:
+                    del full_pillar_load[key][filter_key]
+
+        convert_data = convert_to_str(full_pillar_load)
+        resp_dumped = json.dumps(convert_data, indent=4, sort_keys=True)
```
ypise, line 81: Why do we use dumps here?
Code should be simple:
```
convert_data = self._convert_to_str(full_pillar_load)
with open(CONFSTORE_CLUSTER_CONFIG, 'w') as fd:
    json.dump(convert_data, fd)
```
Indentation actually increases the size of the output file here, which is unnecessary.

api/python/provisioner/commands/pillar_translate.py
```diff
@@ -0,0 +1,84 @@
...
+        CONFSTORE_CLUSTER_CONFIG.write_text(resp_dumped)
+
+        return resp_dumped
```
ypise, line 84: Do we need to return the JSON? If yes, why?
Return, if any, should be a boolean value.

api/python/provisioner/commands/pillar_translate.py
```diff
@@ -0,0 +1,84 @@
...
+class RunArgsPillarTranslate(RunArgsBase):
+    local: bool = RunArgs.local
+
+
+# Method to address ConfStore LIMITATION on accepting ONLY string values
+# Can be commented when the bug is addressed.
+
+# Converts all values(null, bool, int) to string
+
+def convert_to_str(obj, repl=''):
```
ypise, line 50: Should be a class method and not a function. We are not calling it as default method from json.dump, so not required to be a function.

api/python/provisioner/commands/pillar_translate.py
```diff
@@ -0,0 +1,84 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@attr.s(auto_attribs=True)
+class RunArgsPillarTranslate(RunArgsBase):
+    local: bool = RunArgs.local
+
+
+# Method to address ConfStore LIMITATION on accepting ONLY string values
```
ypise, line 45: Convert this comment into a doc_string.

api/python/provisioner/commands/pillar_translate.py
```diff
@@ -0,0 +1,84 @@
...
+# Can be commented when the bug is addressed.
+
+# Converts all values(null, bool, int) to string
+
+def convert_to_str(obj, repl=''):
+    if isinstance(obj, type(None)):
+        return str(obj)
```
ypise, line 52: If we are converting all types to string, why not do it in a single if statement?
```
def convert_to_str(obj, repl=''):
    if isinstance(obj, type(None)) or
        isinstance(obj, bool) or
        isinstance(obj, int):
            return str(obj)
        elif isinstance(obj, Mapping):
            return {k: convert_to_str(v, repl) for k, v in obj.items()}
        return obj
```

ypise (changes requested):
@sradhanand-pati 
In one of our previous config-provisioner code we had used JSON.dump default method to translate null value.
Could you update this code to implement similar solution using JSONEncoder?

api/python/provisioner/commands/pillar_translate.py
divya-vijay: Yes, just for a placeholder in case more types are added to convert to string, kept that way. Addressed as a single line now. Thanks.

api/python/provisioner/commands/pillar_translate.py
divya-vijay: Comment addressed.

api/python/provisioner/commands/pillar_translate.py
divya-vijay: Comment addressed. Thanks

api/python/provisioner/commands/pillar_translate.py
divya-vijay: Tried to keep it similar to `get` , incase user wants to check the value. 
Comment addressed now, returning only a message to the user.

api/python/provisioner/commands/pillar_translate.py
divya-vijay: Just for `pretty print` of JSON or readable format to user, used `dumps()` added the indentation. Since we're not returning only message to the user now, addressed it. Thanks

api/python/provisioner/api_spec.yaml
divya-vijay: Comment addressed.

api/python/provisioner/commands/pillar_translate.py
divya-vijay: Comment addressed.

api/python/provisioner/commands/pillar_translate.py
divya-vijay: Comment addressed.

api/python/provisioner/commands/pillar_translate.py
divya-vijay: Comment addressed.

api/python/provisioner/api_spec.yaml
divya-vijay: Comment addressed.

api/python/provisioner/commands/pillar_export.py
```diff
@@ -0,0 +1,93 @@
...
+            full_pillar_load = PillarGet.run(self, *args, **kwargs)
+            Path(CORTX_CONFIG_DIR).mkdir(exist_ok=True)
+
+            unwanted_keys = ["mine_functions", "provisioner", "glusterfs"]
+            for key in full_pillar_load.keys():
+                for filter_key in list(full_pillar_load[key].keys()):
+                    if filter_key in unwanted_keys:
+                        del full_pillar_load[key][filter_key]
+
+            convert_obj = PillarJSONEncoder()._convert_to_str
```
ypise, line 82: `_convert_to_str` is supposed to be a private method for a class in Python.
The comment was assuming the method would be part of same class. 
Here PillarJSONEncoder is an overhead and unnecessary.
Also, I don't understand why we are capturing the function definition into a variable, how does it help?

ypise (changes requested):

ypise:
@divya-vijay 
Do resolve the codacy reported issues.

api/python/provisioner/commands/pillar_export.py
divya-vijay: Thought having it separately might be easier to enhance for more data types or comment it out later. Addressed now. Thanks

divya-vijay:
> @divya-vijay
> Do resolve the codacy reported issues.

@ya

> @divya-vijay
> Do resolve the codacy reported issues.

@ypise The codacy issue with docstring prevails either way, with or without a new line.

ypise (approved):

api/python/provisioner/commands/pillar_export.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        if (isinstance(obj,
+           (type(None), bool, list, int))):
+            return str(obj)
+        elif isinstance(obj, Mapping):
+            return {k: self._convert_to_str(v, repl) for k, v in obj.items()}
+        return obj
+
+    def run(
+        self, *args, targets: str = ALL_MINIONS, local: bool = False,
```
andkononykhin2, line 67: `targets` and `local` are optional by a parent class but not passed there, and not used. It may lead to some usage limitations in future.

api/python/provisioner/commands/pillar_export.py
```diff
@@ -0,0 +1,90 @@
...
+
+            unwanted_keys = ["mine_functions", "provisioner", "glusterfs"]
+            for key in full_pillar_load.keys():
+                for filter_key in list(full_pillar_load[key].keys()):
+                    if filter_key in unwanted_keys:
+                        del full_pillar_load[key][filter_key]
+
+            convert_data = self._convert_to_str(full_pillar_load, "")
+
+            with open(CONFSTORE_CLUSTER_CONFIG, "w") as file_value:
```
andkononykhin2, line 82: that's ok, but better to use pathlib.Path to avoid mixinf of old (os) and new approaches

andkononykhin2 (approved):

PR #848, opened 2021-01-25
[EOS-16848] fixes case of cached pillar during the update

andkononykhin2:
- re-applies provisioner configuration logic in scope of the update to ensure that updated pillar state is respected


api/python/provisioner/commands/__init__.py
```diff
@@ -808,6 +808,10 @@ def run(self, targets):  # noqa: C901 FIXME
                 try:
                     _update_component('provisioner', targets)
 
+                    # re-apply provisioner configuration to ensure
+                    # that updated pillar is taken into account
+                    _apply_provisioner_config(targets)
```
ypise, line 813: @andkononykhin2 
We also need to update `components.provisioner.config` to refresh pillar, which is being missed before encryption call `components.system.config.pillar_encrypt` (essentially missed in the CSM secret update case)

ypise (changes requested):

api/python/provisioner/commands/__init__.py
ypise: > @andkononykhin2 
> 
> We also need to update `components.provisioner.config` to refresh pillar, which is being missed before encryption call `components.system.config.pillar_encrypt` (essentially missed in the CSM secret update case)

My doubts were cleared during our discussion.

ypise (approved):

api/python/provisioner/commands/__init__.py
andkononykhin2: Thanks

PR #849, opened 2021-01-26
EOS-16399: Network config fix.

ypise:


ypise:
@83bhp 
We need this fix irrespective of EOS-16399.
Also, I've added code to parameterize MTU values.

sradhanand-pati (approved):

PR #851, opened 2021-01-27
[EOS-16869][EOS-17235] Teardown Provisioner Bootstrapped Environment and generic destroy API

mazinamdar:
Used salt-ssh :
1. Stop salt services
2. Unmount gluster volumes
3. remove bricks and volume
4. stop gluster services
5. remove salt config dir
6. remove prvsnr packages
7. remove profile dir
8. remove passwordless ssh config


api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+        setup_type = run_args.setup_type
+        targets = run_args.targets
+        states = deploy_states[states_group]
+        stages = run_args.stages
+        primary = self._primary_id()
+        secondaries = f"not {primary}"
+        nodes = []
+
+        if states_group == 'prvsnr' and targets == ALL_MINIONS:
+            node_list = PillarKey('cluster/node_list')
```
andkononykhin2, line 117: @ypise I guess it is going to be changed and `cluster/node_list` won't be available. Could you update on that?

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+            nodes.append(targets)
+
+        # apply states
+        for state in states:
+            if setup_type == SetupType.SINGLE:
+                self._apply_state(f"components.{state}", primary, stages)
+            else:
+                self._apply_state(f"components.{state}", targets, stages)
+
+        if states_group == 'prvsnr':
```
andkononykhin2, line 131: @mazinamdar could add few comments regarding the logic here and before

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+@attr.s(auto_attribs=True)
+class DestroyVM(Deploy):
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+    _run_args_type = run_args_type
+
+    def _remove_prvsnr(self, nodes):
+        for node in nodes:
+            if local_minion_id() == node:
+                logger.info(f"remove cortx-prvsnr rpms on {node}")
+                run_subprocess_cmd(
```
andkononykhin2, line 69: why we don't use salt state for that? (here and everywhere below)

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+                    "cortx-prvsnr python36-cortx-prvsnr")
+
+    def _remove_passwordless_ssh(self, nodes):
+        for node in nodes:
+            logger.info(f"Remove passwordless configuration with {node}")
+            run_subprocess_cmd(
+                f"rm -rf /root/.ssh")
+
+    def _remove_salt(self, nodes):
+        salt_cache_dir = '/var/cache/salt'
```
andkononykhin2, line 85: in case of glusterfs `/var/cache/salt/master/jobs` would be a mount point. so we should umount that first. Logic should likely be:
- stop/uninstall salt
- umount glusterfs
- cleanup salt dirs
- stop and cleanup gluster

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+
+        if self._is_hw():
+            raise errors.ProvisionerError(
+                "The command is specifically for VM teardown. "
+            )
+
+        if run_args.states is None:  # all states
+            self._run_states('prvsnr', run_args)
+
+        else:
```
andkononykhin2, line 151: `elif` should be good here

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+            if 'prvsnr' in run_args.states:
```
andkononykhin2, line 152: why we consider other values here? if it is going to be expanded in future please add `else` block where `NotImplementedError` is raised

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+deploy_states = dict(
+    prvsnr=[
+        "system.storage.glusterfs.teardown.volume_remove",
+        "system.storage.glusterfs.teardown.remove_dir",
+        "system.storage.glusterfs.teardown.services",
+        "system.storage.glusterfs.teardown.package_remove"
+    ]
+)
+
+run_args_type = build_deploy_run_args(deploy_states)
```
andkononykhin2, line 57: `stages` attr won't make any sense here since only teardown is considered. makes sense to inherit and re-define:
```
@attr.s(auto_attribs=True)
class XXX(run_args_type):
   stages: str = attr.ib(init=False, default=None)
```

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+            node_list = PillarKey('cluster/node_list')
+            pillar = PillarResolver(config.LOCAL_MINION).get([node_list])
+            pillar = next(iter(pillar.values()))
+            nodes = pillar[node_list]
+        else:
+            nodes.append(targets)
+
+        # apply states
+        for state in states:
+            if setup_type == SetupType.SINGLE:
```
andkononykhin2, line 126: in case of provisioner gluster would be cleaned up here and salt - later, but it's a wrong order. teardown should go in opposite way comparing to deployment. please consider to implement states for all and list them in deploy_states

andkononykhin2 (dismissed):

api/python/provisioner/commands/destroy_vm.py
mazinamdar: Salt will not be available at that time that's why used subprocess to run commands

api/python/provisioner/commands/destroy_vm.py
mazinamdar: If we remove salt first then we cannot use salt states to teardown glusterfs. We need to use manual steps/commands to unmount, remove volume , remove bricks dir etc.

I followed steps given in Wiki:
`https://github.com/Seagate/cortx-prvsnr/wiki/Auto-Deploy-VM-Provisioner-CLI-Commands#teardown-provisioner-bootstrapped-environment`

api/python/provisioner/commands/destroy_vm.py
mazinamdar: tomorrow I will start writing logic for remaining components, till now this is partial logic only for teardown of provisioner setup.

api/python/provisioner/commands/destroy_vm.py
ypise: Yes, that's correct.
The srvnode-N keys need to be iterated as there would be no `cluster/node_list`

api/python/provisioner/commands/destroy_vm.py
ypise: For other components, we can wait.
Let's cleanup existing teardown first with existing shell script may be.

api/python/provisioner/commands/destroy_vm.py
ypise: Aren't states part of core logic?
API states don't have states as I remember.
Bootstrap teardown should be purely API teardown states

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+        "system.storage.glusterfs.teardown.services",
+        "system.storage.glusterfs.teardown.package_remove"
+    ]
+)
+
+run_args_type = build_deploy_run_args(deploy_states)
+
+
+@attr.s(auto_attribs=True)
+class DestroyVM(Deploy):
```
ypise, line 61: Let's make this Generic.
`DestoryNode`

api/python/provisioner/commands/destroy_vm.py
ypise: Also, I have a doubt. Should we allow gluster on single node?
If not, then we just have to operate on salt packages.

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,157 @@
...
+    build_deploy_run_args,
+    Deploy
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+deploy_states = dict(
+    prvsnr=[
```
ypise, line 49: In my opinion it would be good to move these to :
`srv/components/system/storage/glusterfs/teardown/init.sls`

Then just make a call to `"system.storage.glusterfs.teardown"` here.

ypise (changes requested):

srv/components/system/storage/glusterfs/teardown/services.sls
```diff
@@ -0,0 +1,31 @@
+#
```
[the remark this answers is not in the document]
mazinamdar: Done

srv/components/system/storage/glusterfs/teardown/remove_dir.sls
```diff
@@ -0,0 +1,35 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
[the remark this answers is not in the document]
mazinamdar: Done

api/python/provisioner/commands/destroy_vm.py
mazinamdar: Done

api/python/provisioner/commands/destroy_vm.py
mazinamdar: We need to unmount from all nodes first then we do delete volumes.

api/python/provisioner/commands/destroy_vm.py
mazinamdar: updated logic

api/python/provisioner/commands/destroy_vm.py
mazinamdar: Added NotImplementedError

api/python/provisioner/commands/destroy_vm.py
mazinamdar: comment is addressed

api/python/provisioner/commands/destroy_vm.py
mazinamdar: Added logic to ignore glusterfs in case of single node

api/python/provisioner/commands/destroy_vm.py
mazinamdar: splitted into 2 things
1. volume_remove -> unmount and remove volume
2. init.sls -> remove bricks, stop services , remove rpms

srv/components/system/storage/glusterfs/teardown/init.sls
```diff
@@ -0,0 +1,22 @@
...
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+include:
+  - components.system.storage.glusterfs.teardown.remove_bricks
+  - components.system.storage.glusterfs.teardown.stop
```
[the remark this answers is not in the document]
mazinamdar: Done

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,311 @@
...
+            self._run_states('iopath', run_args)
+            self._run_states('prereq', run_args)
+            self._run_states('system', run_args)
+            self._run_states('bootstrap', run_args)
+
+        else:
+            if 'bootstrap' in run_args.states:
+                logger.info("Teardown Provisioner Bootstrapped Environment")
+                self._run_states('bootstrap', run_args)
+            if 'system' in run_args.states:
```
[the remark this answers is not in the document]
mazinamdar: Done

srv/components/system/storage/glusterfs/teardown/volume_remove.sls
```diff
@@ -28,8 +28,8 @@ glusterfs_volume_dir_{{ volume['mount_dir'] }}_unmount:
 # remove gluster volume
 glusterfs_volume_{{ volume['name'] }}_removed:
   module.run:
-    - glusterfs.delete_volume:
-      - target: {{ volume['name'] }}
+    - name: glusterfs.delete_volume
+    - target: {{ volume['name'] }}
```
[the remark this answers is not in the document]
mazinamdar: comment is addressed

ypise (approved):
Looks good

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,329 @@
...
+            c_path=c_path,
+            roster_file=roster_file,
+            ssh_options=ssh_options
+        )
+
+    @staticmethod
+    def remove_dir(list_dir, nodes, setup_type):
+        for node in nodes:
+            for di in list_dir:
+                if setup_type == SetupType.SINGLE:
```
andkononykhin2, line 148: can we use salt-ssh here as well?

andkononykhin2 (commented):
@mazinamdar It is a big change. I got general idea but it's hard to follow all the details and all the cases.

I have two options in mind:
- merge that once you are sure it works
- re-think that better, in particular:
  - how to avoid any shell and raw ssh routine (e.g. removal of pub key from authorized keys on remote servers might the final step)
  - how to avoid so many conditionals for both hw and vm and make the logic clearer
  - how to avoid copy-paste places

andkononykhin2 (commented):
updating the review status

api/python/provisioner/api_spec.yaml
```diff
@@ -135,3 +135,7 @@ deploy_jbod:
   type: DeployJBOD
 deploy_dual:
   type: DeployDual
+
+# destroy commands
+destroy_vm:
```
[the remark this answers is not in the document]
mazinamdar: # destroy commands
destroy:
  type: Destroy

for now I have added only destroy as common API for vm as well as hw

api/python/provisioner/commands/destroy_hw.py
```diff
@@ -0,0 +1,252 @@
...
+                if state not in (
+                    "system.storage.glusterfs.teardown.volume_remove",
+                    "system.storage.glusterfs.teardown",
+                    "provisioner.salt.teardown.package_remove"
+                ):
+                    logger.info(f"Applying '{state}' on {primary}")
+                    self._apply_states(state, primary)
+
+                    if state == "provisioner.passwordless_remove":
+                        list_dir = ["/var/cache/salt/"]
```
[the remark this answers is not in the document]
mazinamdar: Not directly related but passwordless is last step so we need to remove few directories before end, I have faced one issue to remove salt cache dir using salt-ssh command so using subprocess to remove this dir.

api/python/provisioner/commands/destroy_hw.py
```diff
@@ -0,0 +1,252 @@
...
+                if state in (
+                    "system.storage.glusterfs.teardown.volume_remove",
+                    "provisioner.teardown",
+                    "provisioner.passwordless_remove",
+                    "system.storage.multipath.teardown"
+                    "misc_pkgs.openldap.teardown",
+                    "misc_pkgs.rabbitmq.teardown"
+                ):
+                    if state == 'provisioner.passwordless_remove':
+                        list_dir = []
```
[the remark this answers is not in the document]
mazinamdar: Few dir path I am getting from config.py like 'profile_base_dir' that's the reason I have not used subprocess to remove them before passwordless step after that we need provide password for each command.

srv/components/system/storage/glusterfs/teardown/volume_remove.sls
```diff
@@ -15,21 +15,28 @@
...
+{% for volume in salt['pillar.get']('gluster:volumes', []) %}
 
 # unmount gluster volume
 glusterfs_volume_dir_{{ volume['mount_dir'] }}_unmount:
   mount.unmounted:
     - name: {{ volume['mount_dir'] }}
     - persist: True
 
+# remove mount dir from fstab
+Remove mount_{{ volume['mount_dir'] }}_present_in_fstab:
```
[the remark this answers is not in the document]
mazinamdar: This is extra check added as Yash asked to me add this.

srv/components/system/storage/glusterfs/teardown/init.sls
```diff
@@ -0,0 +1,21 @@
...
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+include:
+  - components.system.storage.glusterfs.teardown.stop
+  - components.system.storage.glusterfs.teardown.remove_bricks
```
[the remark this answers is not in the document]
mazinamdar: In destroy_vm or destroy_hw API I have called first unmount and volume remove state then this glusterfs.teardown will be called.

As we need to unmount on all nodes first then we go for volume remove. so kept that step as separate

pillar/components/gluster.sls
```diff
@@ -1,4 +1,4 @@
-glusterfs:
+gluster:
```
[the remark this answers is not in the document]
mazinamdar: In deployment we have glusterfs pillar which is slightly different from this. So to avoid overriding I have renamed it.

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1301,7 +1304,7 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
 
         if not run_args.field_setup:
             logger.info("Preparing factory profile")
-            for path in ('srv/salt', 'srv/pillar', '.ssh'):
+            for path in ('srv/config', 'srv/salt', 'srv/pillar', '.ssh'):
```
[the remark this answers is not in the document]
mazinamdar: I thought by just adding /opt/seagate/cortx/provisioner I can use existing roster and master config as it is , that's why not created dynamically again.

Please suggest me, do I need to create roster and other config again in destroy API with only required pillar and  states path?

api/python/provisioner/commands/destroy_hw.py
```diff
@@ -0,0 +1,252 @@
...
+                        list_dir.append(str(config.profile_base_dir().parent))
+                        # list_dir.append(config.CORTX_ROOT_DIR)
+                        DestroyNode.remove_dir(list_dir,
+                                               secondaries,
+                                               setup_type)
+                        DestroyNode.remove_dir(list_dir, [primary], setup_type)
+
+                    logger.info(f"Applying '{state}' on {secondaries}")
+                    # Execute first on secondaries then on primary.
+                    self._apply_states(state, '|'.join(secondaries))
```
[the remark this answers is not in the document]
mazinamdar: comment is addressed

api/python/provisioner/config.py
```diff
@@ -38,10 +38,12 @@
 #  - then rename base.py to config.py
 #  - remove PRVSNR_ prefix
 
+CORTX_ROOT_DIR = Path('/opt/seagate/cortx')
 CORTX_CONFIG_DIR = Path('/opt/seagate/cortx_configs')
 CONFSTORE_CLUSTER_CONFIG = CORTX_CONFIG_DIR / 'provisioner_cluster.json'
 
 PRVSNR_ROOT_DIR = Path('/opt/seagate/cortx/provisioner')
+PRVSNR_TMP_DIR = Path('/tmp/prvsnr')
```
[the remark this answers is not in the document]
mazinamdar: used /tmp/cortx-prvsnr as temp dir

api/python/provisioner/profile.py
```diff
@@ -66,11 +66,12 @@ def setup(
             'base': [str(d.resolve()) for d in add_file_roots] + [
                 str(profile_paths['salt_fileroot_dir']),
                 str(profile_paths['salt_factory_fileroot_dir']),
+                str(config.PRVSNR_FILEROOT_DIR),
```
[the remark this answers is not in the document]
mazinamdar: please correct me if I am wrong,
During destroy I will not be having details of setup_name, so I can not use profile setup in destroy that's why not used add_file_roots or add_pillar_roots
'
paths = config.profile_paths(
            config.profile_base_dir(
                location=setup_location, setup_name=setup_name
            )
        )
`

I am getting roster and master config directly from PRVSNR_FACTORY_PROFILE_DIR path.

api/python/provisioner/commands/destroy_vm.py
```diff
@@ -0,0 +1,329 @@
...
+                if state not in (
+                    "system.storage.glusterfs.teardown.volume_remove",
+                    "system.storage.glusterfs.teardown",
+                    "provisioner.salt.teardown.package_remove"
+                ):
+                    logger.info(f"Applying '{state}' on {primary}")
+                    self._apply_states(state, primary)
+
+                    if state == "provisioner.passwordless_remove":
+                        list_dir = ["/var/cache/salt/"]
```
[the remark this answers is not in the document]
mazinamdar: Will update in improvement

api/python/provisioner/commands/destroy_vm.py
mazinamdar: Will update in improvement

ypise (approved):

mazinamdar:
Improvement JIRA ticket : https://jts.seagate.com/browse/EOS-17613

api/python/provisioner/profile.py
```diff
@@ -66,11 +66,12 @@ def setup(
...
                 str(config.BUNDLED_SALT_FILEROOT_DIR)
             ]
         },
         'pillar_roots': {
-            'base': [
+            'base': [str(config.PRVSNR_PILLAR_DIR)] + [
```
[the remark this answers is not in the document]
mazinamdar: please correct me if I am wrong,
During destroy I will not be having details of setup_name, so I can not use profile setup in destroy that's why not used add_file_roots or add_pillar_roots
'
paths = config.profile_paths(
config.profile_base_dir(
location=setup_location, setup_name=setup_name
)
)
`

I am getting roster and master config directly from PRVSNR_FACTORY_PROFILE_DIR path.

api/python/provisioner/commands/destroy_vm.py
mazinamdar: comment is addressed

api/python/provisioner/commands/destroy_vm.py
mazinamdar: updated the logic to remove salt and profile dir.

cortx-admin:

api/python/provisioner/commands/setup_provisioner.py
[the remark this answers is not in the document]
andkononykhin2: we copy factory data (setup command input arguments, pillars) just to have the same context for later setups (e.g. when we replace a node): we won't need to ask a user to provide us these values but just reuse the same to get the same logic and result env.

But roster and configs are generated on the fly since they are jump server specific (since paths are there). And it is not a big deal to generate them.

>Please suggest me, do I need to create roster and other config again in destroy API with only required pillar and states path?

if you are asking about initial stage of the destroy logic: possibly. I guess you need that for salt-ssh client (roster and master config). If are sure that all paths inside roster and master config would be ok regtardless of a node where Destory is called then your implementation might work. If you are not sure then you need to create these files on the fly. You may check how ReplaceNode calls SetupProvisioner:
  - Replace [prepares profile](https://github.com/Seagate/cortx-prvsnr/blob/ddaef30c7836f2bbb9fc5c44d812a7a6c1ffd149/api/python/provisioner/commands/replace_node.py#L131-L133) from factory one. 
  - Then it [resolves nodes specs](https://github.com/Seagate/cortx-prvsnr/blob/ddaef30c7836f2bbb9fc5c44d812a7a6c1ffd149/api/python/provisioner/commands/replace_node.py#L192-L203). You may consider to build nodes node in another way actually (you would just need to get IP per miinion id and create a list of `Node` class instances)
  - then roster file can be created : you need to pass only the list of nodes and a path to the key which is `/root/.ssh/id_rsa_prvsnr`

To summarize: to create roster you need to have only a list of nodes. You may actually re-use .ssh/config and build roster from it.

To build master config you should call profile setup API like [here](https://github.com/Seagate/cortx-prvsnr/blob/ddaef30c7836f2bbb9fc5c44d812a7a6c1ffd149/api/python/provisioner/commands/setup_provisioner.py#L1151-L1155).

api/python/provisioner/profile.py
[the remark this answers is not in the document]
andkononykhin2: As I mentioned in another comment, please check [logic of ReplaceNode](https://github.com/Seagate/cortx-prvsnr/blob/ddaef30c7836f2bbb9fc5c44d812a7a6c1ffd149/api/python/provisioner/commands/replace_node.py#L141-L164). Having that `paths` you can call [profile.setup](https://github.com/Seagate/cortx-prvsnr/blob/ddaef30c7836f2bbb9fc5c44d812a7a6c1ffd149/api/python/provisioner/commands/setup_provisioner.py#L1151-L1155) to prepare/update necessary salt files inside it. And you can pass `add_file_roots` (e.g. as `add_file_roots=[config.PRVSNR_FILEROOT_DIR]`) if it's really needed.

api/python/provisioner/profile.py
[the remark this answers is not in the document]
andkononykhin2: please check my response for similar file roots related question

pillar/components/gluster.sls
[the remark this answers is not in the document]
andkononykhin2: please add a comment here and mentioned these in a follow-up task as well

andkononykhin2 (changes requested):
please check my comments. Thanks

andkononykhin2:
@mazinamdar As long as #932 has been merged can we follow the same approach for glusterfs to avoid copy-paste between api and core states?

ypise (approved):

ypise:
No longer required.

PR #854, opened 2021-01-28
EOS-12714: Provisioner to install confluent python kafka client in LDR-R2

sradhanand-pati:


ypise (changes requested):
Ensure we cover uninstall in teardown.

PR #856, opened 2021-01-29
EOS-17120: Minor fix in setup.yaml read logic

ypise:


nitin-seagate (approved):
Looks good to me

PR #862, opened 2021-02-02
EOS-17096: Mock relations with CORTX components

dmitrydb:
- Parameterize package names and versions of cortx components
- Create the mock script that will echo/logging components names and mini provisioner API command name
- Create `setup.yml` for each cortx component depended on mini-provisioner API that will invoke the mock script
- Store testing configuration in `misc_pkgs/mocks/cortx` to safely execute salt states


srv/components/misc_pkgs/mocks/cortx/files/setup.mock.yaml/setup.yaml
```diff
@@ -0,0 +1,48 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
ypise, line 2: Can we use '-' in directory name instead of '.'?
'.' notation at times could be misleading.

srv/components/misc_pkgs/mocks/cortx/files/scripts/mock.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+
+import sys
+import logging
+
+FORMAT = '%(asctime)s-%(message)s'
+LOGGING_FILE = '/tmp/mock.log'
+
+
+def main():
```
ypise, line 10: In https://github.com/Seagate/cortx-prvsnr/blob/f256801af97a79255f33031b1cb7400bb741c8a6/srv/_modules/setup_conf.py#L82 we check if the command has a `--help` opion.
Do simulate the behavior to avoid failure.

srv/components/misc_pkgs/mocks/cortx/install.sls
```diff
@@ -0,0 +1,48 @@
...
+    - name: cortx_mock_repo
+    - humanname: CORTX Mock repo
+{# TODO: use path from pillars or any other configuration #}
+    - baseurl: file:///opt/seagate/cortx/provisioner/srv/components/misc_pkgs/files/cortx.mock.repo
+    - enabled: True
+    - gpgcheck: 0
+
+
+/usr/local/bin/mock:
+  file.managed:
```
ypise, line 29: This is no more necessary.
The setup.yaml:cmd should have full path to command executable instead.

ypise (changes requested):

srv/components/misc_pkgs/mocks/cortx/files/setup.mock.yaml/setup.yaml
dmitrydb: I agree. It was just as initial design structure. I fixed that

srv/components/misc_pkgs/mocks/cortx/files/scripts/mock.py
dmitrydb: I think it is not needed here. When we execute the `mock.py` with the `--help` option, the script will log `--help` argument to the log file and will exit with a non-error exit code.

srv/components/misc_pkgs/mocks/cortx/install.sls
dmitrydb: I copy mock script to avoid long paths in setup.yaml. Because the full path to `mock.py` is 
`/opt/seagate/cortx/provisioner/srv/components/misc_pkgs/mocks/cortx/files/scripts/mock.py`

andkononykhin2 (commented):
@dmitrydb  Looks good. But the most unclear thing is how to apply that. I don't see any automation logic that will build rpms + creates repo (they are separate scripts). So could you provide some readme file here to guideline how to use that. Thanks

ypise (approved):

PR #875, opened 2021-02-05
EOS-16927: New API for upgrade: adjust rollback logic for R2 (singlenode)

SomwanshiAnjali:


andkononykhin2 (commented):
@SomwanshiAnjali Generally looks pretty well. The following would be good to add:
- docstrings for new class (please add notes here regarding pillar structure assumptions)
- test scenarios (how it is being tested)

api/python/provisioner/commands/__init__.py
```diff
@@ -685,6 +711,71 @@ def run(self, targets):  # noqa: C901 FIXME
...
+        rollback_pillar = PillarResolver(local_minion).get([rollback_path])
+        rollback_pillar = rollback_pillar[local_minion][rollback_path]
+        cortx_tgt_version = (
+            target_version
+            if target_version in rollback_pillar.keys()
+            else None
+        )
+
+        if cortx_tgt_version is None:
+            logger.exception(
```
andkononykhin2, line 732: that makes sense only inside `catch` block. I guess we need to raise something here instead. If you want to log as well please use logger.error or warning depending on the severity.

api/python/provisioner/commands/__init__.py
```diff
@@ -685,6 +711,71 @@ def run(self, targets):  # noqa: C901 FIXME
...
+                        _rollback_component(component, targets)
+                    logger.info(
+                        "Configurations restored successfully "
+                        f"on target {target}"
+                    )
+            except Exception as exc:
+                logger.exception(
+                    'failed to rollback sw stack, reason {}'
+                    .format(exc)
+                )
```
andkononykhin2, line 776: please add `raise` here to propagate or raise some specific exception

andkononykhin2 (changes requested):

api/python/provisioner/commands/__init__.py
SomwanshiAnjali: Addressed

api/python/provisioner/commands/__init__.py
SomwanshiAnjali: addressed

SomwanshiAnjali:
@andkononykhin2 
Does the order of components matter while invoking components rollback.sls?

api/python/provisioner/commands/__init__.py
```diff
@@ -685,6 +712,68 @@ def run(self, targets):  # noqa: C901 FIXME
...
+                            f"yum history rollback -y {txn_id}",
+                            targets=target
+                        )
+                        logger.info(
+                            f"Rollback on target {target} is completed"
+                        )
+
+                    logger.info('Restoring configurations for components')
+                    for component in (
+                        'motr',
```
nitin-seagate, line 760: Why do we hard-code these values? Can we pull it from static defined global list which maintains components...

api/python/provisioner/__init__.py
```diff
@@ -42,6 +42,7 @@
     set_swupdate_repo,
     sw_update,
     fw_update,
+    rollback_update,
```
nitin-seagate, line 45: Should we change this to
**sw_rollback**

api/python/provisioner/__init__.py
```diff
@@ -78,6 +79,7 @@
     'set_network',
     'set_swupdate_repo',
     'sw_update',
+    'rollback_update',
```
nitin-seagate, line 82: Should we change this to
**sw_rollback**

api/python/provisioner/_api_cli.py
```diff
@@ -174,6 +174,7 @@ def auth_init(username, password, eauth='pam'):
     'set_network',
     'set_swupdate_repo',
     'sw_update',
+    'rollback_update',
```
nitin-seagate, line 177: Should we change this to
**sw_rollback**

api/python/provisioner/api.py
```diff
@@ -296,6 +296,33 @@ def sw_update(targets=ALL_MINIONS, nowait=False):
     )
 
 
+def rollback_update(target_version, targets=ALL_MINIONS, nowait=False):
```
nitin-seagate, line 299: Guess we should change this to
**sw_rollback**

api/python/provisioner/api.py
```diff
@@ -296,6 +296,33 @@ def sw_update(targets=ALL_MINIONS, nowait=False):
...
+    target_version
+        Rollback target old version
+    targets
+        (optional) A host to rollback. Default: all minions
+    nowait
+        (optional) Run asynchronously. Default: False
+    """
+
+    return _api_call(
+        'rollback_update', target_version, targets=targets, nowait=nowait
```
nitin-seagate, line 322: Should we change this to
**sw_rollback**

api/python/provisioner/api_spec.yaml
```diff
@@ -91,6 +91,8 @@ sw_update:
   type: SWUpdate
 fw_update:
   type: FWUpdate
+rollback_update:
```
nitin-seagate, line 94: Should we change this to
**sw_rollback**

api/python/provisioner/api_spec.yaml
```diff
@@ -91,6 +91,8 @@ sw_update:
...
+  type: RollbackUpdate
```
nitin-seagate, line 95: **SwRollback**

api/python/provisioner/commands/__init__.py
```diff
@@ -33,6 +33,7 @@
     PillarSetError,
     SWUpdateError,
     SWUpdateFatalError,
+    SWStackRollbackError,
```
nitin-seagate, line 36: SWRollbackError

test/r2.txt
```diff
@@ -77,4 +77,12 @@ Step-4: Check that all yum repositories are listed in `$ yum repolist enabled` o
...
-         ```
\ No newline at end of file
+         ```
+
+[EOS-16927]: New API for upgrade: adjust rollback logic for R2 (singlenode)
+
+Step-1: Ensure rollback pillar data is present/updated as part of sw_update command
+        `$salt-call pillar.get rollback`
+
+Step-2: Run rollback_update api
```
nitin-seagate, line 87: sw_rollback

nitin-seagate (commented):
@andkononykhin Really appreciate if you can reaffirm my comments (if in-line with your expectations)

api/python/provisioner/commands/__init__.py
andkononykhin2: @nitin-seagate That thing is not only for rollback and something that we should have been addressed. But let's do that as a separate task.

api/python/provisioner/__init__.py
andkononykhin2: @SomwanshiAnjali Agree. It sounds reasonable. Could you adjust here and in similar places?

andkononykhin2:
> @andkononykhin2 @nitin-seagate
> Does the order of components matter while restoring config's for each component through rollback.sls?

@SomwanshiAnjali the scope of the current PR is singlenode disruptive update. So no specific order is implied. I suggest only to do that in a reversed order comparing to update.

api/python/provisioner/commands/__init__.py
andkononykhin2: [EOS-17416](https://jts.seagate.com/browse/EOS-17416) is to cover the current concern. Let's merge the PR and address that in scope of the new task then.

SomwanshiAnjali: Will address as part of EOS-17416

api/python/provisioner/__init__.py
SomwanshiAnjali: addressed

api/python/provisioner/__init__.py
SomwanshiAnjali: addressed

api/python/provisioner/_api_cli.py
SomwanshiAnjali: addressed

api/python/provisioner/api.py
SomwanshiAnjali: addressed

api/python/provisioner/api.py
SomwanshiAnjali: addressed

api/python/provisioner/api_spec.yaml
SomwanshiAnjali: addressed

api/python/provisioner/api_spec.yaml
SomwanshiAnjali: addressed

api/python/provisioner/commands/__init__.py
SomwanshiAnjali: addressed

test/r2.txt
SomwanshiAnjali: addressed

api/python/provisioner/commands/__init__.py
```diff
@@ -685,6 +698,82 @@ def run(self, targets):  # noqa: C901 FIXME
...
+                    else:
+                        logger.info(f"Starting rollback on target {target}")
+                        salt_cmd_run(
+                            f"yum history rollback -y {txn_id}",
+                            targets=target
+                        )
+                        logger.info(
+                            f"Rollback on target {target} is completed"
+                        )
+
```
andkononykhin2, line 757: @SomwanshiAnjali I've realized we forgot about rollback routine around provisioner itself. Please:
- add [these few lines](https://github.com/Seagate/cortx-prvsnr/blob/7bf919462b747634d54d4f5c6978f94b1ec40f0f/api/python/provisioner/commands/__init__.py#L637-L642) here
- create a task in Jira to align that reconfiguration with other components: we need to follow the same approach as for any other cortx component but in that case it might be not so trivial

andkononykhin2 (changes requested):

api/python/provisioner/commands/__init__.py
```diff
@@ -685,6 +698,82 @@ def run(self, targets):  # noqa: C901 FIXME
...
+                            f"yum txn id not available for {target}"
+                        )
+                    else:
+                        logger.info(f"Starting rollback on target {target}")
+                        salt_cmd_run(
+                            f"yum history rollback -y {txn_id}",
+                            targets=target
+                        )
+                        logger.info(
+                            f"Rollback on target {target} is completed"
```
andkononykhin2, line 755: Please adjust the message to `Yum rollback...` since at that moment rollback is not fully completed yet, only in scope of yum packages. Later when everything is done completely for a node makes sense to add some INFO level message (like that one). And one more for the full set of nodes at the final end.

andkononykhin2 (changes requested):

api/python/provisioner/commands/__init__.py
SomwanshiAnjali: Addressed and created EOS-17523

api/python/provisioner/commands/__init__.py
SomwanshiAnjali: Addressed

andkononykhin2 (approved):
@SomwanshiAnjali Thanks. Looks good now. But please consider the following as a follow-up (might be addressed in scope of [EOS-17542](https://jts.seagate.com/browse/EOS-17542)):

- separate API into a python module
- make it reusable so it can be called properly from sw_upgrade command in case of failures

PR #877, opened 2021-02-08
EOS-16670: Convert controller-cli bash scripts to python scripts

rahuljyoti65:

### DIFF
--- a/srv/components/controller/files/python-scripts/xml.py
+++ b/srv/components/controller/files/python-scripts/xml.py
@@ -1,0 +1,91 @@
+import logging
+import os
+import subprocess
+import sys
+import xml.etree.ElementTree as ET
+
+logger = logging.getLogger(__name__)
+
+log_dir = '/var/log/seagate/provisioner/'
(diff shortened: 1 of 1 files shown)


83bhp:
Had one round of review, looks good, let's try to run it together to test various scenarios.
Also please fix the COdacy reported issues.

srv/components/controller/files/python-scripts/xml.py
```diff
@@ -0,0 +1,91 @@
+import logging
```
ypise, line 1: `srv/components/controller/files/python-scripts/xml.py`
`python-scripts` is not necessary, just call it `scripts`

Also, rename the file to: `srv/components/controller/files/scripts/xml_parser.py`

Do we really need this file? I feel we don't need to have a separate parser module in Python, we can use standard xml parser.

Secondly, we don't need xml as we can capture output as JSON. @83bhp need your help with JSON

ypise (changes requested):

cortx-admin:

PR #881, opened 2021-02-08
python36-cortx-prvsnr dependencies added to setup.py

ypise:

### DIFF
--- a/api/python/setup.py
+++ b/api/python/setup.py
@@ -85,6 +85,8 @@ setup(
         ],
     },
     install_requires=[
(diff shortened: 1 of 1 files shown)


api/python/setup.py
```diff
@@ -85,6 +85,8 @@
         ],
     },
     install_requires=[
+        'jsonschema',
+        'requests',
```
andkononykhin2, line 89: why we need that?

ypise:
This PR is not necessary.

PR #882, opened 2021-02-10
[EOS-16926]: Adjust upgrade logic for R2 (singlenode)

dmitrydb:
I created the new SW Upgrade API to use mini-provisioner API calls for components update.


api/python/provisioner/commands/swupgrade.py
```diff
@@ -0,0 +1,147 @@
...
+                # re-apply provisioner configuration to ensure
+                # that updated pillar is taken into account
+                _apply_provisioner_config(targets)
+
+                config_salt_master()
+
+                minion_conf_changes = config_salt_minions()
+
+                for component in COMPONENTS_FOR_UPGRADE:
+                    _update_component(component, local_minion_id())
```
andkononykhin2, line 103: why we target only local minion id but use `targets` for provisioner? let's keep that generic for now even if have the scope of singlenode

api/python/provisioner/commands/swupgrade.py
```diff
@@ -0,0 +1,147 @@
...
+
+            if rollback_error:
+                # unrecoverable state: SW stack is in intermediate state,
+                # no sense to start the cluster ??? verify TODO IMPROVE
+                logger.error(
+                    'Rollback failed: {}'.format(rollback_ctx.rollback_error))
+
+                final_error_t = SWUpdateFatalError
+            elif rollback_ctx is not None:
+                # yum packages are in initial state here
```
andkononykhin2, line 140: [that](https://github.com/Seagate/cortx-prvsnr/blob/7bf919462b747634d54d4f5c6978f94b1ec40f0f/api/python/provisioner/commands/__init__.py#L623-L680) errors processing logic should be kept generally since it's important for rollback. We need just to remove any binding with cluster management.

api/python/provisioner/commands/swupgrade.py
```diff
@@ -0,0 +1,147 @@
...
+        # TODO:
+        #   - create a state instead
+        #   - what about apt and other non-yum pkd managers
+        #   (downgrade is another more generic option but it requires
+        #    exploration of depednecies that are updated)
+        # TODO IMPROVE minions might be stopped here in case of rollback,
+        #      options: set up temp ssh config and rollback yum + minion config
+        #      via ssh as a fallback
+        rollback_ctx = None
+        try:
```
andkononykhin2, line 76: [that](https://github.com/Seagate/cortx-prvsnr/blob/7bf919462b747634d54d4f5c6978f94b1ec40f0f/api/python/provisioner/commands/__init__.py#L527-L531) logic of yum rollback in case of any errors is necessary for upgrade as well. Please restore.

api/python/provisioner/commands/swupgrade.py
```diff
@@ -0,0 +1,147 @@
...
+
+                minion_conf_changes = config_salt_minions()
+
+                for component in COMPONENTS_FOR_UPGRADE:
+                    _update_component(component, local_minion_id())
+            except Exception as exc:
+                raise SWStackUpdateError(exc) from exc
+
+            try:
+                ensure_cluster_is_healthy()
```
andkononykhin2, line 108: as long as we do not manage cluster on that level we don't start it, so that check doesn't make sense here

api/python/provisioner/commands/swupgrade.py
```diff
@@ -0,0 +1,147 @@
...
+        #   - create a state instead
+        #   - what about apt and other non-yum pkd managers
+        #   (downgrade is another more generic option but it requires
+        #    exploration of depednecies that are updated)
+        # TODO IMPROVE minions might be stopped here in case of rollback,
+        #      options: set up temp ssh config and rollback yum + minion config
+        #      via ssh as a fallback
+        rollback_ctx = None
+        try:
+            ensure_cluster_is_healthy()  # TODO: checker.run do that check too
```
andkononykhin2, line 77: we assume that a caller of the upgrade API does manage that cluster and cluster will likely be stopped here (up to external logic). So that check will likely fail and we don't need it here.

andkononykhin2 (changes requested):
@dmitrydb Thank you for the PR. General request: we need to shift our assumptions around the API which are:
- a caller is responsible for cluster management (start, stop. health check)
- we only do our job in scope of sw upgrade for a node without any cluster level acitivities
- rollback is required and might be kept as for sw update r1 logic

api/python/provisioner/commands/swupgrade.py
dmitrydb: Done

api/python/provisioner/commands/swupgrade.py
dmitrydb: Restore the previous logic

api/python/provisioner/commands/swupgrade.py
dmitrydb: Done. I restored the previous logic.

api/python/provisioner/commands/swupgrade.py
dmitrydb: I removed cluster health validation

api/python/provisioner/commands/swupgrade.py
dmitrydb: I fixed that.

api/python/provisioner/commands/sw_upgrade.py
```diff
@@ -0,0 +1,159 @@
...
+                # no sense to start the cluster ??? verify TODO IMPROVE
+                logger.error(
+                    'Rollback failed: {}'.format(rollback_ctx.rollback_error))
+
+                final_error_t = SWUpdateFatalError
+            elif rollback_ctx is not None:
+                if isinstance(
+                    # cluster is stopped here
+                    update_exc, (SWStackUpdateError,
+                                 ClusterMaintenanceDisableError,
```
andkononykhin2, line 136: The following 3 exceptions are not needed here since they won't happen.

andkononykhin2 (changes requested):
please update for one more place

api/python/provisioner/commands/sw_upgrade.py
dmitrydb: I agree. I fixed that

andkononykhin2 (approved):

PR #884, opened 2021-02-10
[EOS-17444] Setup HA on single node vm

mazinamdar:
Signed-off-by: mazinamdar mazhar.inamdar@seagate.com

-------------------------------------------------------------------------------------------------------------
| Action | Node to execute on |
|--------|-----------------------|
| post_install | All nodes |
| config | Primary or active node in cluster |
| init | All nodes |
| cleanup | Primary or active node in cluster |
| reset | All nodes |
| test | Primary or active node in cluster |

### DIFF
--- a/api/python/provisioner/commands/deploy_vm.py
+++ b/api/python/provisioner/commands/deploy_vm.py
@@ -58,7 +58,6 @@ deploy_states = dict(
     prereq=[
         "misc_pkgs.ssl_certs",
         "ha.haproxy",
-        "ha.haproxy.start",
         "misc_pkgs.openldap",
         "misc_pkgs.rabbitmq",
         "misc_pkgs.nodejs",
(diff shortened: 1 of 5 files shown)


api/python/provisioner/commands/deploy_vm.py
```diff
@@ -157,13 +162,16 @@ def _run_states(self, states_group: str, run_args: run_args_type):
             # TODO use salt orchestration
             if setup_type == SetupType.SINGLE:
                 logger.debug("Executing for single node.")
-                if "sync" not in state:
+                if "^sync" not in state:
```
andkononykhin2, line 165: @mazinamdar looks like a regex. Do you think it will do what you want? What about `startswith`?

andkononykhin2 (commented):
Generally looks good. Please just pay attention on one concern.

api/python/provisioner/commands/deploy_vm.py
mazinamdar: oops that is not working as expected, thanks, updated using `startswith`

andkononykhin2 (approved):

Issue #571, opened 2020-10-02
csm.common.errors.CsmSetupError: error(4102): CSM init failed while auto deployment for dual node cluster (VMs)


imvenkip:
INFO    Oct  2 11:17:35 ]: Installing 'components.csm' for both nodes
ERROR: Minions returned with non-zero exit code
For detailed error logs, please see: /var/log/seagate/provisioner/auto-deploy-cortx.log

 

***** FAILED!!*****
Detailed error log is kept at: /var/log/seagate/provisioner/auto-deploy-cortx.log
You have mail in /var/spool/mail/root

-----
following error observed from deploy log file

ID: Stage - Init CSM
    Function: cmd.run
        Name: __slot__:salt:setup_conf.conf_cmd('/opt/seagate/cortx/csm/conf/setup.yaml', 'csm:init')
      Result: False
     Comment: Command "/usr/bin/csm_setup init" run
     Started: 05:18:14.562688
    Duration: 11324.353 ms
     Changes:
              ----------
              pid:
                  10601
              retcode:
                  1
              stderr:
                  Traceback (most recent call last):
                    File "csm/conf/setup.py", line 763, in init
                    File "csm/conf/setup.py", line 453, in _logrotate
                  TypeError: 'NoneType' object is not subscriptable
                  csm_setup command failed: Traceback (most recent call last):
                    File "/var/jenkins/workspace/Cortx-Releases/Cortx-1.0/CSM-Agent/csm/conf/csm_setup.py", line 87, in <module>
                    File "/var/jenkins/workspace/Cortx-Releases/Cortx-1.0/CSM-Agent/csm/conf/csm_setup.py", line 65, in process
                  csm.common.errors.CsmSetupError: error(4102): CSM init : Fail error(4102): csm_setup init failed. Error: 'NoneType' object is not subscriptable - None
              stdout:

imvenkip:
And one more issue is observed. Before deployment, eth-1 and eth-2 have different ip addresses.  After deployment, both have same IP addresses in both nodes.


[root@ssc-vm-c-0983 ~]# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
[...]

PR #886, opened 2021-02-11
EOS-17036:Build #571: No Route to host error observed during s3 operations intermittently

sradhanand-pati:


ypise (approved):

ypise (approved):

cli/src/factory_ops/unboxing/init
```diff
@@ -245,7 +245,8 @@ echo "Ok." | tee -a ${LOG_FILE}
 
 # Re-configure firewall
 echo -n "Re-configuring firewall................................." | tee -a ${LOG_FILE}
-salt "*" state.apply components.system.firewall.config ${salt_opts}
+salt "*" state.apply components.system.firewall.teardown ${salt_opts}
+salt "*" state.apply components.system.firewall ${salt_opts}
```
ujjwalpl, line 249: What about config ? Is this a typo ?

ypise: It's not a typo. This is a correct command call.

ujjwalpl: > components.system.firewall.config is a subset of components.system.firewall. 
I assume components.system.firewall.teardown is also a subset of components.system.firewall. Is that correct ? 
If Yes, Will components.system.firewall lead to calling of components.system.firewall.teardown as well ?

PR #896, opened 2021-02-15
EOS-17565: Update version and license for python36-cortx-prvsnr package

SomwanshiAnjali:


api/python/provisioner/__metadata__.py
```diff
@@ -16,7 +16,7 @@
 #
 
 __title__ = 'cortx-prvsnr'
-__version__ = '0.42.0'
+__version__ = '2.42.0'
```
andkononykhin2, line 19: why that version?

andkononykhin2 (commented):
@SomwanshiAnjali What is a reason for setting 2.42.0. We haven't even ever set 1.0.0. That versioning is used SemVer approach  and such a change doesn't have any sense in scope of that.

api/python/provisioner/__metadata__.py
SomwanshiAnjali: @andkononykhin2 
It was request to follow naming convention for rpm's as `cortx-[component_name]-[version]-[bld]_[git_tag].[platform].rpm` 
and here version suggested to use is `2.0.0` 

but since provisioner version should also be maintained to track the changes, @ypise  suggested to use `2.42.0`

ypise: @andkononykhin2 ,
We need to discuss if 2.42.0 is better or 2.0.42 is.
But for now let's have 2.42.0. We have to align the versions to the one that's being followed for other rpms.

@SomwanshiAnjali we need to check if 2.42.0 is ok on the Jenkins test. I have a feeling this wouldn't suffice.

ypise (approved):

PR #908, opened 2021-02-18
[EOS-17823] Updated logic for getting nodes from pillar and need passwordless ssh…

mazinamdar:
… for single node

1. Need passwordless ssh for single node for validation API
2. Hosts entries should be present before Validation API 
3. fixed get nodes logic in validation API

### DIFF
--- a/api/python/provisioner/commands/auto_deploy.py
+++ b/api/python/provisioner/commands/auto_deploy.py
@@ -106,7 +106,13 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
             setup_ctx.ssh_client.cmd_run(
                 (
                     'salt-call state.apply '
-                    'components.system.config'
+                    'components.system.config.pillar_encrypt'
+                ), targets=setup_ctx.run_args.primary.minion_id
+            )
+            setup_ctx.ssh_client.cmd_run(
+                (
+                    'salt-call state.apply '
+                    'components.system.config.hosts'
                 )
             )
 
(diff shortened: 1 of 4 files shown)


api/python/provisioner/commands/auto_deploy.py
```diff
@@ -106,8 +106,8 @@ def run(self, nodes, **kwargs):
             setup_ctx.ssh_client.cmd_run(
                 (
                     'salt-call state.apply '
-                    'components.system.config.pillar_encrypt'
-                ), targets=setup_ctx.run_args.primary.minion_id
+                    'components.system.config'
```
andkononykhin2, line 109: pillar_encrypt here was called here to enrypt updated pillar once config ini is applied.
The whole system.config appliance would trigger [many other routines](https://github.com/Seagate/cortx-prvsnr/blob/6a8e2ce4a47f40dc55851dba71d75823819106d9/srv/components/system/config/init.sls#L19-L24).

More over later dual deploy will call the same but [after prepare and install](https://github.com/Seagate/cortx-prvsnr/blob/6a8e2ce4a47f40dc55851dba71d75823819106d9/srv/components/system/init.sls#L19-L21).

So, that change doesn't look clean. Could you clarify?
CC @ypise

andkononykhin2 (commented):
Please check my concern. Thanks

api/python/provisioner/commands/auto_deploy.py
mazinamdar: Actually I want to run components.system.config.hosts , @ypise suggested instead of running config.pillar_encrypt and config.hosts in 2 steps, it is safe to run complete system.config at this stage. so added complete system.config here.

ypise (approved):

api/python/provisioner/commands/auto_deploy.py
ypise: `So, that change doesn't look clean. Could you clarify?`
Yes it does and needs to be worked on. We need to split system based on the CES flow.

andkononykhin2 (approved):

PR #909, opened 2021-02-18
[EOS-17598] makes config states acceptable for rollback

andkononykhin2:


srv/components/ha/cortx-ha/config.sls
```diff
@@ -15,23 +15,17 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-include:
-  - components.ha.cortx-ha.install
 
 Run cortx-ha post_install:
   cmd.run:
     - name: __slot__:salt:setup_conf.conf_cmd('/opt/seagate/cortx/ha/conf/setup.yaml', 'ha:post_install')
-    - require:
```
ypise, line 24: Let's maintain these `require` statements to ensure installation.
Any specific reason this is removed here?

srv/components/s3server/config.sls
```diff
@@ -18,8 +18,6 @@
 Stage - Post Install S3Server:
   cmd.run:
     - name: __slot__:salt:setup_conf.conf_cmd('/opt/seagate/cortx/s3/conf/setup.yaml', 's3:post_install')
-    - require:
```
ypise, line 21: Let's maintain these require statements to ensure installation.
Any specific reason this is removed here?

srv/components/ha/cortx-ha/config.sls
```diff
@@ -15,23 +15,17 @@
...
-      - Install cortx-ha
 
 
 {% if "primary" in pillar["cluster"][grains["id"]]["roles"] %}
 
 Run cortx-ha config:
   cmd.run:
     - name: __slot__:salt:setup_conf.conf_cmd('/opt/seagate/cortx/ha/conf/setup.yaml', 'ha:config')
-    - require:
```
ypise, line 33: Let's maintain these require statements to ensure installation.
Any specific reason this is removed here?

srv/components/ha/cortx-ha/config.sls
```diff
@@ -44,5 +38,3 @@ No HA config on secondary node:
 Run cortx-ha init:
   cmd.run:
     - name: __slot__:salt:setup_conf.conf_cmd('/opt/seagate/cortx/ha/conf/setup.yaml', 'ha:init')
-    - require:
```
ypise, line 47: Let's maintain these require statements to ensure installation.
Any specific reason this is removed here?

ypise (changes requested):

srv/components/ha/cortx-ha/config.sls
andkononykhin2: We don't want to apply installation stages during configuration phases in scope of rollback (and it's not the only case, see below). Just few components have that logic and it is not consistent. We need to choose only one approach for all components:

- either call stages (prepare, install, config ....) separately and manage the flow ourselves (e.g. in init.sls or in api logic)
- or let salt to relate all stages so we can apply only a final one.

Keeping in mind the new workflow of the development I think the first one is more applicable for us.

srv/components/s3server/config.sls
andkononykhin2: please check the above response

srv/components/ha/cortx-ha/config.sls
andkononykhin2: please check the above response

srv/components/ha/cortx-ha/config.sls
andkononykhin2: please check the above response

cortx-admin:

srv/components/ha/cortx-ha/config.sls
ypise: Got it.
But then it has to be removed from all other component configs as well.

ypise (approved):

PR #916, opened 2021-02-22
[EOS-18037] Python API for generating roster file

mazinamdar:

Python API to generate roster file
`  --roster-path STR     roster file path (default: /var/lib/seagate/cortx/prov
                        isioner/shared/factory_profile/srv/roster)

  --user STR            user for salt-ssh (default: root)

  --port INT            port for salt-ssh (default: 22)

  --priv-key STR        private key path for salt-ssh (default: /var/lib/seaga
                        te/cortx/provisioner/shared/factory_profile/.ssh/setup
                        .id_rsa)
`
Used default location as factory_profile shared location, as file will be available on all nodes.(as discussed with @ypise )


api/python/provisioner/commands/generate_roster.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+
+
+@attr.s(auto_attribs=True)
+class GenerateRoster(CommandParserFillerMixin):
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+    _run_args_type = RunArgsGenerateRosterAttrs
+
+    @staticmethod
+    def _get_pillar_data(key):
```
ypise, line 83: Should not be a static method.
If we need a static method, let's create a utility class for all such methods that could be reusable for other classes.

api/python/provisioner/commands/generate_roster.py
```diff
@@ -0,0 +1,139 @@
...
+        """Provide map of node and hostname"""
+        host_node_map = {}
+        for node in GenerateRoster._get_nodes_list():
+            host = GenerateRoster._get_pillar_data(
+                       f'cluster/{node}/hostname')
+            host_node_map.update({node: host})
+        return host_node_map
+
+    @staticmethod
+    def _prepare_roster(
```
ypise, line 113: Should not be a static method as the call to this method would be rare. To allocate memory for this is unnecessarily overloading the resources.

api/python/provisioner/commands/generate_roster.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def run(self, **kwargs):
+
+        if Path(kwargs['roster_path']).is_file():
+            logger.warn("Roster file present, Updating roster file.")
+
+        if not Path(kwargs['priv_key']).is_file():
+            raise ValueError('Private key not present')
+
+        host_node_map = GenerateRoster._get_node_host_map()
```
ypise, line 137: Why not self._get_node_host_map()?

ypise (changes requested):

api/python/provisioner/commands/generate_roster.py
mazinamdar: Because of codacy issue I have kept that methods as static otherwise codacy will raise issue for this.

can we ignore such a kind of codacy issue ?

ypise: Yes, raise a ticket for such codacy issues with RE.
We can ignore them if they are genuine features.

mazinamdar: comment is addressed

api/python/provisioner/commands/generate_roster.py
mazinamdar: comment is addressed

api/python/provisioner/commands/generate_roster.py
mazinamdar: comment is addressed

ypise (approved):

PR #917, opened 2021-02-22
EOS-18303: Failhard: Stop immediately on component config failure

ypise:
------------------------------------------------
Failhard: the minions to stop immediately
after the first failure detected

### DIFF
--- a/srv/components/csm/config.sls
+++ b/srv/components/csm/config.sls
@@ -19,10 +19,12 @@
 Stage - Post Install CSM:
   cmd.run:
     - name: __slot__:salt:setup_conf.conf_cmd('/opt/seagate/cortx/csm/conf/setup.yaml', 'csm:post_install')
+    - failhard: True
 
 Stage - Config CSM:
   cmd.run:
     - name: __slot__:salt:setup_conf.conf_cmd('/opt/seagate/cortx/csm/conf/setup.yaml', 'csm:config')
+    - failhard: True
     - require:
       - Stage - Post Install CSM
 
@@ -69,5 +71,6 @@ Add {{ pillar['system']['service-user']['name'] }} user to prvsnrusers group:
 Stage - Init CSM:
   cmd.run:
     - name: __slot__:salt:setup_conf.conf_cmd('/opt/seagate/cortx/csm/conf/setup.yaml', 'csm:init')
+    - failhard: True
     - require:
       - Add {{ pillar['system']['service-user']['name'] }} user to certs group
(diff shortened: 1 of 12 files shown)


nitin-seagate (approved):
Looks good to me...

andkononykhin2 (commented):
@ypise we haven't work in that manner and I am not sure how clear a failure would be: I guess it would be harder to understand what actually failed since salt will mark as failed not only a failure but also not yet checked/run states.

I believe salt [doesn't recommend](https://docs.saltproject.io/en/latest/ref/states/failhard.html#global-failhard) that mode by similar reason.

ypise:
> @ypise we haven't work in that manner and I am not sure how clear a failure would be: I guess it would be harder to understand what actually failed since salt will mark as failed not only a failure but also not yet checked/run states.
> 
> I believe salt [doesn't recommend](https://docs.saltproject.io/en/latest/ref/states/failhard.html#global-failhard) that mode by similar reason.

I agree, which is why we had not enabled it.
However, for mini-provisioner they want the flow to break immediately.
May be we can set it back to `False` in future once mini-provisioner has stabilized for all components.

@andkononykhin2 what's your opinion?

83bhp (approved):

andkononykhin2:
> > @ypise we haven't work in that manner and I am not sure how clear a failure would be: I guess it would be harder to understand what actually failed since salt will mark as failed not only a failure but also not yet checked/run states.
> > I believe salt [doesn't recommend](https://docs.saltproject.io/en/latest/ref/states/failhard.html#global-failhard) that mode by similar reason.
> 
> I agree, which is why we had not enabled it.
> However, for mini-provisioner they want the flow to break immediately.
> May be we can set it back to `False` in future once mini-provisioner has stabilized for all components.
> 
> @andkononykhin2 what's your opinion?

I think [state level option](https://docs.saltproject.io/en/latest/ref/states/failhard.html#state-level-failhard) might be a good alternative. For sure it would require some more changes and copy-pastes. Possibly we can consider to use Jinja macros to make that easier to do in future.

andkononykhin2:
[our tip for Jinja macros](https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/docs/saltstack.md#jinja-macros---sharing-the-logic-in-sls)

ypise:
> > > @ypise we haven't work in that manner and I am not sure how clear a failure would be: I guess it would be harder to understand what actually failed since salt will mark as failed not only a failure but also not yet checked/run states.
> > > I believe salt [doesn't recommend](https://docs.saltproject.io/en/latest/ref/states/failhard.html#global-failhard) that mode by similar reason.
> > 
> > 
> > I agree, which is why we had not enabled it.
> > However, for mini-provisioner they want the flow to break immediately.
> > May be we can set it back to `False` in future once mini-provisioner has stabilized for all components.
> > @andkononykhin2 what's your opinion?
> 
> I think [state level option](https://docs.saltproject.io/en/latest/ref/states/failhard.html#state-level-failhard) might be a good alternative. For sure it would require some more changes and copy-pastes. Possibly we can consider to use Jinja macros to make that easier to do in future.

Done. This seems to be a better option.
We soon would need to implement the macro based deployments. Problem is we might lost granular control over the deployments over CLI.

PR #915, opened 2021-02-22
Update .whitesource

mukul-seagate11:

PR #920, opened 2021-02-23
Update .whitesource

mukul-seagate11:


ypise:
@mukul-seagate11 #915 caused a merge conflict. Which one do we want to maintain?

mukul-seagate11:
> @mukul-seagate11 #915 caused a merge conflict. Which one do we want to maintain?

I resolved the conflict so please merge #915 as the file require an update

PR #925, opened 2021-02-24
[EOS-18351] Refresh machine id of the system

mazinamdar:

1. delete old machine-id
2. create new machine-id
3. sync grains data

### DIFF
--- a/docs/saltstack.md
+++ b/docs/saltstack.md
@@ -53,6 +53,7 @@ Example:
 ```
 
 ### Issue: Grains are not getting refreshed within the same salt state
+Reference: https://github.com/saltstack/salt/issues/55440
 Reason: To apply the state, Salt needs to convert it from YAML format to a data structure. Since it is impossible to parse a YAML file that has Jinja statements inside, Salt needs to parse Jinja first.
 
 In other words, the pipeline looks like this
@@ -65,6 +66,6 @@ In other words, the pipeline looks like this
 Solution:
 
 ```
-1. Access the grain in a second state
-2. use slots
+1. split the logic into two SLS files and resolve the grains in the second one
+2. use slots which are evaluated on step 3
 ```
--- a/docs/saltstack.md
+++ b/docs/saltstack.md
@@ -51,3 +51,20 @@ Example:
 
 {% endmacro %}
 ```
+
+### Issue: Grains are not getting refreshed within the same salt state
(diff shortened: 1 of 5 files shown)


cortx-admin:

srv/components/provisioner/config/refresh_machine_id.sls
```diff
@@ -0,0 +1,46 @@
...
+  file.absent:
+    - name: /etc/machine-id
+
+Refresh machine_id on {{ grains['id'] }}:
+  cmd.run:
+    - name: systemd-machine-id-setup
+
+Sync data:
+  module.run:
+    - saltutil.refresh_grains: []
```
andkononykhin2, line 31: I have strong concerns that refreshed grains would be visible in later stages

andkononykhin2 (commented):
please verify the logic carefully using `state.apply` (for both `salt` and `salt-call`)

srv/components/provisioner/config/refresh_machine_id.sls
mazinamdar: splitted sls into 2 files and added issue in our docs

docs/saltstack.md
```diff
@@ -51,3 +51,20 @@ Example:
 
 {% endmacro %}
 ```
+
+### Issue: Grains are not getting refreshed within the same salt state
+Reason: To apply the state, Salt needs to convert it from YAML format to a data structure. Since it is impossible to parse a YAML file that has Jinja statements inside, Salt needs to parse Jinja first.
```
andkononykhin2, line 56: @mazinamdar please put a reference here to an original salt issue where good discussion happened

docs/saltstack.md
```diff
@@ -51,3 +51,20 @@ Example:
...
+
+1. Parse Jinja
+2. Parse the resulting YAML
+3. Interpret the data structure
+4. Apply the state
+
+Solution:
+
+```
+1. Access the grain in a second state
```
andkononykhin2, line 68: might be better to mention that the logic should be split into to SLS files and for slots - they are executed on a stage 3, e.g.

>1. split the logic into two SLS files and resolve the grains in the second one
>2. use [slots](https://docs.saltproject.io/en/latest/topics/slots/index.html) which are evaluated on step 3

andkononykhin2 (commented):
please address minor comment

docs/saltstack.md
mazinamdar: comment is addressed

docs/saltstack.md
mazinamdar: comment is addressed

andkononykhin2 (approved):

PR #926, opened 2021-02-24
Test helpers for setup routine

andkononykhin2:
Summary:

- adds ability to setup testing env of multiple nodes using pytest testing scenario
- makes setup provisioner logic more flexible for `local` source

Example of usage:

- install docker
- run `pytest test/build_helpers -k test_build_setup_env -s --root-passwd root1 --nodes-num 3`
  - that will start (almost) clean docker containers and configure root passwords as specified
  - print ssh configuration for the prepared env into the output
  - then it would prompt for a key (once any key is pressed it would teardown the env)
  - possible options (you may check them using `pytest test/build_helpers --help`, `custom options` part)
    - `--docker-mount-dev` to mount `/dev` into containers (necessary if you need to mount ISO inside a container)
    - `--root-passwd <STR>` to set a root password for initial ssh access
    - `--nodes-num <INT>` number of nodes, currently it can be from 1 to 3
- in other terminal run `provisioner setup_provisioner --source local --logfile --logfile-filename ./setup.log ... srvnode-1:<IP1> srvnode-2:<IP2> srvnode-3:<IP3>`

### DIFF
--- a/api/python/provisioner/srv/salt/cortx_repos/install.sls
+++ b/api/python/provisioner/srv/salt/cortx_repos/install.sls
@@ -17,7 +17,9 @@
 
 # TODO TEST
 
-{% if "RedHat" in grains['os'] %}
+# Note. we use True here since currently we do not consider
+#       any usage of an alternative option
+{% if "RedHat" in grains['os'] or True %}
 
 setup_yum_salt_repo:
   file.managed:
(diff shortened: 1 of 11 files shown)


cortx-admin:

api/python/provisioner/srv/salt/cortx_repos/install.sls
```diff
@@ -17,7 +17,9 @@
 
 # TODO TEST
 
-{% if "RedHat" in grains['os'] %}
+# Note. we use True here since currently we do not consider
+#       any usage of an alternative option
+{% if "RedHat" in grains['os'] or True %}
```
ypise, line 22: I don't understand this. It would always result in `True`. How does that help?

ypise (approved):

api/python/provisioner/srv/salt/cortx_repos/install.sls
andkononykhin2: That part (RedHat) is used only for local verification cases. The alternative condition just doesn't make sense in our current flow. Generally we should re-write / remove that soon.

PR #935, opened 2021-02-26
[EOS-18405] Integrate refresh machine_id states with setup provisioner API

mazinamdar:
Signed-off-by: mazinamdar mazhar.inamdar@seagate.com

Grains data is not getting refreshed within sls files if we call init.sls for machine_id states.
so applied states separately.

### DIFF
--- a/api/python/provisioner/commands/setup_provisioner.py
+++ b/api/python/provisioner/commands/setup_provisioner.py
@@ -1804,6 +1804,19 @@ class SetupProvisioner(SetupCmdBase, CommandParserFillerMixin):
                 f"\"{len(run_args.nodes)}\""
             ), targets=run_args.primary.minion_id
         )
+
+        # Grains data is not getting refreshed within sls files
+        # if we call init.sls for machine_id states.
+        logger.info("Refresh machine id on the system")
+        for state in [
+            'components.provisioner.config.machine_id.refresh_machine_id',
+            'components.provisioner.config.machine_id.refresh_grains'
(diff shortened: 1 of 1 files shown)


api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1803,6 +1803,18 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
             ), targets=run_args.primary.minion_id
         )
 
+        # Grains data is not getting refreshed within sls files
+        # if we call init.sls for machine_id states.
+        logger.info("Refresh machine id on the system")
+        for state in [
+            'components.provisioner.config.machine_id.refresh_machine_id',
```
andkononykhin2, line 1810: @mazinamdar can't we just apply only `components.provisioner.config.machine_id`?

andkononykhin2 (commented):
@mazinamdar Looks good. Please check one question. Thanks

api/python/provisioner/commands/setup_provisioner.py
mazinamdar: I am facing same issue of grains refresh if I run `components.provisioner.config.machine_id`  so we need to run both states separately.

andkononykhin2: Got it. Thanks

andkononykhin2 (approved):

ypise (approved):

PR #940, opened 2021-02-26
Improve help section of python API

mazinamdar:

Tried to improve description from help for API.

### DIFF
--- a/api/python/provisioner/cli_parser.py
+++ b/api/python/provisioner/cli_parser.py
@@ -141,8 +141,13 @@ def parse_args(args=None, commands: Optional[List] = None):
 
         # TODO description and help strings
         for cmd_name, cmd in commands.items():
+
+            desc = getattr(cmd, '_description', None)
+            if not desc:
+                desc = f'{cmd_name} configuration'
+
             subparser = subparsers.add_parser(
-                cmd_name, description='{} configuration'.format(cmd_name),
+                cmd_name, description='{}'.format(desc),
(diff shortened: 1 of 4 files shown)


api/python/provisioner/cli_parser.py
```diff
@@ -141,8 +141,13 @@ def parse_args(args=None, commands: Optional[List] = None):
 
         # TODO description and help strings
         for cmd_name, cmd in commands.items():
+
+            desc = getattr(cmd, '_description', None)
```
andkononykhin2, line 145: you may optimize that and use default value instead of `None`

api/python/provisioner/commands/auto_deploy.py
```diff
@@ -50,6 +50,7 @@ class AutoDeploy(SetupCmdBase, CommandParserFillerMixin):
         RunArgsSetupProvisionerGeneric,
         deploy_dual.run_args_type
     ]
+    _description = 'API to deploy Cortx on Hardware'
```
andkononykhin2, line 53: I think underscore is not necessary here since we are going to access that variable outside the class/children

andkononykhin2 (changes requested):

api/python/provisioner/cli_parser.py
mazinamdar: comment is addressed

api/python/provisioner/commands/auto_deploy.py

andkononykhin2 (approved):

PR #949, opened 2021-03-02
EOS-18457: Filename encoding error in PIP environment

dmitrydb:
I added the setup of the `LC_ALL` variable during the pip installation of provisioner dependencies.

-----
[View rendered docs/testing.md](https://github.com/Seagate/cortx-prvsnr/blob/EOS-18457/docs/testing.md)

### DIFF
--- a/api/python/provisioner/srv/salt/provisioner/api/install.sls
+++ b/api/python/provisioner/srv/salt/provisioner/api/install.sls
@@ -15,6 +15,12 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
+set_env_vars:
+  environ.setenv:
+    - name: LC_ALL
+    - value: en_US.UTF-8
+    - update_minion: True
+
 api_installed:
 {% if salt['pillar.get']('api_distr') == 'pip' %}
   pip.installed:
@@ -22,6 +28,8 @@ api_installed:
     - bin_env: /usr/bin/pip3
     - upgrade: False  # to reuse already installed dependencies
                       # that may come from rpm repositories
+    - require:
+      - set_env_vars
 {% else %}
   pkg.installed:
     - pkgs:
--- a/api/python/provisioner/srv/salt/provisioner/api/install.sls
+++ b/api/python/provisioner/srv/salt/provisioner/api/install.sls
@@ -15,11 +15,14 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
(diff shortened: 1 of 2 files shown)


api/python/provisioner/srv/salt/provisioner/api/install.sls
```diff
@@ -15,13 +15,21 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
+set_env_vars:
```
andkononykhin2, line 18: in current version that state would be executed always regardless of the api_distr type. Could you please move it under if condition since the issue (I believe) would appear only for pip based setups.

docs/testing.md
```diff
@@ -16,6 +16,15 @@
 The base Docker installation methods for popular Linux distributions are listed here:
 	https://docs.docker.com/engine/install/
 
+To run Docker commands without **sudo**, add your user to **docker** group
```
andkononykhin2, line 19: I am not sure that it is valid for all host systems (name of the group might be different but it is just a guess). Please double-check in docker docs.

docs/testing.md
```diff
@@ -103,6 +112,12 @@ sudo docker run hello-world
   :warning: NOTE: For more details about Python virtual environment see
   https://docs.python.org/3.6/library/venv.html
 
+- Activate created Python environment
+
+  ```bash
+  source venv/bin/activate
+  ```
+
```
andkononykhin2, line 120: Could you  please also update the line 162: currently the maximum number of [nodes is 6](https://github.com/Seagate/cortx-prvsnr/blob/2ebd2bf82bc349a724ae39346f464f20a4c70009/test/conftest.py#L45)

andkononykhin2 (changes requested):
@dmitrydb Good workaround. Thanks. Please just address minor comments.

docs/testing.md
dmitrydb: Yes, sure. I updated the parameter description.

docs/testing.md
dmitrydb: Yes, it is valid. I added the link to official page as a reference for all post-installation procedures.

api/python/provisioner/srv/salt/provisioner/api/install.sls
dmitrydb: Done

andkononykhin2 (approved):

PR #950, opened 2021-03-02
[EOS-18333] Generating storage_set_ids for nodes

mazinamdar:

Each storage set will have 3 nodes
 storage-set-1 -> srvnode-1, srvnode-2, srvnode-3
 storage-set-2 -> srvnode-4, srvnode-5, srvnode-6

```        storage_set_id:
            storage-set-1
    srvnode-2:
        ----------
        rack_id:
            None
        site_id:
            None
        storage_set_id:
            storage-set-1
    srvnode-3:
        ----------
        rack_id:
            None
        site_id:
            None
        storage_set_id:
            storage-set-1
    storage_sets:
        ----------
        storage-set-1:
            - srvnode-1
            - srvnode-2
            - srvnode-3
```


srv/components/provisioner/files/cluster_sls.template
```diff
@@ -38,7 +40,7 @@ cluster:
   srvnode-{{ (num + 1) }}:
     rack_id:
     site_id:
-    storage_set_id: storage-set-1
+    storage_set_id: storage-set-{{  (num//3) + 1 }}
```
andkononykhin2, line 43: please add spaces around binary operator here as well

srv/components/provisioner/files/cluster_sls.template
```diff
@@ -20,12 +20,14 @@
 cluster:
   cluster_id: {{ grains["cluster_id"] }}
   storage_sets:
-    storage-set-1:
-      {% for num in range(num_of_nodes) %}
-      {% if 0 <= num <= 2 %}
+    {% for set_id in range(0, num_of_nodes, 3) %}
```
andkononykhin2, line 23: I guess it's not actually  `set_id` but `first_node_id -1`. Could you elaborate variable names more...
as an option
```python
num_of_sets = num_of_nodes // 3 + bool(num_of_nodes % 3)
# not sure would it work in Jinja or not
```
so we can then build ranges over sets instead of nodes

andkononykhin2 (changes requested):
@mazinamdar I suggest to review logic and variable routine to make the logic clearer.

srv/components/provisioner/files/cluster_sls.template
```diff
@@ -20,12 +20,14 @@
...
+    storage-set-{{  (set_id//3) + 1 }}:
+      {% for num in range(set_id, set_id+3) %}
+      {% if num < num_of_nodes %}
```
ypise, line 26: Is this check still necessary?

srv/components/provisioner/files/cluster_sls.template
ypise: Also remove the extra space before: `(num//3)`

ypise: I've updated the code in a new commit

srv/components/provisioner/files/cluster_sls.template
ypise: Renamed the variable

ypise (approved):

PR #955, opened 2021-03-04
EOS-18456: API to set clusterID to pillar file

divya-vijay:

### DIFF
--- a/api/python/provisioner/commands/set_cluster_id.py
+++ b/api/python/provisioner/commands/set_cluster_id.py
@@ -1,0 +1,109 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com."
+#
+
+# Python API to set unique ClusterID to pillar file
+
+import logging
+from typing import Type
+
+from .. import (
+    inputs,
+    config
+)
+from ..config import LOCAL_MINION
+
+from ..paths import (
+    USER_SHARED_PILLAR
+)
+
+from ..utils import (
+    load_yaml,
+    dump_yaml
+)
+from . import (
+    CommandParserFillerMixin
+)
+from ..vendor import attr
+# from provisioner.api import grains_get
+
+logger = logging.getLogger(__name__)
+
+
+@attr.s(auto_attribs=True)
+class SetClusterId(CommandParserFillerMixin):
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+
+    def run(self, targets=LOCAL_MINION):
+        """set_cluster_id command execution method.
+
+        Checks for cluster id in common cluster file.
+        If not present, generates an id and sets.
+
+        Execution:
+        `provisioner set_cluster_id`
+        Takes no mandatory argument as input.
+
+        """
+        setup_location = config.profile_base_dir().parent
+        for setup_file_name in setup_location.iterdir():
+            setup_name = str(setup_file_name)
+            paths = config.profile_paths(
+                    config.profile_base_dir(
+                    setup_location, setup_name))
+
+            all_minions_dir = (
+                paths['salt_fileroot_dir'] / "provisioner/files/minions/all"
+            )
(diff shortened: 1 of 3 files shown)


api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1868,6 +1868,13 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
                 "salt-call state.apply components.misc_pkgs.ipmi"
             )
 
+        logger.info("Setting unique ClusterID to pillar file")
```
ypise, line 1871: Since, we are not sure of the factory and field flow yet, we would maintain this as an explicit command.
Comment out these lines for now.

api/python/provisioner/commands/set_cluster_id.py
```diff
@@ -0,0 +1,109 @@
...
+
+        """
+        setup_location = config.profile_base_dir().parent
+        for setup_file_name in setup_location.iterdir():
+            setup_name = str(setup_file_name)
+            paths = config.profile_paths(
+                    config.profile_base_dir(
+                    setup_location, setup_name))
+
+            all_minions_dir = (
```
ypise, line 68: What path is this?

api/python/provisioner/commands/set_cluster_id.py
```diff
@@ -0,0 +1,109 @@
...
+                            "proceeding to set ClusterID to pillar file")
+                try:
+                    cluster_uuid = load_yaml(cluster_id_path)['cluster_id']
+
+                    cluster_pillar_file = (
+                                     f"{USER_SHARED_PILLAR._all_hosts_dir}/"
+                                     f"{USER_SHARED_PILLAR._prefix}"
+                                     "cluster.sls"
+                                     )
+                    cluster_content = load_yaml(cluster_pillar_file)
```
ypise, line 84: We should interact with pillar only through Provisioner API. Do not explicitly work with any yaml or sls files.

api/python/provisioner/commands/set_cluster_id.py
```diff
@@ -0,0 +1,109 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com."
+#
+
+# Python API to set unique ClusterID to pillar file
```
ypise, line 18: We also need the 4 steps to set cluster_id here involving check for srvnode-*.

api/python/provisioner/commands/set_cluster_id.py
```diff
@@ -0,0 +1,109 @@
...
+        for setup_file_name in setup_location.iterdir():
+            setup_name = str(setup_file_name)
+            paths = config.profile_paths(
+                    config.profile_base_dir(
+                    setup_location, setup_name))
+
+            all_minions_dir = (
+                paths['salt_fileroot_dir'] / "provisioner/files/minions/all"
+            )
+            cluster_id_path = all_minions_dir / 'cluster_id'
```
ypise, line 71: cluster_is should be checked only inside cluster pillar and nowhere else. We need to maintain that discipline.

ypise (changes requested):

api/python/provisioner/commands/set_cluster_id.py
ypise: 1. Check if cluster-id is set in grains using grains_get API
2. Check if cluster-id is set in pillar using pillar_get API: cluster/cluster_id
3. If cluster-id is not found in #1 and #2, generate a new UUID
4. Set the cluster_id in cluster pillar using pillar_set API: cluster/cluster_id
5. Make an entry of the set cluster_id to each srvnode entry in cluster pillar

ypise (changes requested):

api/python/provisioner/commands/setup_provisioner.py
divya-vijay: @ypise for now, added it to bootstrap flow as we're checking node role for `set_cluster_id` and executing. Thanks

api/python/provisioner/commands/set_cluster_id.py
divya-vijay: @ypise addressed the sequence of steps. 
although, one minor clarification needed here. If cluster_id is not found steps 2 and 3, should we not break the code flow saying "bootstrap is not done, hence cluster_id cannot be set" ? Since we're generating cluster_uuid in bootstrap process also..

api/python/provisioner/commands/set_cluster_id.py
divya-vijay: Done. Removed all file handling from the api since thats prone to changes.

api/python/provisioner/commands/set_cluster_id.py
divya-vijay: Done. Removed all file handling from the api since thats prone to changes. Thanks

api/python/provisioner/commands/set_cluster_id.py
divya-vijay: Removed file handling and paths.

PR #965, opened 2021-03-05
EOS-16932: Adjust SW Upgrade repo setup to use the similar logic as for single repository installation

dmitrydb:
I adjusted SW Upgrade repo setup to use the similar logic as for single repository installation


api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -121,11 +188,17 @@ def dynamic_validation(self, params: inputs.SWUpgradeRepo, targets: str):  # noq
 
         try:
             logger.debug("Configuring upgrade candidate repo for validation")
-            self._prepare_repo_for_apply(candidate_repo, enabled=False)
 
-            self._apply(candidate_repo, targets)
+            self._prepare_single_iso_for_apply(candidate_repo)
```
ypise, line 192: @dmitrydb Could you add some logger statements against these calls to inform what we are attempting?

api/python/provisioner/config.py
```diff
@@ -502,28 +501,12 @@ class ReleaseInfo(Enum):
...
-    """List of supported SW Upgrade Repositories."""
-
-    OS = "os"  # yum repo
-    third_party = "3rd_party"  # yum repo
-    cortx = "cortx_iso"  # yum repo
-    python = "python_deps"  # python index
+class InstallType(Enum):
+    """List of installation types."""
 
+    base = "base"
```
ypise, line 507: If these are to be treated as ENUM, please define the variables in CAPS.

pillar/components/release.sls
```diff
@@ -18,6 +18,7 @@
 release:
     product: LR2
     setup: cortx
+    install_type: base
```
ypise, line 21: Please document what base means and other possible types.

test/r2.txt
```diff
@@ -122,7 +122,7 @@ yum install /root/rpmbuild/RPMS/x86_64/cortx-prvsnr-2.0.0-0_git123456_el7.x86_64
 Step-4: Validate that `repos` states are available and visible for salt commands
 
 ```bash
-salt-call state.id_exists "configure pip" repos.pip_config
+salt-call '*' state.id_exists "configure pip" repos.pip_config
```
ypise, line 125: `salt-call '*' ` won't work :)
Do update this to `salt '*'`

api/python/provisioner/inputs.py
```diff
@@ -877,3 +879,58 @@ def is_iso(self):
...
+                    'source': (f'file://{self.target_build}/{self.release}/'
+                               f'{config.CORTX_ISO_DIR}'),
+                    'is_repo': True
+                },
+                f'{config.CORTX_3RD_PARTY_ISO_DIR}_{self.release}': {
+                    'source': (f'file://{self.target_build}/{self.release}/'
+                               f'{config.CORTX_3RD_PARTY_ISO_DIR}'),
+                    'is_repo': True
+                },
+                f'{config.CORTX_PYTHON_ISO_DIR}_{self.release}': {
```
ypise, line 931: Are we considering possibility of Python to be shipped in separate ISO?
I am of opinion that for now we should stick to CORTX_3RD_PARTY_ISO_DIR

ypise (changes requested):

api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -15,15 +15,15 @@
...
+from provisioner.pillar import PillarKey, PillarResolver
+
+from ..salt import copy_to_file_roots, local_minion_id
 from .set_swupdate_repo import SetSWUpdateRepo
 from .. import inputs, values
-from ..config import (REPO_CANDIDATE_NAME, SW_UPGRADE_REPOS, YUM_REPO_TYPE,
-                      PRVSNR_USER_FILES_SWUPGRADE_REPOS_DIR, RELEASE_INFO_FILE,
-                      ReleaseInfo
-                      )
+from ..config import (REPO_CANDIDATE_NAME, IS_REPO_KEY, RELEASE_INFO_FILE,
```
andkononykhin2, line 25: please wrap that into multiple lines so it can be simpler to understand changes in future

api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -32,8 +32,74 @@
 
 class SetSWUpgradeRepo(SetSWUpdateRepo):
 
-    _REPO_DEST = PRVSNR_USER_FILES_SWUPGRADE_REPOS_DIR
     _BASE_DIR_PILLAR = "release/upgrade/base_dir"
+    _INSTALL_TYPE_PILLAR_KEY = "release/install_type"
+
+    def _configure_install_type(self, install_type: str) -> PillarKey:
```
andkononykhin2, line 38: could double-check return value type declaration and what is actually should be returned?

api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -32,8 +32,74 @@
...
+        Returns
+        -------
+        str
+            return `release/install_type` value for backup
+
+        """
+        pillar_path = PillarKey(self._INSTALL_TYPE_PILLAR_KEY)
+        pillar = PillarResolver(local_minion_id()).get([pillar_path])
+
+        for target, _pillar in pillar.items():
```
andkononykhin2, line 71: we expect only one target (local) here, looping looks as an overhead

api/python/provisioner/srv/salt/repos/init.sls
```diff
@@ -22,18 +22,21 @@
 {% from './install.sls' import repo_added with context %}
 {% from './teardown.sls' import repo_absent with context %}
 
-{% for release, source in pillar['release']['base']['repos'].items() %}
+{% set default_install_type = 'base' %}
+{% set install_type = salt['pillar.get']('release:install_type', default_install_type) %}
```
andkononykhin2, line 26: it means that pillar key should be set beforehand each time we call that state. E.g. the current version of SWUpgrade command doesn't do that so it will fail if that value is different than `upgrade`.

I know we discussed another option - inline pillar: it's better a bit but not a good UX as well.

I believe the best other alternative is to use different states for different routine, e.g.:
- `repos.init` for base repositories (as part of bootstrap)
- `repos.upgrade` for cases of upgrade bundles
To avoid copy paste please use [macros](https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/docs/saltstack.md#jinja-macros---sharing-the-logic-in-sls) , so that install type would be just an argument to a macros call.

api/python/provisioner/inputs.py
dmitrydb: I am following here the scheme which we used before. As far as I know, we have a few tickets about python index configuration for SW Upgrade

test/r2.txt
dmitrydb: Yes, I forgot to update the salt command. I fixed that.

pillar/components/release.sls
dmitrydb: Yes, I will document this and other API details today during the day. Maybe in separate PR

api/python/provisioner/config.py
dmitrydb: Thank you! I updated that

api/python/provisioner/commands/set_swupgrade_repo.py
dmitrydb: Added logging string to `self._prepare_single_iso_for_apply` and before the first `self._apply`

api/python/provisioner/inputs.py
```diff
@@ -877,3 +879,58 @@ def is_iso(self):
...
+            METADATA_ARGPARSER: {
+                'help': 'file path to base directory for SW upgrade'
+            }
+        }
+    )
+
+    @property
+    def pillar_key(self):
+        # the local root pillar key for 'swupgrade' installation type
+        return "repos"
```
andkononykhin2, line 894: It means each new upgrade bundle will replace all current ones. Please consider the different scheme:
```yaml
upgrade
  repos:
    sw_upgrade_<RELEASE>:
    sw_upgrade_<RELEASE>_OS:
    ...
```

Actually the more beter options would be to group them per release. But it would require to review our states deeper.

andkononykhin2 (changes requested):
@dmitrydb Thanks. I checked and believe two major issues are there:
- the support of multiple sw upgrade bundles (e.g. 2.1, 2.2)
- UX in scope of switching between base and upgrade repo bundles setup in states
Please check my comments.

pillar/components/release.sls
ypise: It would be better to add a small comment here itself

ypise (approved):

api/python/provisioner/inputs.py
dmitrydb: Yes, I have already used this scheme in the first version of this PR but without "sw_upgrade_" prefix. In current version of API I group repositories by its version. And added "sw_upgrade_" prefix

api/python/provisioner/srv/salt/repos/init.sls
dmitrydb: Code in `set_swupgrade_repo.py` guarantees that `install_type` pillar will be kept updated. But I implemented the new approach and uses `repos.upgrade` to invoke upgrade logic.

api/python/provisioner/commands/set_swupgrade_repo.py
dmitrydb: This code is no more used for SW Upgrade. Deleted this method

api/python/provisioner/commands/set_swupgrade_repo.py
dmitrydb: This code is no more needed for the newer approach. Deleted this method

api/python/provisioner/commands/set_swupgrade_repo.py
dmitrydb: In my opinion, that approach wastes more space. But I fixed that.

pillar/components/release.sls
dmitrydb: It is no more needed due to this parameter was removed in the new version of this PR.

ypise (approved):

api/python/provisioner/inputs.py
```diff
@@ -877,3 +879,71 @@ def is_iso(self):
 @attr.s(auto_attribs=True)
 class SWUpgradeRepo(SWUpdateRepo):
     _param_di = param_spec['swupgrade/repo']
+    target_build: str = attr.ib(
```
andkononykhin, line 882: who is expected to provide that value? the meaning of it is actually the base directory for upgrade bundle so we can evaluate all repos locations. So it is expected to be an internal logic:
- if source is ISO then it would be a mount point on the target systems
- if source is DIR then it would be a location the directory  on target systems
- if source is an URL then it would the base dir would match the URL

api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -121,11 +150,19 @@ def dynamic_validation(self, params: inputs.SWUpgradeRepo, targets: str):  # noq
 
         try:
             logger.debug("Configuring upgrade candidate repo for validation")
-            self._prepare_repo_for_apply(candidate_repo, enabled=False)
 
-            self._apply(candidate_repo, targets)
+            self._prepare_single_iso_for_apply(candidate_repo)
+
+            base_dir = self._get_mount_dir()
+            candidate_repo.target_build = base_dir
```
andkononykhin, line 157: that's not true for urls

andkononykhin (changes requested):
@dmitrydb I have few concerns for the current version. Please check the comments. Thanks

api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -32,9 +34,35 @@
...
-    _REPO_DEST = PRVSNR_USER_FILES_SWUPGRADE_REPOS_DIR
     _BASE_DIR_PILLAR = "release/upgrade/base_dir"
 
+    @staticmethod
+    def _prepare_single_iso_for_apply(repo: inputs.SWUpgradeRepo):
+        """Prepare repository for apply.
+
+        Parameters
+        ----------
+        repo : inputs.SWUpgrade
```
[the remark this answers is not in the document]
dmitrydb: Done

api/python/provisioner/srv/salt/repos/upgrade.sls
```diff
@@ -15,14 +15,17 @@
...
-  mount.mounted:
-    - name: {{ mount_dir }}
-    - device: {{ source }}
-    - mkmnt: True
-    - fstype: iso9660
-    - persist: True
+{% set base_dir = pillar['release']['upgrade']['base_dir'] %}
 
-{% endmacro %}
+{% for version, repos in pillar['release']['upgrade']['repos'].items() %}
```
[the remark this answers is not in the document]
dmitrydb: Convert this in `upgrade.sls`

api/python/provisioner/inputs.py
```diff
@@ -877,3 +879,71 @@ def is_iso(self):
...
+    @property
+    def pillar_key(self):
+        # the local root pillar key for 'swupgrade' installation type
+        return self.release
+
+    @property
+    def pillar_value(self):
+        prefix = "sw_upgrade"
+
+        if self.is_special() or self.is_remote():
```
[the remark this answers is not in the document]
dmitrydb: Make `if` and `elif` here

api/python/provisioner/inputs.py
dmitrydb: It is internal logic to setup the `target_build` variable

api/python/provisioner/commands/set_swupgrade_repo.py
dmitrydb: Will improve

api/python/provisioner/inputs.py
```diff
@@ -877,3 +879,60 @@ def is_iso(self):
 @attr.s(auto_attribs=True)
 class SWUpgradeRepo(SWUpdateRepo):
     _param_di = param_spec['swupgrade/repo']
+    # file path to base directory for SW upgrade
+    _target_build: str = attr.ib(default=None)
```
andkononykhin, line 883: looks good, JFYI: you may shorten that more as:
```
_target_build: str = None
_enabled: bool = False
```
due to `@attr.s(auto_attribs=True)` [attr lib](https://www.attrs.org/en/stable/api.html) will get that right as well (search for `auto_attribs`)

api/python/provisioner/inputs.py
```diff
@@ -877,3 +879,60 @@ def is_iso(self):
...
+                for repo in (config.OS_ISO_DIR,
+                             config.CORTX_ISO_DIR,
+                             config.CORTX_3RD_PARTY_ISO_DIR,
+                             config.CORTX_PYTHON_ISO_DIR)
+            }
+            res[self.release] = None
+
+            return res
+        elif self.is_remote():
+            return self.source
```
andkononykhin, line 906: in case of remote it would become a `target_build` and we need to setup few repos from that as well (cortx, 3rd parties, os, ...). You may treat that like ISO mounted on the host server.

Please create a follow up task for remote urls support.

api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -155,24 +192,51 @@ def dynamic_validation(self, params: inputs.SWUpgradeRepo, targets: str):  # noq
...
+
+    def _run(self, params: inputs.SWUpgradeRepo, targets: str):
+        repo = params
+
+        base_dir = self._get_mount_dir()
+        repo.target_build = base_dir
+        repo.enabled = True
+
+        # TODO remove that block once that check fails the dynamic validation
+        if self._is_repo_enabled(f'sw_update_{repo.release}'):
```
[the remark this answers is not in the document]
andkononykhin2: ok, we can keep that for now.

api/python/provisioner/inputs.py
```diff
@@ -877,3 +879,71 @@ def is_iso(self):
...
+                             config.CORTX_PYTHON_ISO_DIR)
+            }
+            res[self.release] = None
+
+            return res
+        else:
+            # source = 'iso' if self.source.is_file() else 'dir'
+            return {
+                f'{self.release}': {
+                    'source': f'salt://{self.release}.iso',
```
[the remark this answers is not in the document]
andkononykhin2: thanks

api/python/provisioner/inputs.py
andkononykhin2: looks good now, thanks

api/python/provisioner/srv/salt/repos/upgrade.sls
```diff
@@ -15,14 +15,23 @@
...
+
+    {% set single_iso_repo = {version: repos.pop(version)} %}
+
+    # to mount the single ISO first
+    {{ setup_repos(single_iso_repo, base_dir) }}
+
+    {% set upgrade_repos = dict() %}
+    {% for release in repos %}
+    {% set key =  'sw_upgrade_' ~ release ~ '_' ~ version %}
```
andkononykhin, line 31: minor thing: indentation of loop's inner-logic will make that easier to understand

andkononykhin (approved):
@dmitrydb  Looks good now. Thank you. Could you just create and refer some follow-up tasks to support urls better?

andkononykhin (approved):

andkononykhin2 (approved):

PR #966, opened 2021-03-05
[EOS-18497] cleanup states for glusterfs

mazinamdar:


srv/components/system/storage/glusterfs/teardown/volume_remove.sls
```diff
@@ -23,13 +23,20 @@ glusterfs_volume_dir_{{ volume['mount_dir'] }}_unmount:
     - name: {{ volume['mount_dir'] }}
     - persist: True
 
+# remove mount dir from fstab
```
andkononykhin2, line 26: I remember we discussed that but it looks like as a shoot in the temple ) I believe the above state should take care about that.

andkononykhin2 (approved):

ypise (approved):

PR #971, opened 2021-03-06
EOS-16880: Extend NTP configuration to controller in API.

83bhp:

### DIFF
--- a/api/python/provisioner/api_spec.yaml
+++ b/api/python/provisioner/api_spec.yaml
@@ -47,7 +47,7 @@ pillar_set:
   type: PillarSet
   input_type: PillarInputBase
 set_ntp:
-  type: Set
+  type: SetNTP
   input_type: NTP
   states:
     post:
--- a/api/python/provisioner/commands/__init__.py
+++ b/api/python/provisioner/commands/__init__.py
@@ -351,6 +351,139 @@ class Get(CommandParserFillerMixin):
         return res
 
 
+@attr.s(auto_attribs=True)
+class SetNTP(CommandParserFillerMixin):
+    # TODO at least either pre or post should be defined
+    input_type: Type[
+        Union[inputs.ParamGroupInputBase, inputs.ParamDictItemInputBase]
+    ]
+    pre_states: List[State] = attr.Factory(list)
+    post_states: List[State] = attr.Factory(list)
+
+    _run_args_type = RunArgsUpdate
+
+    # TODO input class type
+    @classmethod
+    def from_spec(cls, input_type: str, states: Dict):
+        return cls(
+            input_type=getattr(inputs, input_type),
+            pre_states=[State(state) for state in states.get('pre', [])],
+            post_states=[State(state) for state in states.get('post', [])]
+        )
+
+    def _apply(self, params, targets):
+        pillar_updater = PillarUpdater(targets)
+
+        pillar_updater.update(params)
+        try:
+            logger.debug('Applying pre states')
+            StatesApplier.apply(self.pre_states)
+            try:
+                logger.debug('Applying pillar changes')
+                pillar_updater.apply()
+
+                logger.debug('Applying post states')
+                StatesApplier.apply(self.post_states)
+            except Exception:
+                logger.warning(
+                    "Starting Rollback: Result of "
+                    "failure to apply pillar changes."
+                )
+                # TODO more solid rollback
+                pillar_updater.rollback()
+                pillar_updater.apply()
+                raise
+        except Exception:
+            logger.exception('Failed to apply changes')
+            # treat post as restoration for pre, apply
+            # if rollback happened
+            StatesApplier.apply(self.post_states)
+            raise
+
+    def _set_ctrl_ntp(self, params, targets):
+        pillar_updater = PillarUpdater(targets)
+
+        #TODO:
+        #  1. ensure pillar data is updated at this stage
+        #  2. confirm what pillar the NTP data is stored in
+        #  3. test dry run
+        #  4. test from CSM UI.
+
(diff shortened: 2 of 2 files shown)


api/python/provisioner/api_spec.yaml
```diff
@@ -47,7 +47,7 @@ pillar_set:
   type: PillarSet
   input_type: PillarInputBase
 set_ntp:
-  type: Set
+  type: SetNTP
```
ypise, line 50: Should this be SetControllerNTP?

api/python/provisioner/commands/__init__.py
```diff
@@ -351,6 +351,139 @@ def run(self, *args, targets=ALL_MINIONS, **kwargs):
...
+            pre_states=[State(state) for state in states.get('pre', [])],
+            post_states=[State(state) for state in states.get('post', [])]
+        )
+
+    def _apply(self, params, targets):
+        pillar_updater = PillarUpdater(targets)
+
+        pillar_updater.update(params)
+        try:
+            logger.debug('Applying pre states')
```
ypise, line 379: logger.info('Applying pre states for controller NTP')

api/python/provisioner/commands/__init__.py
```diff
@@ -351,6 +351,139 @@ def run(self, *args, targets=ALL_MINIONS, **kwargs):
...
+            StatesApplier.apply(self.pre_states)
+            try:
+                logger.debug('Applying pillar changes')
+                pillar_updater.apply()
+
+                logger.debug('Applying post states')
```
ypise, line 385: logger.info('Applying post states for controller NTP')

api/python/provisioner/commands/__init__.py
```diff
@@ -351,6 +351,139 @@ def run(self, *args, targets=ALL_MINIONS, **kwargs):
...
+                StatesApplier.apply(self.post_states)
+            except Exception:
+                logger.warning(
```
ypise, line 388: logger.error("Failed to apply time server changes to pillar for controller."
                    "Starting Rollback: Result of "
                    "failure to apply pillar changes."
)

api/python/provisioner/commands/__init__.py
```diff
@@ -351,6 +351,139 @@ def run(self, *args, targets=ALL_MINIONS, **kwargs):
...
+                        passwd=pillar[passwd],
+                        ntp_server=pillar[ntp_server],
+                        timezone=pillar[timezone]
+                    )
+                )
+            )
+        )
+
+    def _run(self, params, targets):
+        self._apply(params, targets)
```
ypise, line 462: Why do we need to update pillar for Controller?
We should check the pillar values set for server NTP and reuse the same for controller.

ypise (changes requested):

api/python/provisioner/api_spec.yaml
83bhp: This API is not specific to the controller but for both nodes and the controller together.
So, whenever NTP is config command is run it will first set the NTP on nodes followed by controller.

api/python/provisioner/commands/__init__.py
83bhp: the states are for server node not for controller.

api/python/provisioner/commands/__init__.py
83bhp: same, it's for node not the controller.

api/python/provisioner/commands/__init__.py
83bhp: ok, will improvise the error/warning.

api/python/provisioner/commands/__init__.py
83bhp: as said it's combined for server and controller. so first server node will be configured which updates the pillar then controller will set with those pillar values.

api/python/provisioner/commands/__init__.py
```diff
@@ -351,6 +352,138 @@ def run(self, *args, targets=ALL_MINIONS, **kwargs):
...
+            )
+        except Exception:
+            logger.exception('Failed to configure NTP on enclosure')
+            raise
+
+    def _run(self, params, targets):
+        self._apply(params, targets)
+        # Configure NTP only if the setup type is physical
+        from .grains_get import GrainsGet
+        from ..salt import local_minion_id
```
SomwanshiAnjali, line 459: @83bhp  Can we add imports at the beginning and local_minion_id is already imported in this file.
https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/api/python/provisioner/commands/__init__.py#L74

api/python/provisioner/commands/__init__.py
```diff
@@ -351,6 +352,138 @@ def run(self, *args, targets=ALL_MINIONS, **kwargs):
         return res
 
 
+@attr.s(auto_attribs=True)
+class SetNTP(CommandParserFillerMixin):
```
andkononykhin2, line 356: `SetNTP` shouldn't copy-paste the whole `Set` class, please just inherit from the latter and redefine what is different

andkononykhin2 (changes requested):
@83bhp please use other class hierarchy model, it would drastically simplify the overall logic. Also I believe we need to move more logic into salt states, Thanks

api/python/provisioner/commands/__init__.py
83bhp: thanks for pointing, addressed!

83bhp:
I have addressed all the comments in the new PR: https://github.com/Seagate/cortx-prvsnr/pull/979
so this PR is obsolete, hence closing it.

PR #989, opened 2021-03-10
Improvements to HealthMap Schema

ypise:

### DIFF
--- a/srv/_modules/sspl.py
+++ b/srv/_modules/sspl.py
@@ -44,17 +44,27 @@ def merge_health_map_schema(source_json="/tmp/resource_health_view.json"):
   health_map_file = healthmap_schema_pillar['filename']
 
   data = local.cmd('*', 'file.read', [source_json])
-  node1_data = json.loads(data["srvnode-1"])
-  node2_data = json.loads(data["srvnode-2"])
+  node_data = []
+  for node in __pillar__['cluster'].keys():
+    if "srvnode-" in node:
+      node_data = json.loads(data[node])
 
-  node_1_rack_nodes = node1_data["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
-  node_2_rack_nodes = node2_data["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
-  node_1_rack_nodes.update(node_2_rack_nodes)
+  node_rack_nodes = None
+  for i, node_data in enumerate(node_rack_nodes):
+    if i == 0:
+      node_rack_nodes = (
+        node_data[i]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
+      )
+    else:
+      node_rack_nodes.update(
+        node_data[i]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
+      )
+
   local.cmd('*', 'file.mkdir',[health_map_path])
   local.cmd(
     '*',
     'file.write',
-    [os.path.join(health_map_path, health_map_file), json.dumps(node1_data, indent=4)]
+    [os.path.join(health_map_path, health_map_file), json.dumps(node_rack_nodes, indent=4)]
   )
 
   # TODO use salt formulas, CSM suggested to have 777 permission just for hotfix
--- a/srv/_modules/sspl.py
+++ b/srv/_modules/sspl.py
@@ -44,20 +44,20 @@ def merge_health_map_schema(source_json="/tmp/resource_health_view.json"):
   health_map_file = healthmap_schema_pillar['filename']
 
   data = local.cmd('*', 'file.read', [source_json])
-  node_data = []
+  node_data = dict()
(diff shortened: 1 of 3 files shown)


srv/_modules/sspl.py
```diff
@@ -44,17 +44,27 @@ def merge_health_map_schema(source_json="/tmp/resource_health_view.json"):
...
+  node_rack_nodes = None
+  for i, node_data in enumerate(node_rack_nodes):
+    if i == 0:
+      node_rack_nodes = (
+        node_data[i]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
+      )
+    else:
+      node_rack_nodes.update(
+        node_data[i]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
+      )
```
dmitrydb, line 61: In this context it is safe to use `node_data[0]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]` as initial directory for `node_rack_nodes`. But such approach modifies in parallel the original object to which refers variable `node_data`

srv/_modules/sspl.py
```diff
@@ -44,17 +44,27 @@ def merge_health_map_schema(source_json="/tmp/resource_health_view.json"):
...
   data = local.cmd('*', 'file.read', [source_json])
-  node1_data = json.loads(data["srvnode-1"])
-  node2_data = json.loads(data["srvnode-2"])
+  node_data = []
+  for node in __pillar__['cluster'].keys():
+    if "srvnode-" in node:
+      node_data = json.loads(data[node])
+
+  node_rack_nodes = None
+  for i, node_data in enumerate(node_rack_nodes):
```
dmitrydb, line 53: I think this code will not work and will raise an Exception. I can guess that code supposed to be:
```python
for i, _data in enumerate(node_data):
    if i == 0:
      node_rack_nodes = (
        _data["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
      )
    else:
      node_rack_nodes.update(
        _data["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
      )
```

dmitrydb: Also it can be simplified like this
```python
node_rack_nodes = node_data[0]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
for data in node_data:
   node_rack_nodes.update(data[i]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"])
````

**Review by dmitrydb — CHANGES_REQUESTED** [review]
Hi Yash, I posted few comments here

File: srv/_modules/sspl.py (hunk 1 above)
Comment (ypise) [reply srv/_modules/sspl.py]: @dmitrydb, I am not sure I understand, how original object is modified here.
Could you elaborate?

File: srv/_modules/sspl.py (hunk 2 above)
Comment (ypise) [reply srv/_modules/sspl.py]: `node_rack_nodes = node_data[0]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]`
Yes, initially thought about the same, but a null  value for _data would fail. A for loop in a way covers for the null value check.

Comment (ypise) [reply srv/_modules/sspl.py]: > I think this code will not work and will raise an Exception. I can guess that code supposed to be:
> 
> ```python
> for i, _data in enumerate(node_data):
>     if i == 0:
>       node_rack_nodes = (
>         _data["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
>       )
>     else:
>       node_rack_nodes.update(
>         _data["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
>       )
> ```

This was a buggy code and would have never worked. Thanks for pointing it out.
I've corrected it.

File: srv/_modules/sspl.py (hunk 1 above)
Comment (dmitrydb) [reply srv/_modules/sspl.py]: It's because this assignment
```python
    if i == 0:
      node_rack_nodes = (
        node_data[i]["cluster"]["sites"]["001"]["rack"]["001"]["nodes"]
      )
```
copies reference to the dict rather than the data of this dict.
It is better to illustrate this by example:

```python
>>> a = [{"nodes": {"node_1": 1, "node_2": 2} }, {"nodes": {"nodes_3": 3, "nodes_4": 4}}]
>>> for i, _ in enumerate(a):
...     if i == 0:
...             nodes_rack = a[i]["nodes"]
...     else:
...             nodes_rack.update(a[i]["nodes"])
... 
>>> a
[{'nodes': {'node_1': 1, 'node_2': 2, 'nodes_3': 3, 'nodes_4': 4}}, {'nodes': {'nodes_3': 3, 'nodes_4': 4}}]
>>> nodes_rack
{'node_1': 1, 'node_2': 2, 'nodes_3': 3, 'nodes_4': 4}
>>> id(nodes_rack)
139647028594752
>>> id(a[0]["nodes"])
139647028594752
>>> id(a[0]["nodes"]) == id(nodes_rack)
True
```

dmitrydb (approved):

PR #996, opened 2021-03-12
fixes possible failure case in test helper

andkononykhin2:


test/helper.py
```diff
@@ -52,7 +52,7 @@
 localhost = testinfra.get_host('local://')
 
 BUILD_BUNDLE_SCRIPT = (
-    PROJECT_PATH
+    (PROJECT_PATH if PROJECT_PATH else PRVSNR_ROOT_DIR)
```
dmitrydb, line 55: I think we can simplify this to that shorter form
```python
BUILD_BUNDLE_SCRIPT = PROJECT_PATH or PRVSNR_ROOT_DIR
```

dmitrydb (changes requested):
I added one small comment

test/helper.py
andkononykhin2: updated

dmitrydb (approved):

PR #1002, opened 2021-03-14
Update README.md

pujamudaliar:
Signed-off-by: Puja M puja.mudaliar@seagate.com

Update readme file with contributors badge.

-----
[View rendered README.md](https://github.com/pujamudaliar/cortx-prvsnr/blob/pre-cortx-1.0/README.md)


ypise:
@pujamudaliar could you signoff this commit for DCO check?
Please don't bypass as this would cause trouble with main and stable branches later as well.

pujamudaliar:
> @pujamudaliar could you signoff this commit for DCO check?
> Please don't bypass as this would cause trouble with main and stable branches later as well.
@ypise : I have passed DCO.

PR #1011, opened 2021-03-16
Update auto_assign.yml

mukul-seagate11:
Request to update gatekeepers as per requested


ypise (changes requested):
The commit has to be DCO signed.
Do not bypass on browser.

PR #1017, opened 2021-03-17
fixes recent regression in cli parser configuration

andkononykhin2:


api/python/provisioner/inputs.py
```diff
@@ -219,9 +219,10 @@ def prepare_args(
                             }
                         )
                         args = attr_parser_cls(attr_copy, prefix=parser_prefix)
+                        res[args.name] = args.kwargs
                 else:
                     args = attr_parser_cls(_attr, prefix=parser_prefix)
-                res[args.name] = args.kwargs
```
ypise, line 224: Instead of duplicating the assignment, define `args=None` before `if` and maintain the same code.
Then, `res[args.name] = args.kwargs if args else None`

or, inside both if and else
```
res[args.name] = attr_parser_cls(_attr, prefix=parser_prefix).kwargs
```

ypise (changes requested):

api/python/provisioner/inputs.py
ypise: I've made the change as suggested.

ypise (approved):

PR #1018, opened 2021-03-17
EOS-18738: Unskip non-unit test cases

rahuljyoti65:

### DIFF
--- a/api/python/setup.py
+++ b/api/python/setup.py
@@ -40,6 +40,8 @@ tests_require = [
     'pytest-xdist==1.29.0',
     'pytest-timeout==1.3.4',
     'pytest-mock==3.1.0',
+    'coverage==5.3.1',
+    'pytest-cov==2.11.0',
 ]
 
 codacy_require = [
--- a/cli/src/code_coverage/prvsnr-coverage.py
+++ b/cli/src/code_coverage/prvsnr-coverage.py
@@ -32,17 +32,14 @@ def install_coverage():
         "yum install -y gcc cpp python3-devel"
     )
     _run(
-        "pip3 install -U api/python"
-    )
-    _run(
-        "pip3 install -r test-requirements.txt"
+        "pip3 install -U api/python[test]"
     )
 
 
 def prvsnr_coverage():
     print("Executing provisioner test cases and generating coverage report")
     _run(
-        "coverage run -m pytest \
+        "coverage run -m pytest -m unit\
         --cov test/ --cov-report=xml"
     )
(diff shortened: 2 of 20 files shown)


cli/src/code_coverage/prvsnr-coverage.py
```diff
@@ -32,7 +32,7 @@ def install_coverage():
         "yum install -y gcc cpp python3-devel"
     )
     _run(
-        "pip3 install -U api/python"
+        "pip3 install -U api/python[test]"
```
andkononykhin2, line 35: please compare `test` extra in setup.py with test-requirements.txt and add missed dependencies in the former. So the next step (`pip3 install -r test-requirements.txt`) can be removed.

andkononykhin2 (changes requested):
@rahuljyoti65 That is pretty well. Thanks. Just one improvement ask to address. Please

cli/src/code_coverage/prvsnr-coverage.py
rahuljyoti65: Done !

api/python/setup.py
```diff
@@ -40,6 +40,11 @@
     'pytest-xdist==1.29.0',
     'pytest-timeout==1.3.4',
     'pytest-mock==3.1.0',
+    'attrs==19.1.0',
```
andkononykhin2, line 43: no, please remove that, attrs is [part of our repo](https://github.com/Seagate/cortx-prvsnr/tree/pre-cortx-1.0/api/python/provisioner/vendor/attr) actually

api/python/setup.py
```diff
@@ -40,6 +40,11 @@
...
+    'PyYAML==5.1.2',
```
andkononykhin2, line 44: please remove, PyYAML is one of our main dependencies already

api/python/setup.py
```diff
@@ -40,6 +40,11 @@
...
+    'salt==2019.2.0',
```
andkononykhin2, line 45: the same for salt

andkononykhin2 (changes requested):

api/python/setup.py
rahuljyoti65: Done

api/python/setup.py
rahuljyoti65: Done.

api/python/setup.py
rahuljyoti65: Done

andkononykhin2 (approved):

PR #1022, opened 2021-03-18
fixes base logic of cli args parsing

andkononykhin2:
one more failure case was identified

### DIFF
--- a/api/python/provisioner/inputs.py
+++ b/api/python/provisioner/inputs.py
@@ -279,11 +279,14 @@ class ParserFiller:
     @staticmethod
     def from_args(cls, parsed_args: Union[dict, argparse.Namespace], pop=True):
         if isinstance(parsed_args, argparse.Namespace):
-            parsed_args = vars(parsed_args)
+            _parsed_args = vars(parsed_args)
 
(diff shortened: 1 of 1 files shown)


api/python/provisioner/inputs.py
```diff
@@ -314,7 +314,10 @@ def fill_parser(cls, parser, *args, **kwargs):
 
     @classmethod
     def from_args(cls, parsed_args, *args, **kwargs):
-        return ParserFiller.from_args(cls, parsed_args, *args, **kwargs)[0]
+        if cls is ParserFiller:
```
ypise, line 317: `if cls is ParserFiller:`

Should this be?
`if cls is not ParserFiller:`

As else is operating on same class.

ypise (changes requested):

api/python/provisioner/inputs.py
andkononykhin2: Thanks. It's a bug but a bit different, should be:
```
if cls is ParserMixin
```
and it mean that we don't expect parsing for the mixin itself

ypise (approved):

PR #1024, opened 2021-03-18
change logic to get cluster_id from pillar instead of grains

divya-vijay:


api/python/provisioner/commands/__init__.py
```diff
@@ -1381,11 +1370,14 @@ def _passwordless_ssh():
         command = getattr(_mod, cmd_cls)
     except AttributeError:
         try:
-            cmd_mod = importlib.import_module(
-                f'provisioner.commands.{cmd_name}'
-            )
+            if "cluster_id" in cmd_name:
```
andkononykhin2, line 1373: no, please don't add any specific logic here. options:
- either keep get/set commands for cluster-id separated
- introduce support of new optional values inside `type: [module].Type` so module can be parsed from here and used instead of default value which is command name , e.g
```
get_cluster_id:
  type: cluster_id.GetClusterId
```
- implement support for scanning of the whole `commands` directory (advanced task, don't think it should be here)


I believe the 2nd option is the best choice for the moment.

andkononykhin2 (changes requested):
@divya-vijay Thanks. Looks pretty good. But one place requires more accuracy.

api/python/provisioner/commands/__init__.py
divya-vijay: @andkononykhin Sure, thanks for the suggestion. Added support to handle types which are of `module.Type` format. In the hope that, as we migrate common/utility modules, this feature can be leveraged for those modules as well.

api/python/provisioner/commands/cluster_id.py
```diff
@@ -21,22 +21,77 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+@attr.s(auto_attribs=True)
+class GetClusterId(CommandParserFillerMixin):
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+
+    @staticmethod
+    def run(targets=ALL_MINIONS):
```
ypise, line 49: Can we eliminate the static decorator?
```
def run(self, targets=ALL_MINIONS):
```

api/python/provisioner/commands/cluster_id.py
```diff
@@ -21,22 +21,77 @@
...
-from . import CommandParserFillerMixin
-
-from ..vendor import attr
 from provisioner.salt import local_minion_id
 
 logger = logging.getLogger(__name__)
 
 
+@attr.s(auto_attribs=True)
+class GetClusterId(CommandParserFillerMixin):
```
ypise, line 45: Convert the class to `ClusterId`
No parameter mean get. Let's define it to be inline with properties in Python.

api/python/provisioner/commands/cluster_id.py
```diff
@@ -21,22 +21,77 @@
...
+        """
+        try:
+            cluster_data = PillarGet().run(
+                'cluster',
+                targets=targets
+            )
+            cluster_id = []
+
+            for node in cluster_data:
+                if "cluster_id" in cluster_data[node]["cluster"]:
```
ypise, line 67: Does it handle None value for key **_cluster_id_**?
Key could be set with a None value, in which case we might end up appending None or a blank value.

api/python/provisioner/commands/cluster_id.py
```diff
@@ -21,22 +21,77 @@
...
+                    )
+
+            if len(set(cluster_id)) != 1:
+                raise ValueError(
+                     "Error: ClusterID assignment not unique across "
+                     f"the nodes of the cluster: {cluster_id}. "
+                     "To set one, execute `provisioner set_cluster_id` command."
+                )
+
+            return cluster_id[0]
```
ypise, line 85: Avoid multiple return statements, it increases cyclometric complexity of the code.

ypise (changes requested):

api/python/provisioner/commands/__init__.py
```diff
@@ -15,19 +15,26 @@
...
+from typing import List, Dict, Type, Union
 
+from ._basic import (
+    RunArgs,
+    CommandParserFillerMixin,
+    RunArgsBase
+)
+from .check import (
+    Check,
+    SWUpdateDecisionMaker
```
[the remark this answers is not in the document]
andkononykhin2: @ypise I believe @divya-vijay did that just to cleanup the imports. Generally multiple lines are better to track changes (than one-line import for multiple modules)

api/python/provisioner/commands/__init__.py
```diff
@@ -1377,15 +1373,25 @@ def _passwordless_ssh():
 for cmd_name, spec in api_spec.items():
     spec = deepcopy(api_spec[cmd_name])  # TODO
     cmd_cls = spec.pop('type')
+
+    # If the format of 'type' is module.Type
+    if re.search(r'\.', cmd_cls):
+        mod_name = cmd_cls.split('.')[0]
```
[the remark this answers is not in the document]
andkononykhin2: Yes, regex is an overhead here. `split` should be goo enough. I would just suggest to improve what @ypise provided:

```python
mod_name = '.'.join(cmd_cls.split('.')[0:-1])
```

api/python/provisioner/commands/cluster_id.py
```diff
@@ -21,22 +21,77 @@
...
+
+        except ValueError as exc:
+            logger.error(
```
andkononykhin2, line 88: I don't think we need to hide all exceptions here. I would suggest to use `raise` instead of `return None` here

api/python/provisioner/commands/cluster_id.py
```diff
@@ -21,22 +21,77 @@
...
+                "Error in retrieving cluster_id from "
+                f"Pillar data: {str(exc)}"
+            )
+            return None
+
+
 @attr.s(auto_attribs=True)
 class SetClusterId(CommandParserFillerMixin):
```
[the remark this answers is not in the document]
andkononykhin2: @ypise No, please don't consider to add set logic inside a get command.

If you want to hide set command then you may just remove that from api_spec.yaml and our api.py (if it's going to be updated).

To call that implicitly only I guess we need some other set/config like public API where that set cluster id would be called internally.

andkononykhin2 (commented):
Few non critical comments were added

andkononykhin2 (approved):

andkononykhin2 (commented):
Few comments were added

api/python/provisioner/commands/cluster_id.py
divya-vijay: @andkononykhin The reason I moved `raise` to log error and `return None` is because we're calling this api in `set_cluster_id` , which in turn, is called at the end of bootstrap process (`setup_provisioner`). So, when it returns `None` as against an exception, we set the value to pillar as part of `SetClusterId` class. 

Can we handle this flow any other way.?

api/python/provisioner/commands/__init__.py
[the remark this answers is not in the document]
divya-vijay: @ypise yes, just rearranged imports to pythonic sequence.

divya-vijay:
Closing this PR as the new design is implemented in https://github.com/Seagate/cortx-prvsnr/pull/1186 
The design change explained in: https://github.com/Seagate/cortx-prvsnr/pull/1166

PR #1027, opened 2021-03-19
Update auto_assign.yml

mukul-seagate11:
To update the code owners as per information collected


mukul-seagate11:
@ypise, you can merge the PR now

PR #1035, opened 2021-03-22
EOS-17828: firewall ports adjustments

divya-vijay:
[Firewall-Ports-R2](https://seagatetechnology.sharepoint.com/:x:/s/gteamdrv1/tdrive1224/EZRT8UanK0RNi35n8a6uplQBLL-WmNhfjaXElJ7Iq70u2Q?e=tlNe7H) 

Strictly followed this list for open ports and tested on 3-node cluster. 
Removed all Private and Internal network ports and kept only incoming ports and data-public and mgmt-public ports. 
Apart from the list, only retained the initial default services ssh, salt (incoming) and glusterfs (not mentioned in sheet) as such.

### DIFF
--- a/pillar/components/firewall.sls
+++ b/pillar/components/firewall.sls
@@ -2,43 +2,9 @@ firewall:
   data_public:
     services: []
     ports:
-      consul:
-        - 8600/tcp
-        - 8600/udp
-        - 8500/tcp
-        - 8301/tcp
-        - 8301/udp
-        - 8302/tcp
-        - 8302/udp
-        - 8300/tcp
-      dhclient:
-        - 68/udp
-      dhserver:
-        - 67/udp
       haproxy:
         - 443/tcp
-      hare:
-        - 8008/tcp
-      nfs:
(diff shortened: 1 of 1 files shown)


pillar/components/firewall.sls
```diff
@@ -2,43 +2,9 @@ firewall:
...
         - 443/tcp
-      hare:
-        - 8008/tcp
-      nfs:
-        - 2049/tcp
-        - 2049/udp
-        - 32803/tcp
-        - 892/tcp
-        - 875/tcp
-      s3:
```
ypise, line 28: `9443/tcp` has to be part of s3 and not haproxy.

ypise (changes requested):

pillar/components/firewall.sls
divya-vijay: @ypise addressed ! Thanks.

ypise (approved):

andkononykhin2 (approved):

PR #1036, opened 2021-03-23
EOS-18744: I adopted logic to verify if the same upgrade repo has already exists

dmitrydb:
I adopted logic to verify if the same upgrade repo has already exists


andkononykhin2 (approved):

nitin-seagate (approved):
Looks good to me

PR #1037, opened 2021-03-23
[HOTFIX] Fix for 3 node vm for configure_setup

mazinamdar:


83bhp (approved):

api/python/provisioner/commands/configure_setup.py
```diff
@@ -351,7 +351,7 @@ def _key_int_to_list(self, contents):
     def _parse_input(self, input):
         kv_to_dict = dict()
         for key in input.keys():
-            if "," in input.get(key):
+            if input.get(key) and "," in input.get(key):
```
andkononykhin2, line 354: it would work, I just would suggest:
- do not use `input` since it is one of built-in functions
- iterate using `for k, v in input.items()` to avoid calling get multiple times

nitin-seagate (approved):
Looks good to me...

andkononykhin2 (approved):

ypise (approved):

PR #1041, opened 2021-03-23
Fix for durability keys in confstore

ypise:


83bhp (approved):
Looks good, do we not need corresponding changes anywhere else like pillars?

nitin-seagate (approved):
Looks good to me

PR #1043, opened 2021-03-25
EOS-15746: Test docs update & code fix

dmitrydb:
I updated the `r2.txt` testing docs and made small fixes for `set_swupgrade_repo` command

### DIFF
--- a/api/python/provisioner/inputs.py
+++ b/api/python/provisioner/inputs.py
@@ -1003,7 +1003,8 @@ class SWUpgradeRepo(SWUpdateRepo):
             }
         },
         validator=attr.validators.optional(
-            attr.validators.in_(config.HashType)),
+            attr.validators.in_(config.HashType)
+        ),
         default=None,
         converter=lambda x: x and config.HashType(str(x))
     )
--- a/api/python/provisioner/inputs.py
+++ b/api/python/provisioner/inputs.py
@@ -1002,7 +1002,8 @@ class SWUpgradeRepo(SWUpdateRepo):
                 'choices': list(map(lambda elem: elem.value, config.HashType))
             }
         },
-        validator=attr.validators.optional(attr.validators.instance_of(str)),
+        validator=attr.validators.optional(
+            attr.validators.in_(config.HashType)),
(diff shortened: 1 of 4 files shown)


andkononykhin2 (approved):

api/python/provisioner/inputs.py
```diff
@@ -1002,7 +1002,8 @@ class SWUpgradeRepo(SWUpdateRepo):
                 'choices': list(map(lambda elem: elem.value, config.HashType))
             }
         },
-        validator=attr.validators.optional(attr.validators.instance_of(str)),
+        validator=attr.validators.optional(
+            attr.validators.in_(config.HashType)),
```
ypise, line 1006: Just a suggestion. 
It helps if the closing ')' is on an independent line.

test/r2.md
```diff
@@ -0,0 +1,440 @@
...
+```bash
+pytest test/build_helpers/build_bundles
+```
+
+For example, it is enough to delete line which starts with `VERSION:` in file `cortx_iso/RELEASE.INFO`.
+And create the new ISO file with this change.
+
+The following command with broken `single_cortx.iso` should be failed
+```bash
+provisioner set_swupgrade_repo 2.0.1 --source=/root/single_cortx.iso
```
ypise, line 439: The iso path would be: `/opt/isos`
Try not to have anything under `/root`

api/python/provisioner/inputs.py
dmitrydb: It is a matter of taste. PEP8 allows both ways. I love the compact approach. But I fixed it here.

ypise (approved):

test/r2.md
dmitrydb: It is only for testing purposes, I provided this path as just an example

dmitrydb: I changed path to `/opt/iso/single_cortx.iso`.

ypise: > It is only for testing purposes, I provided this path as just an example

Agreed, but it's a good practice generally to avoid `/root`.

PR #1050, opened 2021-03-30
Test File Update: added test scenarios for clusterID and ports adjustments

divya-vijay:

-----
[View rendered test/r2.md](https://github.com/Seagate/cortx-prvsnr/blob/r2-test-file-update/test/r2.md)

### DIFF
--- a/test/r2.md
+++ b/test/r2.md
@@ -479,20 +479,31 @@ Step 3: Verify the confstore is generated across all nodes
 
 ## EOS-17828: Port adjustments post discussions for 3 node
 
-Step 1: Ensure firewall component is present in all the nodes of a 3 node VM setup. If not, install with the below command
+Test scenarios for [EOS-18494](https://jts.seagate.com/browse/EOS-18494) covers all validations related to Firewall ports.
+This is a simple verification to ensure only limited ports are open and mapped accordingly.
+
+Step 1: Ensure firewall component is present in all the nodes of a 3 node VM setup. If not, install with the below command.
 ```bash
 salt '*' state.apply components.system.firewall
 ```
 
-Step 2: Verify the services with corresponding ports, are mapped to all the zones with the below command.
+Step 2: Verify the services with corresponding ports, are mapped to the zones, to all the nodes in the cluster, with this command.
 ```bash
-firewall-cmd --list-all
+for node in srvnode-1 srvnode-2 srvnode-3
+do
+    for zone in "management-zone" "public-data-zone"
+    do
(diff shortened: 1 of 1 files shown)


test/r2.md
```diff
@@ -439,3 +441,27 @@ The following command with broken `single_cortx.iso` should be failed
...
+```
+
+Step 2: Verify the services with corresponding ports, are mapped to all the zones with the below command.
+```bash
+firewall-cmd --list-all
+```
+
+Step 3: Verify if the services are mapped to the zones on all the nodes.
+```bash
+ssh srvnode-2 firewall-cmd --list-all
```
ypise, line 460: Similar to above:
```
for node in srvnode-1 srvnode-2 srvndoe-3
do
    for zone in "management-zone" "public-data-zone" 
    do
        ssh ${node} firewall-cmd --list-all --zone ${zone}
    done
done
```

After this a separate `Step 2` shall no more be necessary.

test/r2.md
```diff
@@ -439,3 +441,27 @@ The following command with broken `single_cortx.iso` should be failed
...
+ssh srvnode-3 firewall-cmd --list-all
+```
```
ypise, line 462: Ideally, `EOS-18494: Validation for Open Ports in Firewall - Salt Execution Module` should be the only required deliverable for validation and the individual commands shall not be required.

ypise (changes requested):

test/r2.md
```diff
@@ -439,3 +441,27 @@ The following command with broken `single_cortx.iso` should be failed
...
+## EOS-17828: Port adjustments post discussions for 3 node
+
+Step 1: Ensure firewall component is present in all the nodes of a 3 node VM setup. If not, install with the below command
+```bash
+salt '*' state.apply components.system.firewall
+```
+
+Step 2: Verify the services with corresponding ports, are mapped to all the zones with the below command.
+```bash
+firewall-cmd --list-all
```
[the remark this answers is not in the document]
divya-vijay: @ypise Thanks! command updated now. I'd tested with `firewall-cmd --list-all-zones`, but since limited zones were affected, that might not be necessary.

test/r2.md
divya-vijay: Comment addressed.

test/r2.md
divya-vijay: Yeah, but I couldn't find scenarios for EOS-18494 yet in `r2.md` . So, for now, added this as simple verification step.

ypise (approved):

PR #1054, opened 2021-03-30
[EOS-18997][EOS-19070] Initial framework for north bound interface TUI

mazinamdar:

Initial framework checkin for TUI which includes
1. Main menu and sub menu
2. static netwok ip window
3. Is primary window
4. Time server window
5. Firewall window
6. Management VIP window
7. hostname window (this window is used as default window for most of menu as many windows are in progress)

starting point for this  API is not finalized yet, till that time we can use python interpreter to start TUI.

menu_1
menu_2
menu_3

-----
[View rendered api/python/provisioner/ui/README.md](https://github.com/Seagate/cortx-prvsnr/blob/EOS-18997/api/python/provisioner/ui/README.md)


api/python/provisioner/ui/README.md
```diff
@@ -0,0 +1,10 @@
+# Provisioner TUI
+
+This is northbound interface TUI for CORTX components that provides terminal UI
```
ypise, line 3: Restructure required:
TUI (Terminal User Interface) provides user with a tool, which enables interaction with node configuration commands. This abstracts the need for user executing commands on shell terminal with necessary parameters; thus minimizing the possibilities of user errors and promoting usability.

api/python/provisioner/ui/README.md
```diff
@@ -0,0 +1,10 @@
...
+to configure system.
+
+## start with provisioner TUI
+
+1. command to start provisioner TUI
+
+-  `python3 api/python/provisioner/ui/main.py`
```
ypise, line 10: The command should not be main.py.
Please move this to under the command API and expose it under command:
`provisioner configure_node`
later to be converted to:
`cortx_setup configure_node`

api/python/provisioner/ui/color_code.py
```diff
@@ -0,0 +1,38 @@
...
+#
+#
+import curses
+import config
+from error import ColorCodeNotDefined
+
+
+class ColorCode:
+
+    @staticmethod
```
ypise, line 26: Avoid static methods inside a class.

api/python/provisioner/ui/config.py
```diff
@@ -0,0 +1,74 @@
...
+    BLACK_N_WHITE = 1
+    GREEN_N_BLACK = 2
+    RED_N_BLACK = 3
+    GREEN_N_WHITE = 4
+
+
+class Key(Enum):
+    EXIT_1 = 10
+    EXIT_2 = 13
+    QUITE = 113
```
ypise, line 33: Do we mean **QUIT** here?
Could you add a one-liner for what these codes mean?

api/python/provisioner/ui/config.py
```diff
@@ -0,0 +1,74 @@
...
+    'Set Management VIP': 'mgmt_vip_primary:IsPrimaryWindow',
+    'Setup Netowrk': menu_network,
+    'Set Host Primary': 'is_primary:PrimaryWindow',
+    'Setup storage': 'hostname:HostnameWindow',
+    'Setup firewall': 'confirm_firewall:FirewallWindow',
+    'Setup Time Server': 'time_server:TimeServerWindow'
+}
+
+
+color_codes = {
```
ypise, line 61: Can we move it to ColorCodes class or at least right after the class for readability.

api/python/provisioner/ui/config.py
```diff
@@ -0,0 +1,74 @@
...
+    ColorCodes.GREEN_N_BLACK.value: [curses.COLOR_GREEN, curses.COLOR_BLACK],
+    ColorCodes.RED_N_BLACK.value: [curses.COLOR_RED, curses.COLOR_BLACK],
+    ColorCodes.GREEN_N_WHITE.value: [curses.COLOR_GREEN, curses.COLOR_WHITE]
+}
+
+
+error_color = ColorCodes.RED_N_BLACK.value
+default_window_color = ColorCodes.GREEN_N_BLACK.value
+menu_color = ColorCodes.GREEN_N_BLACK.value
+default_menu_head = ColorCodes.BLACK_N_WHITE.value
```
ypise, line 72: Can we move it to ColorCodes class or at least right after the class for readability.

ypise (changes requested):

api/python/provisioner/ui/error.py
```diff
@@ -0,0 +1,26 @@
...
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+#
+
+
+class PrvsnrException(Exception):
```
andkononykhin2, line 21: we already have [a set of base exceptions](https://github.com/Seagate/cortx-prvsnr/blob/ddd9651ef721c2778a420f04624816b72695fb74/api/python/provisioner/errors.py#L50-L66). Please don't introduce a new one.

api/python/provisioner/ui/form_window.py
```diff
@@ -0,0 +1,202 @@
...
+        color_code = kwargs['color_code']
+        self._parent = kwargs['component']
+
+        selected_rows, edit = FormWindow.parse_input(kwargs)
+        self._window.border()
+        col_code_attr = ColorCode.get_color_pair(color_code)
+        self.create_menu_head()
+
+        x = self.get_max_width() // 2
+        y = self.get_max_height() // 3 - 1
```
andkononykhin2, line 101: That code is not so easy, I would suggest:
- add some comments
- do not use hard-coded values (like `// 3`) in multiple places. Please define some human-understandable constants. So it would become easier to understand and maintain.

api/python/provisioner/ui/form_window.py
```diff
@@ -0,0 +1,202 @@
...
+        self._window.addstr(y, 3, f"Please enter {self.component_type}"
+                            " for this machine")
+        self.off_attr(col_code_attr)
+
+        values = list(self.data.keys())
+        count_m = len(values)
+        x = 6
+
+        for idx, row in enumerate(values):
+            y = self.get_max_height() // 3 - count_m // 2 + (idx + 1) * 2
```
andkononykhin2, line 114: why we evaluate the same multiple times (`self.get_max_height() // 3`)? Is there a chance that the result may change in the middle. If now - please use some well-named variable instead.

api/python/provisioner/ui/form_window.py
```diff
@@ -0,0 +1,202 @@
...
+            self._window.addstr(y, x, f"{values[selected_rows]}:")
+
+            if edit:
+                if ('validation' in self.data[values[selected_rows]]):
+                    validate = self.data[values[selected_rows]]['validation']
+                else:
+                    validate = None
+                user_content = TextBox(
+                    self,
+                    1,
```
andkononykhin2, line 139: hard-codes: please unwrap the magic

api/python/provisioner/ui/form_window.py
```diff
@@ -0,0 +1,202 @@
...
+                    x + 20,
+                    self.get_max_height() // 4
+                ).create_textbox(
+                    color_code,
+                    self.data[values[selected_rows]]['default'],
+                    validate
+                )
+                self.data[values[selected_rows]]['default'] = user_content
+            else:
+                self._window.addstr(y, 24,
```
andkononykhin2, line 151: hard-code

api/python/provisioner/ui/form_window.py
```diff
@@ -0,0 +1,202 @@
...
+    def process_input(self, color_code):
+        current_row = 0
+        values = list(self.data.keys())
+
+        while 1:
+            key = self._window.getch()
+            self._window.clear()
+
+            if key == curses.KEY_UP and current_row > 0:
+                current_row = current_row - 1
```
andkononykhin2, line 164: as an option: `current_row -= 1`

api/python/provisioner/ui/form_window.py
```diff
@@ -0,0 +1,202 @@
...
+
+            elif current_row == len(values) and key == curses.KEY_LEFT:
+                current_row = current_row + 1
+
+            elif current_row == len(values) + 1 and key == curses.KEY_RIGHT:
+                current_row = current_row - 1
+
+            elif key == curses.KEY_ENTER or key in (config.Key.EXIT_1.value,
+                                                    config.Key.EXIT_2.value):
+                if current_row < len(values) and current_row >= 0:
```
andkononykhin2, line 177: could you comment each case?

api/python/provisioner/ui/header.py
```diff
@@ -0,0 +1,34 @@
...
+from color_code import ColorCode
+
+
+class HeaderWindow(Window):
+
+    def create_window(self, **kwargs):
+        color_code = kwargs['color_code']
+        col_code_attr = ColorCode().get_color_pair(color_code)
+        self.on_attr(col_code_attr)
+        self._window.addstr(self._max_h // 2,
```
andkononykhin2, line 30: hard-codes: what if we decide to change a divider later or make it dynamic?

api/python/provisioner/ui/main.py
```diff
@@ -0,0 +1,51 @@
...
+
+def main(stdscr):
+    curses.initscr()
+    curses.curs_set(0)
+    ColorCode.init()
+    for code, color in config.color_codes.items():
+        ColorCode.create_color_pair(code, color[0], color[1])
+
+    wind = Window(stdscr)
+    header_window = curses.newwin(wind._max_h // 5, wind._max_w - 1, 1, 1)
```
andkononykhin2, line 35: magic

api/python/provisioner/ui/main.py
```diff
@@ -0,0 +1,51 @@
...
+    HeaderWindow(header_window).create_window(color_code=config.menu_color)
+    main_window = curses.newwin(
+                      wind._max_h - wind._max_h // 5,
```
andkononykhin2, line 38: and here

api/python/provisioner/ui/main_menu.py
```diff
@@ -0,0 +1,140 @@
...
+        col_code_attr = ColorCode.get_color_pair(color_code)
+        self.create_menu_head()
+
+        for idx, row in enumerate(self._menu):
+            x = self.get_max_width() // 2 - len(row) // 2
+            y = self.get_max_height() // 2 - len(self._menu) // 2 + idx
+
+            if idx == selected_rows:
+                self.on_attr(col_code_attr)
+                self._window.addstr(y, x-3, ">> ")
```
andkononykhin2, line 57: hard codes plus missed spaces around binary operator

api/python/provisioner/ui/validation.py
```diff
@@ -0,0 +1,49 @@
...
+                if ip != str(value):
+                    result = False
+        except ValueError:
+            result = False
+        return result
+
+    @staticmethod
+    def hostname(hostname):
+        result = True
+        hostname_regex = r"^([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z]{2,6}$"
```
andkononykhin2, line 43: how are we confident with such a regex? I guess there might be a set of cases we we don't expect for now. e.g. this one in [django](https://github.com/django/django/blob/6726d750979a7c29e0dd866b4ea367eef7c8a420/django/core/validators.py#L47-L49). Can't we use some python build-in module for that?

andkononykhin2 (changes requested):
@mazinamdar Thank you. It is really a big start for the topic! But the main problems are:
- lack of comments for some specific parts
- many hard-codes

Please check the comemnts.

api/python/provisioner/ui/README.md
mazinamdar: comment is addressed

api/python/provisioner/ui/color_code.py
mazinamdar: comment is addressed , kept only one utility method as static as it is mothing to do with class

api/python/provisioner/ui/config.py
mazinamdar: Added below ColorCode class

api/python/provisioner/ui/config.py
mazinamdar: Added below ColorCode class

api/python/provisioner/ui/form_window.py
mazinamdar: Tried to add comments at most of the places. 
Added variable which in places of hard coded values.

FYI, In some cases still few hard coded values may appear as I am trying to adjust most of the object at right place on screen.

api/python/provisioner/ui/form_window.py
mazinamdar: added variable in place of this

api/python/provisioner/ui/form_window.py
mazinamdar: This is length of textbox window, added this value in config fetching from config now.

api/python/provisioner/ui/form_window.py
mazinamdar: used variable in place of this

api/python/provisioner/ui/form_window.py
mazinamdar: comment is addressed

api/python/provisioner/ui/form_window.py
mazinamdar: Done

api/python/provisioner/ui/header.py
mazinamdar: This will just add tittle at middle of header window.

In main we  are dividing screen into 2 windows 
1. header (1/5th)
2. Main window(4/5th)

Changing window size from main will adjust everything according to new window size

api/python/provisioner/ui/main.py
mazinamdar: added comments for more clarity

api/python/provisioner/ui/main.py
mazinamdar: added comments for more clarity

api/python/provisioner/ui/main_menu.py
mazinamdar: comment is addressed

api/python/provisioner/ui/validation.py
mazinamdar: not able to find better way to validate hostname. I don't think we have any inbuild python package for this.

Can you please suggest some better way for validation of hostname?

api/python/provisioner/ui/README.md
mazinamdar: facing some issue directly calling TUI from API, will cover this comment in next ticket. EOS-18273

api/python/provisioner/ui/error.py
andkononykhin: addressed

api/python/provisioner/ui/form_window.py
andkononykhin: addressed

api/python/provisioner/ui/form_window.py
andkononykhin: addressed

api/python/provisioner/ui/form_window.py
andkononykhin: addressed

ypise (approved):

api/python/provisioner/ui/validation.py
ypise: I believe, what we are validating here is just that the hostname is actually in FQDN format with no captital letters or special characters.
Does this handle multiple '.' though?

mazinamdar: yes , it handle multiple dots.

andkononykhin2: >Can you please suggest some better way for validation of hostname?

I provided more  comprehensive one from django. We can keep the current one for now but please add some TODO comment. thanks

andkononykhin2 (approved):

api/python/provisioner/ui/validation.py
mazinamdar: comment is addressed, Added TODO comment

PR #1060, opened 2021-03-31
EOS-18920: Patch-1: create separate validation modules and remove from configure_setup api

divya-vijay:

### DIFF
--- a/api/python/provisioner/commands/validator/__init__.py
+++ b/api/python/provisioner/commands/validator/__init__.py
@@ -25,6 +25,10 @@ from .validator import (FileValidator,
                         ReleaseInfoValidator
                         )
 
+from .network import NetworkParamsValidator
+from .node import NodeParamsValidator
+from .enclosure import StorageEnclosureParamsValidator 
+
 __all__ = [
     "FileValidator",
     "DirValidator",
@@ -34,5 +38,8 @@ __all__ = [
     "YumRepoDataValidator",
     "ContentFileValidator",
     "ReleaseInfoContentScheme",
-    "ReleaseInfoValidator"
+    "ReleaseInfoValidator",
+    "NetworkParamsValidator",
+    "NodeParamsValidator",
+    "StorageEnclosureParamsValidator"
 ]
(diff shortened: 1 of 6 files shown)


ypise (approved):

api/python/provisioner/commands/validator/__init__.py
```diff
@@ -14,25 +14,31 @@
 # For any questions about this software or licensing,
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
+
 from .validator import (FileValidator,
                         DirValidator,
                         FileSchemeValidator,
                         YumRepoDataValidator,
                         HashSumValidator,
-                        YumRepoDataValidator,
```
andkononykhin2, line 22: why it was removed?

andkononykhin2 (changes requested):
@divya-vijay Thanks. Could you please just address few comments?

api/python/provisioner/commands/validator/network.py
```diff
@@ -0,0 +1,51 @@
...
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+# Module with all network-related validations
+
+
+import logging
+
+from ...inputs import (
```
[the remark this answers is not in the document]
divya-vijay: yes, thought so :) But in existing `commands/validators/validator.py`, it was this way. so, kept it as such to maintain uniformity. Addressed now.

api/python/provisioner/commands/validator/__init__.py
divya-vijay: Duplicated twice, so removed once.

andkononykhin2: Oh, right. Thanks

api/python/provisioner/commands/validator/network.py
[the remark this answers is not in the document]
andkononykhin2: I see. We missed that before. Thank you for the update.

andkononykhin2 (approved):

PR #1062, opened 2021-03-31
North-Bound API for node - Stage 1

SomwanshiAnjali:

North-Bound API for:
- set_hostname
- setup_firewall


api/python/provisioner/commands/setup_timeserver.py
```diff
@@ -0,0 +1,52 @@
...
+class SetupTimeServer(CommandParserFillerMixin):
+    """Defines logic to setup time server on system"""
+
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+    _run_args_type = RunArgsEmpty
+
+    def run(self):
+        logger.info("Setting up NTP time server")
+        try:
+            StatesApplier.apply(['components.system.chrony'], targets=LOCAL_MINION)
```
andkononykhin2, line 48: we have similar logic as `set_ntp` command and it allows configuration during the runtime (on field). On any change of configuration it calls [config, restart for noes and config for controller](https://jts.seagate.com/secure/XrayTestRepositoryAction!default.jspa?entityKey=TEST&path=EOS%2FEES%2FProvisioner#). 

I guess that new API is considered for deployment only? Is it true? What about controllers configuration?

api/python/provisioner/commands/setup_firewall.py
```diff
@@ -0,0 +1,48 @@
...
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+    _run_args_type = RunArgsEmpty
+
+    def run(self):
+        logger.info("Setting up CORTX firewall")
+        try:
+            StatesApplier.apply(['components.system.firewall'], targets=LOCAL_MINION)
+        except Exception as exc:
+            logger.error("Firewall configuration failed")
+            raise SaltError(exc)
```
andkononykhin2, line 47: to things here:
- a first argument here is a human readable message, makes sense to think is it ok to pass an exception object
- why do we need to wrap an `exc` into `SaltError`, can't we just re-raise the original one

api/python/provisioner/commands/set_hostname.py
```diff
@@ -0,0 +1,85 @@
...
+                fun_kwargs=dict(
+                    name='set_hostname',
+                    hostname=server_hostname,
+                    apply_hostname=True,
+                    retain_settings=True
+                ),
+                targets=LOCAL_MINION
+            )
+        except Exception as exc:
+            raise SaltError(exc)
```
andkononykhin2, line 83: please check my comment in `setup_firewall.py` for similar place. Additionally I would say that there it's likely no need to use try-except at all.

api/python/provisioner/commands/set_hostname.py
```diff
@@ -0,0 +1,85 @@
...
+
+            logger.info("Fetching server hostname from cluster pillar")
+            local_minion = local_minion_id()
+            cluster_path = PillarKey('cluster')
+            cluster_dict = PillarResolver(local_minion).get([cluster_path])
+            cluster_dict = cluster_dict[local_minion][cluster_path]
+            server_hostname = cluster_dict[server_id]['hostname']
+            if not server_hostname:
+                logger.error("Failed to get hostname from cluster pillar")
+                raise SaltError("Failed to get hostname from cluster pillar")
```
andkononykhin2, line 69: I don't think it's a salt error. I think something more related from errors.py or define some more specific error if needed. (e.g. some `PillarError`)

api/python/provisioner/commands/set_hostname.py
```diff
@@ -0,0 +1,85 @@
...
+class SetHostname(CommandParserFillerMixin):
+    """Defines logic to set hostname for system"""
+
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+    _run_args_type = RunArgsEmpty
+
+    def run(self):
+        try:
+            logger.info("Fetching server_id from grains data")
+            server_id = list(function_run(
```
andkononykhin2, line 52: `local_minion_id()` does the similar operation, you can use it instead

andkononykhin2 (changes requested):
@SomwanshiAnjali Thanks. It looks good. But I have few concerns though. Please check

api/python/provisioner/commands/setup_firewall.py
SomwanshiAnjali: Addressed

api/python/provisioner/commands/set_hostname.py
SomwanshiAnjali: Addressed

api/python/provisioner/commands/set_hostname.py
SomwanshiAnjali: Addressed

api/python/provisioner/commands/set_hostname.py
SomwanshiAnjali: Addressed

api/python/provisioner/commands/setup_timeserver.py
SomwanshiAnjali: @andkononykhin2 This api is for `Field: Node config` and here we are not considering time_server details at command-line. 
also currently the requirement does not include NTP configuration on controller 

hence we decided to keep both the logic's different.

api/python/provisioner/commands/set_hostname.py
```diff
@@ -0,0 +1,74 @@
...
+                fun_kwargs=dict(
+                    name='set_hostname',
+                    hostname=server_hostname,
+                    apply_hostname=True,
+                    retain_settings=True
+                ),
+                targets=LOCAL_MINION
+            )
+        except Exception as exc:
+            raise exc
```
andkononykhin2, line 73: as long as we just re-raise without any logging and additional logic please just remove try-except block

api/python/provisioner/commands/setup_firewall.py
```diff
@@ -0,0 +1,48 @@
...
+    """Defines logic to setup CORTX firewall on system"""
+
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+    _run_args_type = RunArgsEmpty
+
+    def run(self):
+        logger.info("Setting up CORTX firewall")
+        try:
+            StatesApplier.apply(['components.system.firewall'], targets=LOCAL_MINION)
+        except Exception as exc:
```
andkononykhin2, line 45: you don't need `exc` for re-raise since just `raise` will do the same. Also `logger.error` doesn't provide any details.
I would suggest something like:

  either

```python
except:
    logger.exception("Firewall configuration failed: {}")
    raise
```
OR
```python
except Exception as exc:
    logger.error(f"Firewall configuration failed: {exc}")
    raise
```

api/python/provisioner/commands/setup_timeserver.py
```diff
@@ -0,0 +1,52 @@
...
+        except Exception as exc:
+            logger.error("NTP configuration failed")
```
andkononykhin2, line 50: please check my comment for firewall setup

andkononykhin2 (changes requested):
@SomwanshiAnjali please check my comments: they are minor ones but important to keep python style

api/python/provisioner/commands/setup_timeserver.py
andkononykhin2: I tend to resolve that discussion for now. Just few general thoughts:

- it's not a good UX if we have multiple API for the similar things: that configuration anyway requires settings in pillar the difference is:
  -  in one case we consider a change to pillar
  - in another - we don't consider change but just want to apply/re-apply

I don't think we should publish these APIs for now. I believe we need more discussion on that and implement just one API that satisfies both cases. (Actually I think `set_ntp` already does that: you can pass special value `PRVSNR_UNCHANGED` so it won't update the pillar but just re-apply).

andkononykhin2: CC @ypise

api/python/provisioner/commands/set_hostname.py
```diff
@@ -0,0 +1,74 @@
...
+            cluster_path = PillarKey('cluster')
+            cluster_dict = PillarResolver(local_minion).get([cluster_path])
+            cluster_dict = cluster_dict[local_minion][cluster_path]
+            server_hostname = cluster_dict[local_minion]['hostname']
+            if not server_hostname or server_hostname is values.MISSED::
+                logger.error("Failed to get hostname from cluster pillar")
+                raise ValueError("Failed to get hostname from cluster pillar")
+
+            logger.info(f"Setting hostname to {server_hostname}")
+            StateFunExecuter.execute(
```
andkononykhin2, line 62: can't we implement that as state.apply to some state (similar to other new apis): that state can implement essentially the same logic:

- verify that necessary pillars are presented and conditionally fail
- call `network.system` module otherwise

api/python/provisioner/commands/setup_timeserver.py
SomwanshiAnjali: @andkononykhin2  removed setup_timeserver.py so that we can reuse `set_ntp`

api/python/provisioner/commands/set_hostname.py
SomwanshiAnjali: remove SetHostname class and update Set wrapper to perform set_hostname

api/python/provisioner/commands/setup_firewall.py
SomwanshiAnjali: Updated to use Set wrapper class for firewall

api/python/provisioner/commands/setup_timeserver.py
SomwanshiAnjali: removed setup_timeserver.py to reuse current set_ntp api

api/python/provisioner/commands/set_hostname.py
SomwanshiAnjali: implemented as state.apply in srv/components/system/config/set_hostname.sls

andkononykhin2 (approved):

api/python/provisioner/commands/set_hostname.py
andkononykhin2: Thanks

ypise (approved):

ypise (approved):

PR #1064, opened 2021-04-01
EOS-16939: Change signature of set SW update and upgrade commands

dmitrydb:
I removed 'release' parameter from list of necessary parameters for `set_swupgrade_repo` and `set_swupdate_repo` commands

### DIFF
--- a/api/python/provisioner/api.py
+++ b/api/python/provisioner/api.py
@@ -210,7 +210,7 @@ def set_network(dry_run=False, nowait=False, **kwargs):
 
 
 def set_swupdate_repo(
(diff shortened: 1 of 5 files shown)


andkononykhin2 (changes requested):
@dmitrydb It looks good. I just think that we might not need to change the signature for the old swupdate routine but only for new swupgrade API (since we might break CSM in some cases). 

Could you change only swupgrade inputs and related code?

ypise (approved):

andkononykhin2 (approved):

PR #1066, opened 2021-04-01
EOS-19107:Adding confstore keys for kafka and modifying the template delimiter

sradhanand-pati:


ypise (approved):

api/python/provisioner/commands/confstore_export.py
```diff
@@ -54,6 +55,24 @@ class ConfStoreExport(CommandParserFillerMixin):
     input_type: Type[inputs.NoParams] = inputs.NoParams
     description = "Export pillar template data to ConfStore"
 
+    def _encrypt_value(self, key, value):
+        from cortx.utils.security import cipher
```
mazinamdar, line 59: can we use dynamic imports here using importlib?

ypise: @sradhanand-pati 
Make sure to try this in your next PR. While you do that, ensure the code still fails with some error if the package is missing.

ypise (approved):

mazinamdar (approved):
looks good, just make sure to update code in next PR for import

api/python/provisioner/commands/confstore_export.py
```diff
@@ -54,6 +55,24 @@ class ConfStoreExport(CommandParserFillerMixin):
...
+        if not value:
+            return value
+        cypher_name = key.split('>')[0]
+        if 'bmc' in key:
```
dmitrydb, line 63: I suggest here to use constants for all strings like: `bmc`, `storage_enclosure`, `cluster_id`, etc.

api/python/provisioner/commands/confstore_export.py
```diff
@@ -98,12 +117,23 @@ def run(self, **kwargs):
 
             Conf.load('provisioner', pillar[PillarKey(pillar_key)])
 
-            for data in template_data:
+            delimiter = '=>'
+            for i,data in enumerate(template_data):
                 if data:
-                    Conf.set("provisioner", data.split(':')[0], data.split(':')[1])
+                    key = data.split(delimiter)[0]
+                    value = data.split(delimiter)[1]
```
dmitrydb, line 124: You can simplify code here to avoid multiple `split`-calls like this
```python
key, value = data.split(delimeter)
```

api/python/provisioner/commands/confstore_export.py
```diff
@@ -98,12 +117,23 @@ def run(self, **kwargs):
 
             Conf.load('provisioner', pillar[PillarKey(pillar_key)])
 
-            for data in template_data:
+            delimiter = '=>'
```
dmitrydb, line 120: I think it is better to put `'=>'` and `'>'` to constants. Why do we use these characters as delimiters? Based on the key representation used in consul for keys we can use `/`-character as keys delimiter.

api/python/provisioner/commands/confstore_export.py
dmitrydb, line 63: For future improvements: instead of multiple `if-elif-esle` blocks we can suggest using for parsing some schemes independent from names like `bmc` and `storage_enclosure`.

api/python/provisioner/commands/confstore_export.py
```diff
@@ -98,12 +117,23 @@ def run(self, **kwargs):
...
+                        value = self._encrypt_value(key, value)
+                        template_data[i] = key + delimiter + value
+                    Conf.set("provisioner", key, value)
+
             Conf.save("provisioner")
             logger.info("Template loaded to confstore")
 
+            with open(template_file_path,'w') as stream:
+                for data in template_data:
+                    stream.writelines(data + '\n')
```
dmitrydb, line 135: Are you sure that this API will work here? `stream.writelines` expects an iterable of strings. So here, based on logic `data + '\n'` it looks like we are passing just a single string instead of list/tuple/any iterable object of strings. Please, double-check this logic, and if we write here only 1 string with a newline symbol consider using `stream.write()` method instead

dmitrydb (changes requested):
Please, consider taking into account my few comments. Thank you.

api/python/provisioner/commands/confstore_export.py
sradhanand-pati: Yes. i will do it @mazinamdar @ypise

api/python/provisioner/commands/confstore_export.py
sradhanand-pati: >Are you sure that this API will work here?
Yes. it works.
but it makes sense to use write() method. I will update it.

api/python/provisioner/commands/confstore_export.py
sradhanand-pati: Yes. will update

api/python/provisioner/commands/confstore_export.py
sradhanand-pati: Here - EOS-19150 few things have been listed down which would go as improvement for this.
There i will incorporate these changes.

dmitrydb: Will be resolved in https://jts.seagate.com/browse/EOS-19150

dmitrydb: Will be resolved in
https://jts.seagate.com/browse/EOS-19150

api/python/provisioner/commands/confstore_export.py
dmitrydb: From Sradhanand
>This '>' delimiter is used by our confstore.
right now we are not using consul's confstore.
we are using the confstore which foundation team has created.

dmitrydb (approved):
Looks good for me. I approved this PR

api/python/provisioner/commands/confstore_export.py
sradhanand-pati: >I think it is better to put '=>' and '>' to constants. 

Yes i will address this against EOS-19150.

>Why do we use these characters as delimiters? Based on the key representation used in consul for keys we can use /-character as keys delimiter.

currently we are not using consul as our confstore. The confstore we are using has some specific delimiters which they have specified.

PR #1067, opened 2021-04-01
EOS-19102: Erroneous HAproxy configuration

ypise:

### Erroneous HAproxy configuration is binding to *:5000
Such configuration has been introduced with https://github.com/Seagate/cortx-prvsnr/pull/868/commits/17a9c556717b0dc65087e8458eb94891ba1d72f8.

Because HAproxy binds to *:5000, UDS cannot bind to port 5000 and hence CSM cannot reach its server.

### DIFF
--- a/srv/components/ha/haproxy/files/haproxy.cfg
+++ b/srv/components/ha/haproxy/files/haproxy.cfg
@@ -73,27 +73,27 @@ defaults
 #---------------------------------------------------------------------
 # main frontend which proxys to the backends
 #---------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


andkononykhin2 (approved):

83bhp:
So, we don't need frontend and backend section anymore?

ypise:
> So, we don't need frontend and backend section anymore?

Not needed as those are injected by S3 config and CSM config for UDS resp.

83bhp (approved):

PR #1069, opened 2021-04-02
[EOS-19135] Provisioner CLI to trigger northbound TUI

mazinamdar:

Used subprocess to call TUI script as salt_cmd is not working in this case.

`
 provisioner configure_node
`

### DIFF
--- a/api/python/provisioner/commands/configure_node.py
+++ b/api/python/provisioner/commands/configure_node.py
@@ -1,0 +1,47 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


api/python/provisioner/commands/configure_node.py
```diff
@@ -0,0 +1,47 @@
...
+
+    input_type: Type[inputs.NoParams] = inputs.NoParams
+
+    def run(self, **kwargs):
+        ui_script_path = str(CONFIG_MODULE_DIR / 'ui/main.py')
+
+        if not Path(ui_script_path).is_file():
+            raise ProvisionerError(f'{ui_script_path} file is missing')
+
+        run_subprocess_cmd([f"python3 {ui_script_path}"],
```
andkononykhin, line 45: Can't we just call ui/main.py::somefun which triggers curses? Why do you need a separate process?

andkononykhin (commented):
@mazinamdar please check my only question. Thanks

api/python/provisioner/commands/configure_node.py
mazinamdar: comment is addressed.

ypise (approved):

PR #1072, opened 2021-04-05
EOS-18920: Patch-3: module movement for bootstrap logic

divya-vijay:
1. More modules are created in `bootstrap/` path, to be updated after testing completely.
2. some suggestions are made for validations in the code implementation space, to be updated once list of validations are ready.


api/python/provisioner/commands/bootstrap/node_grains.py
```diff
@@ -0,0 +1,135 @@
+#
```
andkononykhin2, line 1: @divya-vijay please either name that file as `node.py` or create a separate one for `Node` class

andkononykhin2 (changes requested):
@divya-vijay Please check one request comment. The rest is clean. Thanks

api/python/provisioner/commands/bootstrap/node_grains.py
divya-vijay: Thanks! I've renamed it to `node.py`.

andkononykhin2 (approved):

ypise (approved):

PR #1078, opened 2021-04-06
[EOS-19173] build 607 - Update - Release Version is Still shown as 1.0

tanujashinde0405:


ypise:
Shall be tracked against #1082

PR #1089, opened 2021-04-07
[EOS-19218] Remove hctl cluster shutdown from hare

mazinamdar:

1. hctl cluster start / shutdown is not part of hare.
2. Improve error logging

### DIFF
--- a/api/python/provisioner/commands/destroy.py
+++ b/api/python/provisioner/commands/destroy.py
@@ -156,8 +156,8 @@ class DestroyNode(Deploy):
                 self.setup_ctx.ssh_client.state_apply(
                     f"components.{state}"
                 )
-        except Exception:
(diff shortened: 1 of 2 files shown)


ypise (approved):

api/python/provisioner/commands/destroy.py
```diff
@@ -156,8 +156,8 @@ def _apply_states(self, state, targets=None):
                 self.setup_ctx.ssh_client.state_apply(
                     f"components.{state}"
                 )
-        except Exception:
-            logger.warn(f"Failed {state} on {targets}")
+        except Exception as exc:
+            logger.warn(f"Failed {state} on {targets} Error: {str(exc)}")
```
andkononykhin2, line 160: for such cases in future please consider to use logger.exception instead

andkononykhin2 (approved):

PR #1107, opened 2021-04-12
EOS-18657: Better support of remote URLs for SW Upgrade API

anurag-modi:
Signed-off-by: Anurag Modi <anurag.modi@seagate.com>


ypise:
If a PR against a JIRA, it's required to have JIRA ticket ID in commit message and PR title.

api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -17,6 +17,7 @@
 import logging
 from pathlib import Path
 from typing import Type
+import requests
```
andkononykhin2, line 20: `requests` are good but it needs to be mentioned in our dependencies list in `setup.py`. In some environment latest versions of that package might be not available from python index repos (at least for the initial time: it should be noted by devops team).

So please:
- add that dependency as `requests>=2.14.2` (since epel-7 provides it as rpm) into `setup.py`

api/python/provisioner/inputs.py
```diff
@@ -1112,7 +1112,27 @@ def pillar_value(self):
 
             return res
         elif self.is_remote():
-            return self.source
+            return {
+                f'os': {
+                    'source': (f'{self.source}/os'),
```
andkononykhin2, line 1117: parenthesis are not needed, please remove

andkononykhin2, line 1117: we assume that remote repo is actually a remotely mounted ISO so directory names should be the same as for ISO. So, please use these constants instead of hard-coed values (e.g. here it should be `config.OS_ISO_DIR`)

api/python/provisioner/srv/salt/repos/upgrade.sls
```diff
@@ -21,11 +21,13 @@
 
 {% for version, repos in pillar['release']['upgrade']['repos'].items() %}
 
-    {% set single_iso_repo = {version: repos.pop(version)} %}
+    {% if not repos.startswith(('http://', 'https://')) %}
```
andkononykhin2, line 24: `repos` here is a dictionary but not a string, I guess the better way is to check whether that dictionary has `version` as one of the keys: if [true then it's an ISO](https://github.com/Seagate/cortx-prvsnr/blob/e447eb4f00aaeb955838abe77d3d2d7ae0e5917a/api/python/provisioner/inputs.py#L1140).

andkononykhin2 (changes requested):
@anurag-modi Thank you. General idea looks good. Thanks. But I feel it requires sufficient testing since few places in code looks suspicious.

api/python/provisioner/commands/set_swupgrade_repo.py
```diff
@@ -266,18 +268,26 @@ def dynamic_validation(self, params: inputs.SWUpgradeRepo, targets: str):  # noq
...
-                ) from e
+                    ) from e
 
-            release_file = (f'{iso_mount_dir}/{CORTX_ISO_DIR}/'
+                release_file = (f'{iso_mount_dir}/{CORTX_ISO_DIR}/'
                             f'{RELEASE_INFO_FILE}')
+            else:
+                release_file = '{candidate_repo}/cortx_iso/{RELEASE_INFO_FILE}'
+                r = requests.get(release_file)
+                with open("RELEASE.INFO",'wb') as f:
```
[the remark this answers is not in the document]
dmitrydb: I agree, we do not need to save here the content in `tmp` directory. I parsed the returned remote `RELEASE.INFO` content directly after requests execution

api/python/provisioner/commands/set_swupgrade_repo.py
dmitrydb: Done

api/python/provisioner/inputs.py
dmitrydb: Done

dmitrydb: Done

api/python/provisioner/srv/salt/repos/upgrade.sls
dmitrydb: Done

api/python/provisioner/inputs.py
```diff
@@ -1113,23 +1113,26 @@ def pillar_value(self):
             return res
         elif self.is_remote():
             return {
-                f'os': {
-                    'source': (f'{self.source}/os'),
+                f'{config.OS_ISO_DIR}': {
```
andkononykhin2, line 1116: just `config.OS_ISO_DIR: ...` should be good

api/python/provisioner/inputs.py
```diff
@@ -1113,23 +1113,26 @@ def pillar_value(self):
...
+                    'source': f'{self.source}/os/{config.OS_ISO_DIR}',
                     'is_repo': True,
                     'enabled': self.enabled
                 },
-                f'cortx_iso': {
-                    'source': (f'{self.source}/cortx_iso'),
+                f'{config.CORTX_ISO_DIR}': {
+                    'source': f'{self.source}/cortx_iso/'
```
andkononykhin2, line 1122: to ensure that concatenation will take please please add parenthesis here and in similar places

api/python/provisioner/srv/salt/repos/upgrade.sls
```diff
@@ -21,13 +21,15 @@
 
 {% for version, repos in pillar['release']['upgrade']['repos'].items() %}
 
-    {% if not repos.startswith(('http://', 'https://')) %}
-        {% set single_iso_repo = {version: repos.pop(version)} %}
+    {% set single_iso_repo = {version: repos.pop(version, None)} %}
 
-        # to mount the single ISO first
-        {{ setup_repos(single_iso_repo, base_dir) }}
+    { % if single_iso_repo % }
```
andkononykhin2, line 26: I guess it would be always True

andkononykhin2 (commented):
@dmitrydb Thanks for helping to finalize the PR. Please check my comments.

api/python/provisioner/srv/salt/repos/upgrade.sls
dmitrydb: Fixed

api/python/provisioner/inputs.py
dmitrydb: Fixed

api/python/provisioner/inputs.py
dmitrydb: Fixed

andkononykhin2 (approved):

PR #1111, opened 2021-04-13
EOS-18494: Salt module to validate open firewall ports

divya-vijay:
Done:
1. Renamed `lyveutil` salt module to `lyveutils` across the code. 
2. A new Salt validation module for verifying open firewall ports.


srv/_modules/lyveutils.py
```diff
@@ -0,0 +1,82 @@
+#
```
ypise, line 1: Did you use `git mv srv/_modules/lyveutil.py srv/_modules/lyveutils.py`?
It seems not which is why this is showing up as a new file here.

srv/_modules/lyveutils.py
```diff
@@ -0,0 +1,82 @@
...
+    for zone in fw_pillar:
+        for services, ports in fw_pillar[zone]["ports"].items():
+            validate_services.append(services)
+            for pt in ports:
+                port_num = int(''.join(filter(str.isdigit, pt)))
+                validate_ports.append(port_num)
+
+    for port, service in zip(validate_ports, validate_services):
+        tcp_port = __utils__['process.simple_process'](f"netstat -ltp | grep {port}")
+        tcp_svc = __utils__['process.simple_process'](f"netstat -lt | grep {service}")
```
ypise, line 70: Use https://github.com/Seagate/cortx-prvsnr/blob/99eef712cdab764f34ef8cd7f2622ee801fea037/api/python/provisioner/utils.py#L299

srv/_modules/lyveutils.py
```diff
@@ -0,0 +1,82 @@
...
+
+        udp_port = __utils__['process.simple_process'](f"netstat -lup | grep {port}")
```
ypise, line 72: Instead use https://github.com/Seagate/cortx-prvsnr/blob/99eef712cdab764f34ef8cd7f2622ee801fea037/api/python/provisioner/utils.py#L299

srv/_modules/lyveutils.py
```diff
@@ -0,0 +1,82 @@
...
+        udp_svc = __utils__['process.simple_process'](f"netstat -lu | grep {service}")
+
+    if (tcp_port or udp_port) and (tcp_svc or udp_svc):
+        logger.info("Success: Validation of open firewall ports")
+
+        return True
```
ypise, line 78: intermediate returns increases cyclometric complexity.
Set a flag variable and return the variable.

srv/_utils/process.py
```diff
@@ -1,4 +1,3 @@
-#
 # Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
ypise, line 1: Not necessary, instead use https://github.com/Seagate/cortx-prvsnr/blob/99eef712cdab764f34ef8cd7f2622ee801fea037/api/python/provisioner/utils.py#L299

ypise (dismissed):

srv/_modules/cluster.py
```diff
@@ -71,11 +70,7 @@ def storage_device_config():
                 time.sleep(_sleep_time)
 
                 logger.info(f"Command to populate multipath devices: {cmd}")
-                device_list = subprocess.Popen([cmd],
-                                        shell=True,
-                                        stdout=subprocess.PIPE,
-                                        stderr=subprocess.PIPE
-                                    ).stdout.read().decode("utf-8").splitlines()
+                device_list = __utils__['process.simple_process'](cmd)
```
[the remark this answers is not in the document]
andkononykhin2: @divya-vijay I agree with @ypise : we shouldn't go that path again and `run_subprocess_cmd` is likely generic and good enough to satisfy your need.

But it would bring some other questions:
 - how to import provisioner from within salt env: salt env sometimes may be tricky and may miss some paths in python search list.  That's why [this code](https://github.com/Seagate/cortx-prvsnr/blob/99eef712cdab764f34ef8cd7f2622ee801fea037/srv/_modules/prvsnr.py#L27-L50) was implemented.
 - how to call it, options:
   - call directly from our salt modules
   - call from inside __utils __:process (like you do) so it might be easier to consume by custom modules
   (I think the latter makes sense, it will also hide provisioner import routine)

So I suggest you to try the following options:

- keep new module in utils and put there:
  - provisioner import routine
  - wrapper around provisioner.utils.run_subprocess_cmd (like one that currently call popen)
- call the former from `prvsnr.py` module
- call the latter from all modules that require shell routine

andkononykhin2 (changes requested):
@divya-vijay The PR looks interesting (actually I wasn't aware about saltstack [utility extension modules](https://docs.saltproject.io/en/latest/topics/utils/index.html)). But it requires some more structuring. Please check my comments. Thank you

ypise (commented):
I am neutral on this as I won't be able to review this week.

srv/_modules/cluster.py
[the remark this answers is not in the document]
divya-vijay: @andkononykhin Thank you for this suggestion. This enhances the design better and makes it more structured, I agree. However, for now, since our scope is restricted to only shell commands, I've removed the `srv/_utils` and re-used the existing `run_subprocess_cmd` with some slight modifications to handle the same. And, as we move forward to face more utility functions in `srv/`, I've also added a snippet of this suggestion into the code for others' clarity.

srv/_modules/cluster.py
```diff
@@ -112,7 +107,7 @@ def jbod_storage_config():
     _metadata_field = "cluster/{0}/storage/metadata_devices".format(_target_node)
 
     _cmd = "multipath -ll | grep mpath | sort -k2.2 | awk '{ print $1 }'"
-    _device_list = subprocess.Popen([_cmd],shell=True,stdout=subprocess.PIPE).stdout.read().decode("utf-8").splitlines()
+    _device_list = __utils__['process.simple_process'](_cmd)
```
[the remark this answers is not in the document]
divya-vijay: Addressed.

srv/_modules/controller_cli.py
```diff
@@ -55,7 +54,7 @@ def fetch_enclosure_serial():
 
     logger.info("[ INFO ] Running controller-cli utility to get enclosure serial...")
     _cmd = f"sh {ctrl_cli_utility} host -h {host} -u {user} -p '{secret}' {_opt} | grep -A2 Serial | tail -1 > /etc/enclosure-id"
-    subprocess.Popen([_cmd],shell=True,stdout=subprocess.PIPE).stdout.read().decode("utf-8").splitlines()
+    __utils__['process.simple_process'](_cmd)
```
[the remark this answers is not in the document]
divya-vijay: Addressed.

srv/_modules/lyveutils.py
divya-vijay: @ypise Initially, it showed as renamed, but as I introduced new file in `_utils`, it showed as, 
`srv/_modules/lyveutil.py -> srv/_utils/process.py`
Hope its addressed in latest commit. Thanks

srv/_modules/lyveutils.py
divya-vijay: Done!

srv/_modules/lyveutils.py
divya-vijay: Addressed. Thanks

api/python/provisioner/utils.py
```diff
@@ -304,10 +304,10 @@ def run_subprocess_cmd(cmd, **kwargs):
     )
 
     _kwargs.update(kwargs)
-    _kwargs['check'] = True
 
     if type(cmd) is str:
         cmd = cmd.split()
+        _kwargs['check'] = True
```
divya-vijay, line 311: `check=True` did not fetch the right output for some shell commands earlier addressed with `Popen`, so moved it to other non-list commands.

srv/components/system/firewall/sanity_check.sls
```diff
@@ -102,7 +102,3 @@ Verify {{ network_interface }} ports:
 
 {% endfor %}
```
divya-vijay, line 104: Removed the step from `sanity_check`, as firewall validations command can be individually run as a manual step post-deployment.

api/python/provisioner/utils.py
andkononykhin: I don't think that we can relate command form (str or list) with check. Please don't mix that. If you need to be able to pass `check=False` then just make it as argument opposite to current force approach, e.g.:
```python
def run_subprocess_cmd(cmd, check=True, **kwargs):
    _kwargs = dict(
        universal_newlines=True,
        stdout=subprocess.PIPE,
        stder=subprocess.PIPE,
        check=check
    )
...
```

srv/_modules/cluster.py
```diff
@@ -48,18 +45,22 @@ def storage_device_config():
     ]
     for node in server_nodes:
         cluster_dict = provisioner.pillar_get(f"cluster/{node}/roles", targets=node)
+
         if "primary" in cluster_dict[node][f"cluster/{node}/roles"]:
-            cmd = "multipath -ll | grep prio=50 -B2|grep mpath|sort -k2.2 | awk '{ print $1 }'"
+            cmd = ("multipath -ll | grep prio=50 -B2 |"
```
andkononykhin, line 50: I am not sure that it can be passed to `run_subprocess_cmd` since it will split the string command into space-separated sub-parts. I suggest to make split logic optional, e.g. only if shell=False

srv/_modules/lyveutils.py
```diff
@@ -0,0 +1,106 @@
...
+# salt-call saltutil.clear_cache && salt-call saltutil.sync_modules
+# salt-call lyveutils.decrypt "component" "secret"
+
+import logging
+import os
+import sys
+
+# Update PYTHONPATH to include Provisioner API installed at:
+# /usr/local/lib/python3.6/site-packages/cortx-prvsnr-*
+sys.path.append(os.path.join(
```
andkononykhin, line 28: that logic is not good since it is based on hard-coded assumption regrading provisioner installation path and finally it's a copy-paste. How to improve:
- move that into some shared location (it is what I referred in my previous review) and re-use it from there for any module
- use [this](https://github.com/Seagate/cortx-prvsnr/blob/7ab6832cc2add72c9c2d386e185933f50f1a0d6e/srv/_modules/prvsnr.py#L27-L50) logic that does similar but without any hard-codes

andkononykhin (commented):
@divya-vijay To summarize. I see the following major things:
- `check=False` should be exposed as argument 
- command with `shell=True` should be always passed as a single item array so run_subprocess_cmd should be adjust here as well

Not critical for the current PR but:
-  not good that we have hard-coded python path and that approach is copy-pasted multiple times.
Please address at least the first block. For others point please create a follow up Jira.
Thank you

srv/_modules/cluster.py
divya-vijay: @andkononykhin Checked this - and it is as expected. Since we're not passing cmd as string, it doesn't split and takes as such.
`Command to execute: multipath -ll | grep prio=50 -B2 | grep mpath|sort -k2.2 | awk '{ print $1 }'`

api/python/provisioner/utils.py
divya-vijay: @andkononykhin no, what I meant was, when `check=True` and some shell commands return non-zero code, it raises exception instead of returning empty stdout.
Addressed now, thank you!

divya-vijay:
> @divya-vijay To summarize. I see the following major things:
> 
> * `check=False` should be exposed as argument
> * command with `shell=True` should be always passed as a single item array so run_subprocess_cmd should be adjust here as well
> 
> Not critical for the current PR but:
> 
> * not good that we have hard-coded python path and that approach is copy-pasted multiple times.
>   Please address at least the first block. For others point please create a follow up Jira.
>   Thank you


1. `check=False` is sent as an argument for required shell commands.
2. Already satisfied: all commands with `shell=True` are  passed as single array.
3. I will raise a new JIRA to update import design for PYTHONPATH, outside of api/ scope. currently, I can't seem to, as I am facing some access issue in creating new tickets.

Thank you!

andkononykhin (approved):

andkononykhin2 (approved):

PR #1118, opened 2021-04-15
[EOS-19230] API to configure mgmt network on node

mazinamdar:
1. API to configure mgmt network on the system.
2. Updated Set API to support local pillar data.
3. Removed comment from init.sls from network.mgmt.public as it is getting concatenated with dhcp string `- proto: dhcp`.


api/python/provisioner/params_spec.yaml
```diff
@@ -57,6 +57,13 @@ release:
   _path: release.sls
   target_build: release/target_build
 
+mgmt_network:
+  _path: cluster.sls
+  mgmt_public_ip: cluster/srvnode-0/network/mgmt/public_ip
```
andkononykhin2, line 62: @mazinamdar what `srvnode-0` means here?

andkononykhin2 (commented):
@mazinamdar Generally it looks clean. Thank you!

I have only one concern: `local` pillar mode was added as a kind of internal need when we worked on unboxing (or replacement, not sure) and GlusterFS infra could be not yet ready. So we needed that only for temporary usage and copied the data to shared pillar once GlusterFS becomes ready (@ypise please confirm).

So generally we shouldn't expose (consider that some one outside provisioner component) would use) it since it's our specific logic.

And the question is: what it the purpose of this new API?

mazinamdar:
@andkononykhin2 Actually we have exactly similar scenario, this is north bound API to configure only management network.
At the start each node will be independently configured for hostname, network , firewall and storage. There will not be any cluster and also gluster will be absent. We assumed that we have salt-master running on each node and each node will be considered as 'srvnode-0'. Clustering will be the next step here.

cc @ypise

api/python/provisioner/params_spec.yaml
mazinamdar: Initially each node will be considered as srvnode-0.

ypise (approved):

83bhp (approved):

PR #1120, opened 2021-04-15
EOS-19231: North Bound Command API Implementation - Data Network

SomwanshiAnjali:

### DIFF
--- a/api/python/provisioner/inputs.py
+++ b/api/python/provisioner/inputs.py
@@ -612,6 +612,9 @@ class NTP(ParamGroupInputBase):
 @attr.s(auto_attribs=True)
 class Hostname(ParamGroupInputBase):
     _param_group = 'hostname'
+    hostname: str = ParamGroupInputBase._attr_ib(
+        _param_group, descr="hostname to be set"
+    )
 
 
 @attr.s(auto_attribs=True)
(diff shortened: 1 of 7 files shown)


andkononykhin2 (commented):
@SomwanshiAnjali Looks good. Thank you. I guess the current PR depends on #1118. And we should wait until all concerns there are resolved. In particular [this one](https://github.com/Seagate/cortx-prvsnr/pull/1118#pullrequestreview-636681586).

api/python/provisioner/inputs.py
```diff
@@ -612,13 +612,42 @@ class NTP(ParamGroupInputBase):
...
 
 
+@attr.s(auto_attribs=True)
+class DataNetwork(ParamGroupInputBase):
+    _param_group = 'data_network'
+    data_public_ip: str = ParamGroupInputBase._attr_ib(
+        _param_group, descr="node public data interface IP",
+        validator=Validation.check_ip4
+    )
+    data_gateway: str = ParamGroupInputBase._attr_ib(
```
ypise, line 632: This is optional parameter. Would validation fail if it's blank?

ypise (approved):

api/python/provisioner/inputs.py
SomwanshiAnjali: Yes, It will skip validation if value is missing
https://github.com/Seagate/cortx-prvsnr/blob/aeb708e705abd86731b69c2f1898a7a340eafefb/api/python/provisioner/inputs.py#L576

PR #1131, opened 2021-04-19
EOS-19490: Fix - hostname reset on reboot

SomwanshiAnjali:
hostname was getting reset to old one on reboot due to "operation not permitted" error
fix:
1. Remove `i` attribute from /etc/hostname 
2. Use hostnamectl command to set the hostname

### DIFF
--- a/srv/components/system/config/set_hostname.sls
+++ b/srv/components/system/config/set_hostname.sls
@@ -15,12 +15,21 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-Update hostname file permissions:
+Make hostname file modifiable:
   cmd.run:
     - name: "chattr -i /etc/hostname"
 
 Set hostname:
-  cmd.run:
-    - name: "hostnamectl set-hostname {{ pillar['cluster'][grains['id']]['hostname'] }}"
+  network.system:
+    - name: set_hostname
+    - hostname: {{ pillar['cluster'][grains['id']]['hostname'] }}
+    - apply_hostname: True
+    - retain_settings: True
     - require:
-      - Update hostname file permissions
+      - Make hostname file modifiable
+
+Restore hostname file protection:
+  cmd.run:
+    - name: "chattr +i /etc/hostname"
+    - require:
+      - Make hostname file modifiable
--- a/srv/components/system/config/set_hostname.sls
+++ b/srv/components/system/config/set_hostname.sls
@@ -15,9 +15,12 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
+Update hostname file permissions:
+  cmd.run:
+    - name: "chattr -i /etc/hostname"
+
 Set hostname:
-  network.system:
-    - name: set_hostname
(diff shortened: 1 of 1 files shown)


srv/components/system/config/set_hostname.sls
```diff
@@ -15,9 +15,12 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
+Update hostname file permissions:
+  cmd.run:
+    - name: "chattr -i /etc/hostname"
+
 Set hostname:
-  network.system:
```
83bhp, line 19: after changing the attr can we still use network.system way to update hostname? Or is there any other salt standard module to do the same?

83bhp (approved):

srv/components/system/config/set_hostname.sls
```diff
@@ -15,9 +15,12 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
+Update hostname file permissions:
```
andkononykhin, line 18: could you make that more specific, e.g. `Make hostname file modifiable`

srv/components/system/config/set_hostname.sls
```diff
@@ -15,9 +15,12 @@
...
+  cmd.run:
+    - name: "chattr -i /etc/hostname"
```
andkononykhin, line 20: shouldn't we restore that protection once hostname is updated?

andkononykhin (commented):
@SomwanshiAnjali Please check y suggestion. Thanks

srv/components/system/config/set_hostname.sls
```diff
@@ -15,9 +15,12 @@
...
+
 Set hostname:
-  network.system:
-    - name: set_hostname
-    - hostname: {{ pillar['cluster'][grains['id']]['hostname'] }}
-    - apply_hostname: True
-    - retain_settings: True
+  cmd.run:
+    - name: "hostnamectl set-hostname {{ pillar['cluster'][grains['id']]['hostname'] }}"
```
ypise, line 24: Prefer converting this to a custom module.
Then handle the `chattr` driven immutability inside the custom set_hostname module.

ypise (changes requested):

srv/components/system/config/set_hostname.sls
SomwanshiAnjali: @83bhp 
yes we can use `network.system`, verified and updated PR accordingly

srv/components/system/config/set_hostname.sls
SomwanshiAnjali: Updated

srv/components/system/config/set_hostname.sls
SomwanshiAnjali: Updated

srv/components/system/config/set_hostname.sls
SomwanshiAnjali: Used `network.system` module instead of set-hostname command

ypise (approved):
Agreeing to converting the logic to a custom module in separate PR.

PR #1132, opened 2021-04-19
EOS-19532: Correct the Kafka configuration

83bhp:
1. Update listener in server.properties
2. Add delay in start and stop of Kafka services


mazinamdar (approved):
looks good to me just minor suggestion

srv/components/cortx_utils/init.sls
```diff
@@ -19,4 +19,6 @@ include:
   - .install
   - .post_install
   - .config
+{% if grains['id'] == 'srvnode-1' %}
```
mazinamdar, line 22: will checking primary node will help here ? 

`{% if "primary" in pillar["cluster"][grains["id"]]["roles"] %}`

83bhp: The suggestion from Foundation was to run it on any node, not just primary.. so kept srvnode-1 at present, will have to find a way to run it on all nodes one after another later on.

PR #1143, opened 2021-04-20
[EOS-18240] adds test api helpers and examples

andkononykhin2:


test/testapi/integration.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def __attrs_post_init__(self):
+        # TODO more smarter logic to get interface that is asseccible from host
+        # (relates to https://github.com/hashicorp/vagrant/issues/2779)
+        if self.interface is None:
+            if (
+                isinstance(self.remote, h.VagrantMachine) and
+                (self.remote.provider == 'vbox')
+            ):
+                self.interface = 'enp0s8'
```
dmitrydb, line 64: Do I understand correctly that `enp0s8` is the constant interface name for IPv4 settings that are needed for VirtualBox configuration?

test/testapi/integration.py
```diff
@@ -0,0 +1,224 @@
...
+    @property
+    def ssh_host(self):
+        return self.host.interface(self.interface).addresses[0]
+
+    @property
+    def tmpdir(self):
+        if self._tmpdir is None:
+            tmpdir_function = self.request.getfixturevalue('tmpdir_function')
+            # TODO non linux systems
+            self._tmpdir = Path('/tmp') / tmpdir_function.relative_to('/')
```
dmitrydb, line 85: Do we need to use here `tempfile` python module?

dmitrydb (approved):
It looks good for me. I just left few comments here

test/testapi/integration.py
andkononykhin2: yes

test/testapi/integration.py
andkononykhin2: we can, but it would make the final path complicated. That path actually is created inside temporary docker containers but constructed being on the host. So that place requires more thinking.

PR #1150, opened 2021-04-21
[EOS-19542] improve ci pipelines

andkononykhin2:


dmitrydb (approved):
It looks good to me. I approve this PR. Only one question. Do we assume that the docker image will support `apt` package manager?

83bhp (approved):

ypise (approved):

devops/ci/Dockerfile.integration.tmpl
```diff
@@ -0,0 +1,21 @@
+FROM {parent}
+
+USER root
+
+# docker
+COPY setup_docker.sh /tmp/
+RUN bash /tmp/setup_docker.sh \
```
ypise, line 7: Can all the run statements be clubbed to call a utility script?
Would reduce number of intermittent images.

devops/ci/Jenkinsfile.ci.integration
```diff
@@ -55,15 +57,32 @@ def buildUnitDocker() {
...
+        dockerfile,
+        "devops/jenkins/cortx_jenkins/agent",
+        " --build-arg docker_gid=$docker_gid"
     )
 }
 
 def cleanDocker(){
-    sh("docker rmi -f ${dockerImage} || true")
+    sh("docker rmi -f ${dockerImageUnitName} || true")
+    sh("docker rmi -f ${dockerImageIntegrationName} || true")
```
ypise, line 85: Let's push the images to GitHub registry.
@shailesh-vaidya could you help with registry details?

andkononykhin2:
> It looks good to me. I approve this PR. Only one question. Do we assume that the docker image will support `apt` package manager?

We have set of different images which might be grouped into:
- utility ones - will just prepare an environment (python) and run tests - run env
- test ones - these are created by test scenarios themselves and currently it's a target platform - centos 7.8.2003.

Generally we care mostly about the second group since for them we need to use target systems. No limitation for first group so far.

devops/ci/Dockerfile.integration.tmpl
andkononykhin2: It's not always a bad idea actually. If we combine docker setup with other pkgs installation then docker would be re-installed each minor change in not related stuff. Let's keep that discussion separately.

PR #1177, opened 2021-04-27
EOS-19773: adding pillar_encrypt as part of bootstrap

divya-vijay:

### DIFF
--- a/api/python/provisioner/commands/setup_provisioner.py
+++ b/api/python/provisioner/commands/setup_provisioner.py
@@ -1757,8 +1757,11 @@ class SetupProvisioner(SetupCmdBase, CommandParserFillerMixin):
                     "{\"inline\": {\"no_encrypt\": True}}"
                 )
 
-        logger.info("Refresh enclosure id on the system")
+        logger.info(
+             "Encrypt pillar values and Refresh enclosure id on the system"
+        )
         for state in [
+            'components.system.config.pillar_encrypt',
             'components.system.storage.enclosure_id',
             'components.system.config.sync_salt'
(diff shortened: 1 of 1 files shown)


ypise (approved):

ypise (approved):

api/python/provisioner/commands/setup_provisioner.py
```diff
@@ -1757,8 +1757,11 @@ def _run(self, nodes, **kwargs):  # noqa: C901 FIXME
                     "{\"inline\": {\"no_encrypt\": True}}"
                 )
 
-        logger.info("Refresh enclosure id on the system")
+        logger.info(
+             "Encrypt pillar values and Refresh enclosure id on the system"
+        )
         for state in [
+            'components.system.config.pillar_encrypt',
```
83bhp, line 1764: How does just encrypting the pillar solves this issue?

divya-vijay: I should have mentioned, my bad!
In `pillar_encrypt`, we install cortx module, which was expected for utils (used by `lyveutils` in `controller_cli`). Since that was missing, `enclosure-id` file was not getting written with the expected value and iterating over 0.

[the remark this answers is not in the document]
divya-vijay: Sure, yeah, I re-ran bootstrap in HW and VM setup with this in place and it worked without throwing any tantrum, so I thought for now, it's not breaking. But only with end-to-end deployment, can we be 100% sure, and we're on it now.

PR #1184, opened 2021-04-28
EOS-19760: fix for OVA bootstrap script issue

rahuljyoti65:
Bootstrap script execution was not proceeding further and was stuck at "Configuring haproxy and s3" stage. This was because kafka was not running.
Fix:
1. starting kafka before configuring s3server.
2. increased the delay to 30s after starting zookeeper. ( bcoz on executing kafta.start using salt only zookeeper was starting and not kafka)

### DIFF
--- a/cli/src/virtual_appliance/bootstrap.sh
+++ b/cli/src/virtual_appliance/bootstrap.sh
@@ -40,8 +40,13 @@ salt-call state.apply components.system.network.mgmt.public
 salt-call state.apply components.system.network.data.public
 salt-call state.apply components.system.network.data.direct
 
+salt-call state.apply components.system.config.sync_salt
+
 #reconfigure /etc/hosts
 salt-call state.apply components.system.config.hosts
+
+#reconfigure kafka
+salt-call state.apply components.misc_pkgs.kafka.start 
 
 #reconfigure firewall
 salt-call state.apply components.system.firewall.teardown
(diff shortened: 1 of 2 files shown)


cli/src/virtual_appliance/bootstrap.sh
```diff
@@ -64,6 +66,9 @@ provisioner pillar_set s3clients/s3server/ip \"${DATA_IP}\"
 salt "*" state.apply components.s3clients.config | tee -a "${LOG_FILE}"
 provisioner confstore_export
 
+#reconfigure hare CDF
+salt-call state.apply components.hare.config.prepare
```
ypise, line 70: This should not be necessary on R2.
Check with hare team, which interface they define configures the CDF file.

srv/components/misc_pkgs/kafka/start.sls
```diff
@@ -30,7 +30,7 @@ Start zoopkeper:
 Wait for zookeeper to start:
```
ypise, line 30: The sanity_check takes care of this, thus this wait should be removed.

ypise (changes requested):

rahuljyoti65:
Clsoing this PR as the issue is addressed in another PR https://github.com/Seagate/cortx-prvsnr/pull/1190

PR #1185, opened 2021-04-28
Fixes for firewall

ypise:

### DIFF
--- a/pillar/components/firewall.sls
+++ b/pillar/components/firewall.sls
@@ -10,9 +10,9 @@ firewall:
     services:
       - ssh
       - ftp
-      {%- if salt['cmd.run']('rpm -qa glusterfs-server') %}
+      {#%- if salt['cmd.run']('rpm -qa glusterfs-server') %#}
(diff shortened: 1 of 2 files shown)


pillar/components/firewall.sls
```diff
@@ -10,9 +10,9 @@ firewall:
     services:
       - ssh
       - ftp
-      {%- if salt['cmd.run']('rpm -qa glusterfs-server') %}
+      {#%- if salt['cmd.run']('rpm -qa glusterfs-server') %#}
```
ypise, line 13: This is probably a reason for the cache info output during destroy.

83bhp: will this skip the glusterfs teardown?

ypise: It won't skip teardown. This is just for firewall port for GlusterFS

ypise:
This change is unnecessary.

ypise:
This change is unnecessary.

PR #1204, opened 2021-05-04
EOS-15754: Automated testing for R2 upgrade release bundle setup API

dmitrydb:
I added tests for SW upgrade validators API


ypise (approved):

andkononykhin2 (approved):
@dmitrydb Approved. But please mark these tests as unit ones (`@pytest.mark.unit`) so they start to be collected as unit. And create the follow-up task: need to separate all the inner test cases (e.g. all `with` blocks) into special tests. Thanks

dmitrydb:
@andkononykhin2 Thank you for the reference. I found a place in our `helper.py` where we setup this 'unit'-marker. But there is an interested thing here: `pytest -m unit --collect-only` outputs my new test cases even without `@pytest.mark.unit`. Also, I have one suggestion here: my new test is only one place now where we use `@pytest.mark.unit`. Do we need to update other tests accordingly?

PR #1217, opened 2021-05-05
Fix minor issues in deploy and destroy.

83bhp:
1. Move the dead code of validation check for consul before running
   cortx-ha component.
2. Fix typo in destroy - change rabbitmq to rabbitmq.teardown

### DIFF
--- a/api/python/provisioner/commands/deploy.py
+++ b/api/python/provisioner/commands/deploy.py
@@ -313,6 +313,9 @@ class Deploy(CommandParserFillerMixin):
                     self._apply_state(
                         f"components.{state}", secondaries, stages
                     )
+                else:
+                     # Execute on all targets
+                    self._apply_state(f"components.{state}", targets, stages)
 
     def _update_salt(self, targets=config.ALL_MINIONS):
         # TODO IMPROVE why do we need that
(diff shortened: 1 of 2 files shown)


api/python/provisioner/commands/deploy.py
```diff
@@ -307,14 +307,12 @@ def _run_states(self, states_group: str, run_args: run_args_type):
                     "ha.cortx-ha"
                 ):
                     # Execute first on primary then on secondaries.
+                    if state == "ha.cortx-ha":
+                        self.ensure_consul_running()
                     self._apply_state(f"components.{state}", primary, stages)
                     self._apply_state(
                         f"components.{state}", secondaries, stages
                     )
-                else:
```
ypise, line 314: We need else.
If it doesn't get filtered in any conditionals above, it still has to be executed for generic case.

ypise (changes requested):

api/python/provisioner/commands/deploy.py
83bhp: Ok.

83bhp: addressed

ypise (approved):

PR #1221, opened 2021-05-06
[EOS-20486] Create copy of id_rsa_prvsnr as id_rsa 

mazinamdar:

### DIFF
--- a/api/python/provisioner/srv/salt/ssh/config.sls
+++ b/api/python/provisioner/srv/salt/ssh/config.sls
@@ -41,6 +41,8 @@ ssh_pub_key_deployed:
     - requires:
       - ssh_dir_created
 
+# Special case for cortx-ha: paramiko by default use id_rsa key
+
 copy_ssh_priv_key_deployed:
   file.managed:
     - name: /root/.ssh/id_rsa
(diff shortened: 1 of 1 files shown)


ypise (approved):

api/python/provisioner/srv/salt/ssh/config.sls
```diff
@@ -41,6 +41,25 @@ ssh_pub_key_deployed:
     - requires:
       - ssh_dir_created
 
+copy_ssh_priv_key_deployed:
+  file.managed:
+    - name: /root/.ssh/id_rsa
```
andkononykhin2, line 46: @mazinamdar please consider to optimize that to reduce copy-paste. Please check an [example](https://docs.saltproject.io/en/latest/ref/states/all/salt.states.file.html#operations-on-regular-files-special-files-directories-and-symlinks) in the official docs.
>The names parameter, which is part of the state compiler, can be used to expand the contents of a single state declaration into multiple, single state declarations. Each item in the names list receives its own individual state name and is converted into its own low-data structure. This is a convenient way to manage several files with similar attributes.

andkononykhin2 (changes requested):

83bhp (approved):

api/python/provisioner/srv/salt/ssh/config.sls
mazinamdar: comment is addressed

andkononykhin2 (approved):

api/python/provisioner/srv/salt/ssh/config.sls
andkononykhin2: Thank you

PR #1233, opened 2021-05-11
EOS-20503: Disable 2GB swap on LR R1 SMS2

SomwanshiAnjali:
Signed-off-by: Anjali Somwanshi <anjali.somwanshi@seagate.com>

PR #1236, opened 2021-05-11
SMS-2 fixes

SomwanshiAnjali:
Addressed:
1. EOS-20519: Open glusterfs ports for hw
2. EOS-20503: Disable 2GB swap on LR R1 SMS2

### DIFF
--- a/srv/components/system/storage/teardown/init.sls
+++ b/srv/components/system/storage/teardown/init.sls
@@ -22,8 +22,13 @@ include:
 {% endif %}
 
 Mount default OS swap device:
-  cmd.run:
-    - name: swapon -a || true
+  module.run:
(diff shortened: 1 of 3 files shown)


ypise:
Shall be delivered through individual PRs:
- #1233 
- #1234

srv/components/system/storage/teardown/init.sls
```diff
@@ -22,8 +22,13 @@ include:
 {% endif %}
 
 Mount default OS swap device:
-  cmd.run:
-    - name: swapon -a || true
+  module.run:
+    - mount.swapon:
+      - name: /dev/dm-1
```
83bhp, line 27: like on the other PR, we need to remove this as well, right?

PR #1237, opened 2021-05-11
EOS-19773: Do not encrypt controller passwd for HW

divya-vijay:

### DIFF
--- a/srv/components/controller/files/scripts/controller-cli.sh
+++ b/srv/components/controller/files/scripts/controller-cli.sh
@@ -523,8 +523,8 @@ main()
     if [[ "$update_fw" = true || "$restart_ctrl_opt" = true
         || "$shutdown_ctrl_opt" = true || "$ntp_opt" = true
         || "$show_license" = true ]]; then
-        echo "main(): Decrypting the password received from api" >> $logfile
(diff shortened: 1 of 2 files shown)


srv/components/controller/files/scripts/controller-cli.sh
```diff
@@ -523,8 +523,8 @@ main()
     if [[ "$update_fw" = true || "$restart_ctrl_opt" = true
         || "$shutdown_ctrl_opt" = true || "$ntp_opt" = true
         || "$show_license" = true ]]; then
-        echo "main(): Decrypting the password received from api" >> $logfile
-        pass=`salt-call lyveutils.decrypt storage ${pass} --output=newline_values_only`
+        #echo "main(): Decrypting the password received from api" >> $logfile
+        #pass=`salt-call lyveutils.decrypt storage ${pass} --output=newline_values_only`
```
ypise, line 527: Don't change it here.
If possible, we should find a way to make direct calls to the targeted method instead of commenting this here.

ypise (changes requested):

divya-vijay:
@rahuljyoti65 I believe you're discussing with SSPL team to fetch `enclosure_id` using their REST API, in which case, we can bypass this step from current design. 
@ypise Since this PR was a temporary fix, closing it now.

PR #1243, opened 2021-05-12
[cortx-ha-sequence] run cortx-ha sequentially on secondaries

mazinamdar:

Retrieve node list from pillar and run cortx-ha sequentially on all secondary nodes.

### DIFF
--- a/api/python/provisioner/commands/deploy.py
+++ b/api/python/provisioner/commands/deploy.py
@@ -21,6 +21,7 @@ import logging
 from .. import (
     config,
     inputs,
+    values,
     errors
 )
 from ..salt import (
@@ -42,6 +43,10 @@ from .setup_provisioner import SetupCtx
 from .configure_setup import (
     RunArgsConfigureSetupAttrs,
     SetupType
+)
+from ..pillar import (
+    PillarKey,
+    PillarResolver
 )
 
 
@@ -168,6 +173,22 @@ class Deploy(CommandParserFillerMixin):
     _run_args_type = run_args_type
     setup_ctx: Optional[SetupCtx] = None
 
+    @staticmethod
+    def _get_node_list():
+        """Retrieve pillar data."""
+        pillar_key = PillarKey("cluster")
+        pillar = PillarResolver(config.LOCAL_MINION).get([pillar_key])
+        pillar = next(iter(pillar.values()))
+        if not pillar[pillar_key] or pillar[pillar_key] is values.MISSED:
+            raise ValueError("value is not specified for cluster")
+        else:
+            cluster = pillar[pillar_key]
+            nodes = []
+            for key in cluster:
+                if 'srvnode' in key:
+                    nodes.append(key)
+            return nodes
+
     def _primary_id(self):
         if self.setup_ctx:
             return self.setup_ctx.run_args.primary.minion_id
@@ -254,7 +275,7 @@ class Deploy(CommandParserFillerMixin):
                 else:
                     logger.warning(f"State {_state} is missed, ignored")
 
-    def _run_states(self, states_group: str, run_args: run_args_type):
+    def _run_states(self, states_group: str, run_args: run_args_type):  # noqa: C901, E501
         # FIXME VERIFY EOS-12076 Mindfulness breaks in legacy version
         setup_type = (
             SetupType.SINGLE
@@ -270,9 +291,6 @@ class Deploy(CommandParserFillerMixin):
 
         primary = self._primary_id()
         secondaries = f"not {primary}"
-        # Temperory fix for EOS-20526
-        second_node = "srvnode-2"
-        third_node = "srvnode-3"
 
         hw_states = [
             "system.storage.multipath",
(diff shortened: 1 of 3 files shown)


api/python/provisioner/commands/deploy_vm.py
```diff
@@ -144,12 +158,16 @@ def _run_states(self, states_group: str, run_args: run_args_type):
...
                     )
-                elif state in ("ha.cortx-ha"):
-                    #Execute firt on primary, then on second node and then on the third
+                elif state in (
+                    "ha.cortx-ha"
+                ):
+                    # Execute first on primary then on secondary sequentially.
+                    nodes = DeployVM._get_node_list()
                     self._apply_state(f"components.{state}", primary, stages)
-                    self._apply_state(f"components.{state}", second_node, stages)
```
andkononykhin2, line 150: Don't you think that we need similar changes in https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/api/python/provisioner/commands/deploy.py?

andkononykhin2 (changes requested):

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -144,12 +158,16 @@ def _run_states(self, states_group: str, run_args: run_args_type):
                     self._apply_state(
                         f"components.{state}", secondaries, stages
                     )
-                elif state in ("ha.cortx-ha"):
-                    #Execute firt on primary, then on second node and then on the third
+                elif state in (
+                    "ha.cortx-ha"
```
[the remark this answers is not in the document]
mazinamdar: added comma, comment is addressed

api/python/provisioner/commands/deploy_vm.py
mazinamdar: comment is addressed

api/python/provisioner/commands/deploy.py
```diff
@@ -312,14 +330,18 @@ def _run_states(self, states_group: str, run_args: run_args_type):
                     self._apply_state(
                         f"components.{state}", secondaries, stages
                     )
-                elif state in ("ha.cortx-ha"):
-                    # Execute first on primary then on secondaries in a sequence.
-                    self.ensure_consul_running()
```
andkononykhin2, line 317: why that was removed?

api/python/provisioner/commands/deploy_dual.py
```diff
@@ -189,10 +189,18 @@ def _run_states(self, states_group: str, run_args: run_args_type):
                     self._apply_state(
                         f"components.{state}", secondaries, stages
                     )
+                elif state in (
+                    "ha.cortx-ha",
```
andkononykhin2, line 193: that code won't be run since the same condition is presented above, please check line 167

api/python/provisioner/commands/deploy_dual.py
```diff
@@ -189,10 +189,18 @@ def _run_states(self, states_group: str, run_args: run_args_type):
...
+                    # Execute first on primary then on secondary sequentially.
+                    nodes = Deploy._get_node_list()
+                    self._apply_state(f"components.{state}", primary, stages)
+                    if primary in nodes:
+                        nodes.remove(primary)
+                    for node in nodes:
+                        self._apply_state(f"components.{state}", node, stages)
                 else:
                     self._apply_state(f"components.{state}", targets, stages)
-                    if state == "ha.cortx-ha":
```
andkononykhin2, line 194: not sure how that condition can be True but please double-check if we need to ensure consul in scope of cortx-ha setup

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -144,12 +136,16 @@ def _run_states(self, states_group: str, run_args: run_args_type):
                     self._apply_state(
                         f"components.{state}", secondaries, stages
                     )
-                elif state in ("ha.cortx-ha"):
-                    #Execute firt on primary, then on second node and then on the third
+                elif state in (
+                    "ha.cortx-ha",
+                ):
+                    # Execute first on primary then on secondary sequentially.
+                    nodes = Deploy._get_node_list()
```
andkononykhin2, line 143: shouldn't we worry about running consul here as well?

andkononykhin2 (changes requested):

api/python/provisioner/commands/deploy.py
mazinamdar: `check_consul_running` was specially written for dual node and It has old consul service names now consul service names are changed. That's why I have removed. If we required this check we can add this check in separate PR.

api/python/provisioner/commands/deploy_dual.py
mazinamdar: comment is addressed

api/python/provisioner/commands/deploy_dual.py
mazinamdar: `check_consul_running` was specially written for dual node and It has old consul service names now consul service names are changed. That's why I have removed. If we required this check we can add this check in separate PR

api/python/provisioner/commands/deploy_vm.py
mazinamdar: same as above comment, if required we can have separate PR for consul check

andkononykhin2 (approved):

api/python/provisioner/commands/deploy_dual.py
```diff
@@ -189,10 +188,18 @@ def _run_states(self, states_group: str, run_args: run_args_type):
...
                         f"components.{state}", secondaries, stages
                     )
+                elif state in (
+                    "ha.cortx-ha",
+                ):
+                    # Execute first on primary then on secondary sequentially.
+                    nodes = Deploy._get_node_list()
+                    self._apply_state(f"components.{state}", primary, stages)
+                    if primary in nodes:
+                        nodes.remove(primary)
```
ypise, line 198: if condition might not be necessary here.
Primary would always be present, even in case of a single node. So this condition shall always be true.

api/python/provisioner/commands/deploy_vm.py
```diff
@@ -144,12 +136,16 @@ def _run_states(self, states_group: str, run_args: run_args_type):
...
                     self._apply_state(f"components.{state}", primary, stages)
-                    self._apply_state(f"components.{state}", second_node, stages)
-                    self._apply_state(f"components.{state}", third_node, stages)
-
+                    if primary in nodes:
+                        nodes.remove(primary)
```
ypise, line 146: Same here, this is unnecessary.

ypise (commented):
@mazinamdar 
Few comments, but not a reason to block the merge.

ypise (approved):

PR #1249, opened 2021-05-14
EOS-20430: ISO query API - get CORTX SW upgrade version info

dmitrydb:
Commit the base version of Get SW Upgrade Info command


api/python/provisioner/commands/upgrade/get_swupgrade_info.py
```diff
@@ -0,0 +1,142 @@
...
+
+            upgrade_releases = pillars[local_minion].get("repos", None)
+
+            if not upgrade_releases:
+                raise ValueError("release/upgrade/repos pillars are missed")
+
+            upgrade_releases = list(upgrade_releases.keys())
+            upgrade_releases.remove(REPO_CANDIDATE_NAME)
+
+            release = max(upgrade_releases, key=version.parse)
```
andkononykhin, line 140: @dmitrydb Cool! But would work only in case Cortx release versions satisfy PEP-440 which is likely but who knows... Could you make a note here? Thanks

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
```diff
@@ -0,0 +1,142 @@
...
+        # NOTE: Format is following
+        # ```
+        #  {
+        #      'cortx-motr': {
+        #             'version': '2.0.0-277',
+        #          },
+        #  }
+        # ```
+        #
+        # TODO: Along the with 'version', field we need to add
```
andkononykhin, line 98: please add a reference to Jira task here

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
```diff
@@ -0,0 +1,142 @@
...
+        dict:
+            return dictionary with CORTX packages and their versions
+
+        """
+        local_minion = local_minion_id()
+
+        if release is None:
+            # NOTE: take the latest release from SW upgrade repositories
+
+            # TODO: make get pillar API public and available for others to
```
andkononykhin, line 128: please consider [that](https://github.com/Seagate/cortx-prvsnr/blob/affa29980a3a5c5dc0e58d4b037de0d5b5f1595e/api/python/provisioner/commands/upgrade/sw_upgrade_node.py#L56) option, it should be sufficient. Also you may consider to fetch `release/upgrade/repos` key-path directly so you won't need to check  it against `None`.

andkononykhin (commented):
@dmitrydb Looks clean! Please check my minor comments though. Thank you

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
dmitrydb: I added a note here about our assumption regarding release version formatting. But I have some doubts that the current format is valid according to PEP-440. Fortunately `packaging` parses it correctly:
```python
from packaging import version

a = version.parse('1.1.1-207')
print(a)
print(a.post)
```
And output
```
1.1.1.post207
207
```

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
dmitrydb: Done

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
dmitrydb: I updated code to use suggest approach. But anyway we need to have a convenient API to get pillar values without constructing long dict keys.

andkononykhin2: Agree. Thanks

andkononykhin2 (approved):

PR #1251, opened 2021-05-15
EOS-20595: Deployment failing while applying utils component

ypise:

### DIFF
--- a/srv/components/misc_pkgs/kafka/teardown.sls
+++ b/srv/components/misc_pkgs/kafka/teardown.sls
@@ -35,6 +35,14 @@ Remove zookeeper log directory:
   file.absent:
     - name: /var/log/zookeeper
 
+Remove accepted epoch:
+  file.absent:
+    - name: /var/lib/zookeeper/version-2/acceptedEpoch
+
+Remove current epoch:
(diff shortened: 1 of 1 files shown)


srv/components/misc_pkgs/kafka/teardown.sls
```diff
@@ -35,6 +35,14 @@ Remove zookeeper log directory:
   file.absent:
     - name: /var/log/zookeeper
 
+Remove accepted epoch:
```
sachinpunadikar, line 38: Better cleanup entire /var/lib/zookeeper

sachinpunadikar (changes requested):
Cleanup the entire directory as mentioned in comment

srv/components/misc_pkgs/kafka/teardown.sls
ypise: @sachinpunadikar 
We are doing that already on line 32 above.
Good catch. We don't need this change. There's something else we need to check.

ypise:
These changes have already been covered in the teardown.sls file line 32.
There is no need to deliver this PR.

PR #1267, opened 2021-05-16
EOS-20415: Setup 3 node K8s cluster

ypise:
Kubernetes setup using saltstack formulae.

### DIFF
--- a/srv/k8s/containerd/config.sls
+++ b/srv/k8s/containerd/config.sls
@@ -1,0 +1,33 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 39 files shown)


andkononykhin2 (commented):
@ypise Looks good and clean. Just few suggestions:
- docker setup has been already covered previously (when we checked option of Gluster servers inside docker) using 3rd party [salt-formula](https://github.com/saltstack-formulas/docker-formula). Makes sense to consider that option since it's portable and I believe enough tested by the community:
  - in our code it was used as [salt gitfs remote](https://github.com/Seagate/cortx-prvsnr/blob/7b1617d019ffad42a4679d3c530926649ee64e17/srv/components/provisioner/salt_master/files/master#L814-L815) and is disabled currently.
  - an alternative option (in case of offline setup) there is [a documented option](https://docs.saltproject.io/en/latest/topics/development/conventions/formulas.html#adding-a-formula-directory-manually) how to use that offline
  - [licensing](https://github.com/saltstack-formulas/docker-formula/blob/master/LICENSE) looks fine for both options (Apache 2.0)
- could you please add some references to the k8s setup docs which I guess you used for k8s configuration automation?

Thank you

ypise:
I've submitted this PR with the code as a reference.
This is an isolated code and would not affect normal deployments.

PR #1283, opened 2021-05-21
EOS-19898: LR Signature module integrated with CLI

divya-vijay:

-----
[View rendered docs/design_updates.md](https://github.com/Seagate/cortx-prvsnr/blob/lr-node-stamping-cli/docs/design_updates.md)


ypise (approved):

lr-cli/cortx_setup/commands/signature/get.py
```diff
@@ -0,0 +1,76 @@
...
+
+        Execution:
+        `cortx_setup signature get --key 'key'`
+
+        """
+        try:
+            lr_sign_file = Path("/etc/node-signature.yaml")
+            index = "signature"
+
+            if not (lr_sign_file.exists() and
```
ypise, line 52: Raise an exception under this if,
Remove the else block.
This would reduce the complexity.

lr-cli/cortx_setup/commands/signature/get.py
```diff
@@ -0,0 +1,76 @@
...
+                    lr_sign_file.stat().st_size != 0):
+
+                self.logger.warning("Node Signature was never set in database. "
+                               "Set it first with `signature set` command.")
+                lr_sign = "No data found. Execute `signature set` command."
```
ypise, line 57: Where is this assignment used?

lr-cli/cortx_setup/commands/signature/get.py
```diff
@@ -0,0 +1,76 @@
...
+
+            else:
+                Conf.load(index, f'yaml://{lr_sign_file}')
+                lr_sign = Conf.get(index, f'signature>{key}')
+
+                if not lr_sign:
```
ypise, line 63: Since we have decided flat key-> value structure, it would make sense to just raise an exception under this if.
Second level check would be unnecessary.

lr-cli/cortx_setup/commands/signature/get.py
```diff
@@ -0,0 +1,76 @@
...
+                    self.logger.warning(
+                       "Given key '%s' is possibly "
+                       "not found in Signature data." % key
+                    )
+                    lr_sign = Conf.get(index, 'signature')
+
+            self.logger.info(f"LR Signature: {lr_sign}")
```
ypise, line 70: Let this be debug as no one would check the console, but the signature file for validation.

lr-cli/cortx_setup/commands/signature/set.py
```diff
@@ -0,0 +1,97 @@
...
+                user_msg = ("A valid signature is mandatory to set for stamping. "
+                    "Expected Signature format: \"--key 'key' --value 'value' \"."
+                )
+                self.logger.error(user_msg)
+                raise Exception(user_msg)
+
+            if not (lr_sign_file.exists() and
+                lr_sign_file.stat().st_size != 0
+            ):
+                self.logger.warning("Node Signature was never set "
```
ypise, line 70: Same comment as get.

lr-cli/cortx_setup/commands/signature/set.py
```diff
@@ -0,0 +1,97 @@
...
+            Conf.save(index)
+
+            # TODO: With refined approach, check if this signature data
+            # should be written to grains. If yes, what format?
+
+            # Make the file immutable for further editing
+            _cmd = f"chattr +i {lr_sign_file} && lsattr {lr_sign_file}"
+            subprocess.Popen(_cmd, shell=True)     # nosec
+
+            self.logger.info("LR-Signature Stamping done for current node")
```
ypise, line 90: This should be a debug log.

lr-cli/cortx_setup/commands/signature/get.py
divya-vijay: In a fresh setup or when run before signature was set, currently this would return the user msg to the user. Should we raise exception and stop the execution in such case?

lr-cli/cortx_setup/commands/signature/get.py
divya-vijay: In return statement. If a signature value is not found, it returns this msg.

lr-cli/cortx_setup/commands/signature/set.py
divya-vijay: Same logic, but here, in `set`, for the very first time when we execute, file would not be present, so we proceed to set it. Raising exception would not be needed.

lr-cli/cortx_setup/commands/signature/get.py
divya-vijay: This was kept in case (most cases) the user is not aware of all signature keys and tries to get an unknown key. So, it gives a warning and returns the entire signature value. 

```
# cortx_setup signature get business-name
nodecli - WARNING - Given key 'business-name' is possibly not found in Signature data.
nodecli - INFO - LR Signature: {'address': 'nodeaddress001', 'name': 'srvnode-1'}
```

lr-cli/cortx_setup/commands/signature/get.py
divya-vijay: This logger was placed temporarily as we did not return values on console in the first commit of LR-CLI integration. (this was the only `get` in all of Factory APIs) 
I think @mazinamdar has added this in latest commit.

lr-cli/cortx_setup/commands/signature/get.py
divya-vijay: Updated to raise exception for `get` and warning for `set`. Thank you!

lr-cli/cortx_setup/commands/signature/set.py
divya-vijay: Temporary logger. removed now.

lr-cli/cortx_setup/commands/signature/set.py
```diff
@@ -58,38 +59,35 @@ def run(self, key=None, value=None):
...
+                    lr_sign_file.stat().st_size != 0):
                 self.logger.warning("Node Signature was never set "
-                               "in ConfStore. Setting now.")
+                                    "in ConfStore. Setting now.")
 
             else:
                 # Immutable file. Make it editable first.
                 _cmd = f"chattr -i {lr_sign_file}"
                 subprocess.Popen(_cmd, shell=True)     # nosec
+                sleep(1)
```
divya-vijay, line 76: When `set` api is executed multiple times back-to-back, fails 1 in 10 times with permission error, as it takes a second to process this immutable step.

divya-vijay: Yes, updated.

ypise (approved):

PR #1286, opened 2021-05-24
EOS-19956: Implement API to query ISO contents

dmitrydb:
I implemented API to query ISO content. I added test to cover this case

### DIFF
--- a/api/python/provisioner/commands/upgrade/get_swupgrade_info.py
+++ b/api/python/provisioner/commands/upgrade/get_swupgrade_info.py
@@ -25,11 +25,11 @@ from provisioner import inputs
 from provisioner.commands.upgrade.set_swupgrade_repo import CortxISOInfo, \
     SetSWUpgradeRepo
 from provisioner.config import (CORTX_ISO_DIR, REPO_CANDIDATE_NAME,
-                                SWUpgradeInfoFields, RELEASE_INFO_FILE)
+                                RELEASE_INFO_FILE)
 from provisioner.errors import SWUpdateRepoSourceError
 from provisioner.pillar import PillarResolver, PillarKey
 
-from provisioner.salt import local_minion_id, cmd_run
+from provisioner.salt import local_minion_id
 from provisioner.commands import CommandParserFillerMixin, commands
 from provisioner.vendor import attr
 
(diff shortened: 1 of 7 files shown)


api/python/provisioner/commands/upgrade/set_swupgrade_repo.py
```diff
@@ -50,6 +50,8 @@
                          ReleaseInfoValidator,
                          YumRepoDataValidator,
                          HashSumValidator)
+from ...vendor import attr
```
andkononykhin2, line 53: please use absolute imports, three dots is too far

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
```diff
@@ -121,6 +132,13 @@ def run(self, release: str = None) -> dict:
         """
         local_minion = local_minion_id()
 
+        if iso_path is not None:
+            # if the `iso_path` is set up, we ignore the `release` parameter
+            # NOTE: get `SetSWUpgradeRepo` from list of commands since all
+            #  objects are correctly defined (post- and pre- states)
+            set_swupgrade_repo_obj = commands['set_swupgrade_repo']
```
andkononykhin2, line 139: why we can't use `SetSWUpgradeRepo` class directly?

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
```diff
@@ -121,6 +132,13 @@ def run(self, release: str = None) -> dict:
...
+            return set_swupgrade_repo_obj.run(iso_path, dry_run=True)
```
andkononykhin2, line 140: It means we return different result format depending on inputs. Could you update to return the same format to include:
- metadata
- packages

It would be useful for other tasks also. Thanks

andkononykhin2 (changes requested):
@dmitrydb Looks good. But I believe we should improve that a bit. Please check my comments. Thanks

api/python/provisioner/commands/upgrade/set_swupgrade_repo.py
dmitrydb: Done

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
dmitrydb: Done

andkononykhin2 (approved):
@dmitrydb Looks very good. Thank you

PR #1287, opened 2021-05-24
EOS-20659: Field Prepare: Time configuration using cortx_setup cli command

rahuljyoti65:

### DIFF
--- a/lr-cli/cortx_setup/commands/node/prepare/time.py
+++ b/lr-cli/cortx_setup/commands/node/prepare/time.py
@@ -1,0 +1,51 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+from cortx_setup.commands.command import Command
+from provisioner import set_ntp
+
+
+class NodePrepareTime(Command):
+    """Cortx Setup API for configuring time"""
+    _args = {
+        'server': {
+            'type': str,
+            'default': 'time.seagate.com',
+            'optional': True
+        },
+        'timezone': {
+            'type': str,
+            'default': 'UTC',
+            'optional': True
+        }
+    }
+
+    def run(self, **kwargs):
+        """Time configuration command execution method.
+
+        Execution:
+        `cortx_setup node prepare time --server <ntp-server> --timezone <timezone>`
+        """
+        ntp_server = kwargs.get('server')
+        ntp_timezone = kwargs.get('timezone')
+        self.logger.info(f"Configuring Time")
+        try:
+            set_ntp(server=ntp_server, timezone=ntp_timezone)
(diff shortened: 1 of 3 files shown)


lr-cli/cortx_setup/commands/node/prepare/time.py
```diff
@@ -0,0 +1,51 @@
...
+        """Time configuration command execution method.
+
+        Execution:
+        `cortx_setup node prepare time --server <ntp-server> --timezone <timezone>`
+        """
+        ntp_server = kwargs.get('server')
+        ntp_timezone = kwargs.get('timezone')
+        self.logger.info(f"Configuring Time")
+        try:
+            set_ntp(server=ntp_server, timezone=ntp_timezone)
```
mazinamdar, line 47: set_ntp will set time server for system and controller so can we have logic of updating date of only system/node in this API ?

@ypise  can you please suggest , 
1. We can update pillar for server and timezone in system.sls
2. from this API call states which update time for system/node only.

rahuljyoti65: comment addressed as discussed

mazinamdar (approved):
looks good to me

lr-cli/cortx_setup/commands/node/prepare/time.py
```diff
@@ -0,0 +1,94 @@
...
+
+        PillarSet().run(
+            'system/ntp/time_zone',
+            ntp_timezone,
+            targets=node_id,
+            local=True
+        )
+
+        try:
+            self.logger.info(f"Setting time on node with server={ntp_server} & timezone={ntp_timezone}")
```
ypise, line 86: Convert this to debug log.
Ideally we would want this to be equivalent to verbose.

lr-cli/cortx_setup/commands/node/prepare/time.py
```diff
@@ -0,0 +1,94 @@
...
+            self.set_server_time()
+            chassis = grains_get("hostname_status:Chassis")[node_id]["hostname_status:Chassis"]
+            if chassis == "server":
```
ypise, line 89: We have decided to use the enclosure id instead, haven't we?
```
if not machine_id in enclosure_id:
    self.logger.debug("")
    self.set_enclosure_time()
```

lr-cli/cortx_setup/commands/node/prepare/time.py
```diff
@@ -0,0 +1,94 @@
...
+                self.logger.info(f"Setting time on enclosure with server={ntp_server} & timezone={ntp_timezone}")
```
ypise, line 90: All logs in this file should be debug or error.

lr-cli/cortx_setup/commands/node/prepare/time.py
```diff
@@ -0,0 +1,94 @@
...
+                self.set_enclosure_time()
+        except Exception as e:
+            print(e)
+            self.logger.error("Time configuration failed")
```
ypise, line 94: print(e) not necessary.
use `self.logger.exception` instead.

ypise (changes requested):
@rahuljyoti65 
Address the suggested comments.
Also, the check for enclosure time set should be purely based on valid enclosure ID

lr-cli/cortx_setup/commands/node/prepare/time.py
mazinamdar: @ypise Actually we thought our main target is to just check it is HW or vm, so I asked him to use that.
If there are other cases then we can change to above code block.

lr-cli/cortx_setup/commands/node/prepare/time.py
rahuljyoti65: Done

lr-cli/cortx_setup/commands/node/prepare/time.py
rahuljyoti65: Done

lr-cli/cortx_setup/commands/node/prepare/time.py
rahuljyoti65: Done

ypise (approved):

PR #1289, opened 2021-05-24
EOS-20658:Temporary assignment of Management VIP to Salt Minion during Field Prepare

sradhanand-pati:
`cortx_setup node prepare server --site_id <site-id> --rack_id <rack-id> --node_id <node-id> --mgmt_vip <mgmt-vip> `
1. It will update the respective values to confstore as well as pillar.
2. Add mgmt_vip as master to minion config file.
3. Create alias of mgmt_vip in the primary node with public management interface


lr-cli/cortx_setup/commands/node/prepare/config.py
```diff
@@ -0,0 +1,105 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+from cortx.utils.conf_store import Conf
+from provisioner.commands import PillarSet
+from provisioner.salt import local_minion_id
+from ...common_utils import (
```
mazinamdar, line 21: can you please use absolute path for all import which are triples

lr-cli/cortx_setup/commands/node/prepare/server.py
```diff
@@ -0,0 +1,68 @@
...
+)
+from provisioner.config import PRVSNR_ROOT_DIR
+from .config import NodePrepareServerConfig
+from ...common_utils import get_pillar_data
+
+
+class NodePrepareServer(NodePrepareServerConfig):
+
+    def _ip_not_pingable(self, ip: str):
+        res = os.system(f"ping -c 1 {ip}")
```
mazinamdar, line 33: can we use cmd_run here ?

sradhanand-pati: here it is expected that ip should not be pingable.
so if using cmd_run in case of non-zero retcode it fails directly.
i tried modifying the command but it didn't help.

ypise (approved):

PR #1293, opened 2021-05-25
EOS-20667: Field API for Storageset

divya-vijay:


lr-cli/cortx_setup/commands/storageset/add/node.py
```diff
@@ -0,0 +1,92 @@
...
+            )
+
+        if len(server_node) > node_count:
+            raise Exception(
+               f"Invalid count: {len(server_node)} number of nodes received. "
+               f"Given Storageset can accept a maximum of {node_count} nodes. "
+               "Update it with `cortx_setup storageset create` command."
+            )
+
+        self.logger.info(
```
ypise, line 75: Change all info logs to debug logs as per guidelines we received during design review.

ypise (approved):
Overall I am good with this change. Need to be careful with 2 aspects:
1. Ensure we don't have info logs as we don't want any additional messages to Console while executing the command, unless there's an error or exception
2. Coordinate with @rahuljyoti65 to address the merge conflicts that might arise in setup.py

lr-cli/cortx_setup/commands/storageset/add/node.py
divya-vijay: Updated. Thanks.

PR #1294, opened 2021-05-26
[EOS-19957][EOS-19964] adds provisioner resources implementation

andkononykhin2:
That PR solves two things:

- provides new resource based implementation for provisioner resources installation on a set of nodes
- makes automation for sw upgrade testing infra preparation much easier

New resources are described in updated [resource.md](https://github.com/andkononykhin2/cortx-prvsnr/blob/EOS-19957_EOS-19964_provisioner-resources/docs/Development/resources.md) doc.

-----
[View rendered docs/Development/resources.md](https://github.com/andkononykhin2/cortx-prvsnr/blob/EOS-19957_EOS-19964_provisioner-resources/docs/Development/resources.md)

### DIFF
--- a/api/python/provisioner/commands/validator/validator.py
+++ b/api/python/provisioner/commands/validator/validator.py
@@ -14,7 +14,6 @@
 # For any questions about this software or licensing,
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
-import hashlib
 import logging
 from abc import abstractmethod, ABC
 from hmac import compare_digest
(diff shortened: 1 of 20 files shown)


api/python/provisioner/commands/validator/validator.py
```diff
@@ -348,22 +347,11 @@ def validate(self, path: Path):
...
-
-        with open(path, 'rb') as fh:
-            while True:
-                data = fh.read(4096)
-                if not data:
-                    break
-                hash_method.update(data)
-
-        if not compare_digest(hash_method.digest(), self.hash_sum):
+        if not compare_digest(hash_obj.digest(), self.hash_sum):
```
ypise, line 352: Please add a comment here mentioning what this digest is and how we are using it in this context of code.

api/python/provisioner/config.py
```diff
@@ -583,6 +585,7 @@ class HashType(Enum):
 
     MD5 = "md5"
     SHA256 = "sha256"
+    SHA384 = "sha384"
```
ypise, line 588: Why this odd one?
Ideally we should stick with the more secure **_sha512_** only.

ypise (approved):
Overall no issues.
Just a few suggestions.

api/python/provisioner/config.py
andkononykhin2: Ok. I set 512 as a default. But I am not sure that ts would be more secure in all the cases. Please check [that](https://en.wikipedia.org/wiki/Hash_function_security_summary), length extension attacks part..

api/python/provisioner/commands/validator/validator.py
andkononykhin2: done

PR #1299, opened 2021-05-26
EOS-20793: remove /etc setup files

divya-vijay:


srv/components/provisioner/teardown/remove_setup_files.sls
```diff
@@ -0,0 +1,43 @@
...
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+
+cluster-id immutable:
+  cmd.run:
```
ypise, line 20: Better to check all files with immutable attributes in a loop, if exists.

ypise (approved):
Check signature and cluster_id files in loop instead of if for existence.
Rest looks good.

PR #1308, opened 2021-05-30
EOS-19896: Factory - Configure storage enclosure

83bhp:


lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -0,0 +1,251 @@
...
+    $ cortx_setup storage config --mode {primary|secondary} --ip <ip-address> --port <port-number>
+    '''
+    _args = {
+        'name': {
+            'type': str,
+            'default': None,
+            'optional': True,
+            'help': 'Set name of the enclosure e.g enc1-r10'
+        },
+        'type': {
```
ypise, line 49: Set `dest` on this argument to avoid type in 
```def run(self, name=None, type=None,```
Check with @SomwanshiAnjali

ypise (changes requested):

lr-cli/cortx_setup/commands/storage/config.py
83bhp: addressed it by using **kwargs

ypise (approved):

PR #1322, opened 2021-06-03
EOS-20788 sw upgrade follow up fixes

andkononykhin2:


ypise (approved):

api/python/provisioner/commands/upgrade/sw_upgrade_node.py
```diff
@@ -83,18 +86,21 @@ def upgrade_sw(self, sw, sw_data, targets):
         # FIXME hard-coded
         StatesApplier.apply([f"{sw_data['base_sls']}.install"], targets)
 
-    def upgrade(self, sw_data, targets=config.ALL_TARGETS):
-        logger.info("Fire pre-upgrade event (node level)")
-        # FIXME pre-upgrade calls
+    def upgrade(self, sw_data, no_events=False, targets=config.ALL_TARGETS):
```
dmitrydb, line 89: It would be great to add doc-string with method parameters descriptions (what does the `no_events` parameter mean?).

dmitrydb (approved):
Looks good to me. I approve of this PR.

api/python/provisioner/commands/upgrade/sw_upgrade_node.py
andkononykhin2: right, let me address that separately

PR #1328, opened 2021-06-06
EOS-20920: added validation for config.ini interfaces and devices

tanujashinde0405:
Signed-off-by: Tanuja Shinde <tanuja.shinde@seagate.com>

### DIFF
--- a/api/python/provisioner/commands/validator/config_sections.py
+++ b/api/python/provisioner/commands/validator/config_sections.py
@@ -19,7 +19,9 @@
 
 import logging
 from provisioner.vendor import attr
-
+from provisioner.salt import cmd_run,local_minion_id
+from provisioner.config import ALL_MINIONS
+from provisioner.api import grains_get
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


ypise (approved):

api/python/provisioner/commands/validator/config_sections.py
```diff
@@ -120,3 +122,40 @@ def _validate_node_count(self, num_of_nodes, parsed):  # noqa: C901
         except Exception as exc:
             raise ValueError(f"Config Failed to apply: {str(exc)}")
 
+    def _validate_config_interfaces(self,config) :
+        """
+        Validate Interfaced passed to config.ini.
+        """
+        for section in config.sections() :
+            if ("srvnode_default" in section) :
```
mazinamdar, line 130: I think we can also provide interfaces in srvnode-1,2,3 section in config.ini in case of different interfaces on each node.

divya-vijay: +1 
Even though these are default section params, would be better to keep it generic to any `srvnode-`

api/python/provisioner/commands/validator/config_sections.py
```diff
@@ -120,3 +122,41 @@ def _validate_node_count(self, num_of_nodes, parsed):  # noqa: C901
...
+                            f"interfaces should be from {ip4_interfaces}"
+                          )
+                    raise ValueError(msg)
+
+    def _validate_config_devices(self,config,number_of_nodes) :
+        """
+        Validate devices passed to config.ini.
+        """
+        for section in config.sections() :
+            devices = cmd_run(
```
divya-vijay, line 148: Would be better to move `devices` initialization outside of loop.

divya-vijay (approved):

api/python/provisioner/commands/validator/config_sections.py
```diff
@@ -120,3 +122,36 @@ def _validate_node_count(self, num_of_nodes, parsed):  # noqa: C901
...
+                "lsblk -nd -o NAME -e 11|grep -v sda|sed 's|sd|/dev/sd|g'|paste -s -d, -",
+                targets=ALL_MINIONS
+            )
+        for section in config.sections() :
+
+            if ("srvnode" in section):
+                if (config.has_option(section,'storage.cvg.0.data_devices')):
+                    if 'default' in section:
+                        node_list = ['srvnode-'+str(num) for num in range(1, int(number_of_nodes)+1)]
```
mazinamdar, line 145: please add spaces between operator
`['srvnode-' + str(num) for num in range(1, int(number_of_nodes) + 1)]`

mazinamdar (approved):
looks good , please address that minor comment

PR #1331, opened 2021-06-07
EOS-20918: Create granular command for mini-provisioner config

sradhanand-pati:
command - 
`cortx_setup node prepare cortx`
This command would deploy platform states and the prepare stages for cortx_components.


mazinamdar (approved):
looks good to me

PR #1333, opened 2021-06-07
EOS-20969: Bootstrap and Env Prepare steps

divya-vijay:


lr-cli/cortx_setup/commands/cluster/config/component.py
```diff
@@ -34,49 +62,176 @@
...
+            self.logger.debug("Setting up Cluster ID on the system")
+            cluster_id.ClusterId().run()
+
+            self.logger.debug("Encrypting config data")
+            EncryptPillar().run()
+
+            self.logger.debug("Exporting to Confstore")
+            confstore_export.ConfStoreExport().run()
+
+            self.logger.info("Bootstrap done. Proceeding to deploy components..")
```
divya-vijay, line 216: Overall only 2 info msgs - both starting at bootstrap and deploy. so, the user will not be too blank on process.

ypise (approved):

PR #1338, opened 2021-06-08
EOS-21033, EOS-21058: Fail if any destroy stage fails

ypise:


mazinamdar (approved):
looks good to me

PR #1342, opened 2021-06-09
EOS-20669: Update provisioner to use new ISO structure in scope of upgrade

dmitrydb:
The initial commit for the new ISO structure for the SW upgrade. I added the base changes for the new ISO structure.


api/python/provisioner/commands/validator/validator.py
```diff
@@ -173,6 +173,9 @@ class DirValidator(PathValidator):
         if True validation raises an Exception if the directory
         doesn't exist
 
+    empty: bool
```
andkononykhin, line 176: empty_ok might be more clear here

api/python/provisioner/commands/validator/validator.py
```diff
@@ -235,6 +242,8 @@ def validate(self, path: Path):
             raise ValidationError(reason=f"'{path}' is not a directory")
 
         if self.files_scheme:
+            if self.empty and not any(path.iterdir()):
```
andkononykhin, line 245: If empty is false and no dirs makes sense to fail fast right here

api/python/provisioner/commands/upgrade/set_swupgrade_repo.py
```diff
@@ -105,11 +198,218 @@ def __str__(self):
...
+        """
+        return f'{iso_root_dir}/{RELEASE_INFO_FILE}'
+
+
 @attr.s(auto_attribs=True)
 class SetSWUpgradeRepo(SetSWUpdateRepo):
 
     input_type: Type[inputs.SWUpgradeRepo] = inputs.SWUpgradeRepo
     _BASE_DIR_PILLAR = "release/upgrade/base_dir"
+    _version_proxy: dict = attr.ib(
```
andkononykhin, line 368: If no input for that attribute is expected then init=False can be used

andkononykhin (approved):

andkononykhin2 (approved):

PR #1346, opened 2021-06-09
EOS-21101: lr-cli fetch enclosure_id from controller with plain password

rahuljyoti65:

### DIFF
--- a/lr-cli/cortx_setup/commands/storage/commons.py
+++ b/lr-cli/cortx_setup/commands/storage/commons.py
@@ -15,9 +15,6 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-import hashlib
-import json
-import requests
 from logging import exception
 from ..command import Command
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -89,155 +93,262 @@ class StorageEnclosureConfig(Command):
...
+
+        if enc_file_path.exists():
+            if Path(enc_file_path).stat().st_size != 0:
+                self.flag = True
+                with open(enc_file_path, "r") as file:
+                    self.enclosure_id = file.read().replace('\n', '')
+        else:
+            self.enclosure_id = None
+
+    def store_enclosure_id(self):
```
mazinamdar, line 110: If possible can we use concept of mapping to avoid writing similar code for store functions.

refer this: https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/lr-cli/cortx_setup/commands/node/prepare/config.py#L34

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -89,155 +134,122 @@ class StorageEnclosureConfig(Command):
         }
     }
 
+    def __init__(self):
+        super().__init__()
+        if enc_file_path.exists():
+            if Path(enc_file_path).stat().st_size != 0:
+                with open(enc_file_path, "r") as file:
+                    enclosure_id = file.read().replace('\n', '')
```
ypise, line 142: Something seems to have been missed here.

lr-cli/cortx_setup/commands/storage/commons.py
```diff
@@ -56,3 +59,41 @@ def get_enc_id(self, targets=None, force=False):
                 return self.fetch_enc_id(targets)
             except:
                 raise exception("Error generating the enclosure ID")
+
+
+class EnclosureInfo(Command):
```
ypise, line 64: Move this class to a separate file. It improves the readability.

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -89,155 +134,122 @@ class StorageEnclosureConfig(Command):
...
-                Conf.set(
-                    'node_info_index',
-                    f'storage_enclosure>{enc_id}>controller>{mode}>port',
-                    port
-                )
+
+        if user != None and password != None and ip != None and port != None:
+            # fetch enclosure serial/id
+            enclosure_id = EnclosureInfo(ip, user, password, port).get_enclosure_serial()
+            machine_id = grains_get("machine_id")[node_id]["machine_id"]
```
ypise, line 180: I have a feeling machine_id and enclosure_id should be class instance variables.

ypise (changes requested):

lr-cli/cortx_setup/commands/storage/commons.py
rahuljyoti65: Done

lr-cli/cortx_setup/commands/storage/config.py
rahuljyoti65: Comment addressed!

lr-cli/cortx_setup/commands/storage/config.py
rahuljyoti65: Here I'm just checking if /etc/enclosure_id file is present and it is not empty and then storing the file content(i.e value) in the enclosure_id variable.

ypise (approved):

83bhp (approved):

PR #1376, opened 2021-06-16
EOS-21287, EOS-21227, EOS-21197: lr-cli factory server config and factory storage config for vm

rahuljyoti65:

### DIFF
--- a/lr-cli/cortx_setup/commands/server/config.py
+++ b/lr-cli/cortx_setup/commands/server/config.py
@@ -1,0 +1,89 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+from ..command import Command
+from cortx.utils.conf_store import Conf
+from cortx_setup.commands.common_utils import get_machine_id
+from cortx_setup.config import CONFSTORE_CLUSTER_FILE
+from provisioner.salt import local_minion_id
+from provisioner.commands import PillarSet
+
+
+class ServerConfig(Command):
+
+    _args = {
+        'name': {
+            'type': str,
+            'default': None,
+            'optional': True,
+            'help': 'Set name of the server'
+        },
+        'type': {
+            'type': str,
+            'default': None,
+            'optional': True,
+            'choices': ['VM', 'HW'],
+            'help': 'Server type e.g {VM|HW}'
+        }
+    }
+
+
+    def run(self, **kwargs):
+        """
+        Server configuration command execution method
+
+        Command:
+        `cortx_setup server config --name <name> --type {VM|HW}}`
+        """
+
+        node_id = local_minion_id()
+        machine_id = get_machine_id(node_id)
+
+        conf_pillar_map = {
+            'name': (
+                f'cluster/{node_id}/name',
+                f'server_node>{machine_id}>name'
+            ),
+            'type': (
+                f'cluster/{node_id}/type',
+                f'server_node>{machine_id}>type'
+            )
+        }
+
+        Conf.load(
+            'node_info_index',
+            f'json://{CONFSTORE_CLUSTER_FILE}'
+        )
+
+        for key, value in kwargs.items():
+            if value:
+                if key in conf_pillar_map.keys():
+                    self.logger.debug(f"Updating pillar with key:{conf_pillar_map[key][0]} and value:{value}")
+                    PillarSet().run(
+                        conf_pillar_map[key][0],
+                        value,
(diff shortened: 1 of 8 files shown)


lr-cli/cortx_setup/commands/server/config.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        Server configuration command execution method
+
+        Command:
+        `cortx_setup server config --name <name> --type {VM|HW}}`
+        """
+
+        conf_pillar_map = {
+            'name': (
+                'server/name',
```
mazinamdar, line 53: should it be something like
```
pillar_key : cluster/srvnode-0/type
confstore: server_node>{machine_id}>type
```
?

rahuljyoti65: @mazinamdar @83bhp will the node_id be present at the time of server config?

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -81,12 +83,14 @@ class StorageEnclosureConfig(Command):
         'controller_type': {
             'type': str,
             'optional': True,
+            'choices': ['gallium', 'indium', 'virtual'],
```
ypise, line 86: Avoid using 'gallium', 'indium'.
Instead we could use 'choices': ['<one of Seagate controller names>', 'virtual']

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -174,6 +178,14 @@ def store(self, key, value):
             value
         )
 
+    def store_in_file(self):
```
ypise, line 181: Do not repeat. This is being handled in https://github.com/Seagate/cortx-prvsnr/blob/151c7d164ea4aae589a4a43fdc533da791883e0c/srv/components/system/storage/enclosure_id/config/set.sls#L22
If it is not setting the values in file, update the module code instead.

@tanujashinde0405 you have a better understanding of this part, need your help here.

lr-cli/cortx_setup/commands/server/config.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Command:
+        `cortx_setup server config --name <name> --type {VM|HW}}`
+        """
+
+        node_id = local_minion_id()
+        machine_id = get_machine_id(node_id)
+
+        conf_pillar_map = {
+            'name': (
```
ypise, line 57: @mazinamdar this looks shabby to capture both pillar and confstore format without clear demarcation.
Can this be improvised as:
```
{
  'name': (
    'pillar_key': f'cluster/{node_id}/hostname',
    'confstore_key': f'server_node>{machine_id}>name'
  ),
  'type': (
    'pillar_key': f'cluster/{node_id}/node_type',
    'confstore_key': f'server_node>{machine_id}>type'
  )
}
```

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -191,91 +203,112 @@ def run(self, **kwargs):
...
-            self.refresh_conf_pillar_map()
+        setup_type = Conf.get (
+            f'server_node>{self.machine_id}>type'
+        )
 
-            self.logger.debug(f"Writing storage enclosure_id to the file {enc_file_path}")
-            file = open(enc_file_path, "w")
-            file.write(self.enclosure_id + "\n")
-            file.close()
+        if setup_type == "VM":
```
ypise, line 211: Most of code after this can be dropped in favor of https://github.com/Seagate/cortx-prvsnr/blob/151c7d164ea4aae589a4a43fdc533da791883e0c/srv/components/system/storage/enclosure_id/config/set.sls

lr-cli/cortx_setup/commands/storage/config.py
rahuljyoti65: These choices are mentioned in the factory confluence page for the controller_type argument.

lr-cli/cortx_setup/commands/storage/config.py
rahuljyoti65: @ypise we can't use set.sls bcoz the fetch_enclosure_id method requires the ip, user, secret and port to be available in the pillar which is not the case during factory storage config.

lr-cli/cortx_setup/commands/server/config.py
rahuljyoti65: @ypise yes we can use nested dictionary or rather we can have some convention defined as in index 0 for pillar_key and index 1 for confstore key for such logic in our codebase - this will reduce the redundant witing of the "pillar_key" and "confstore_key" for each argument

lr-cli/cortx_setup/commands/storage/config.py
rahuljyoti65: I dont think we can drop the current code instead we can drop the set.sls file as whatever that module does is already handled during factory storage config

lr-cli/cortx_setup/commands/storage/config.py
ypise: Ok, then we have no choice :(

lr-cli/cortx_setup/commands/server/config.py
ypise: >index 0 for pillar_key and index 1 for confstore key
In the long run that turns out to be a pain area. We have faced that with Networking definition in config.ini in R1.

rahuljyoti65: Comment addressed!
created 2 different dictionaries pillar_key_map and conf_key_map for better demarcation of pillar and confstore keys

83bhp (approved):
Looks good to me.

lr-cli/cortx_setup/commands/server/config.py
mazinamdar: at initial every node will be srvnode-0 , once we do bootstrap we get actual node_id

mazinamdar (approved):
looks good to me

PR #1378, opened 2021-06-16
Fix for jenkins deploy job failure

rahuljyoti65:

### DIFF
--- a/devops/ci/single_node_deploy/deploy.jenkinsfile
+++ b/devops/ci/single_node_deploy/deploy.jenkinsfile
@@ -127,12 +127,11 @@ node ('provisioning'){
                     yum-config-manager --add-repo "${CORTX_RELEASE_REPO}/3rd_party/"
                     yum-config-manager --add-repo "${CORTX_RELEASE_REPO}/cortx_iso/"
 
-                    echo '''
+                    cat << EOL > /etc/pip.conf
                     [global]
                     timeout: 60
                     index-url: $CORTX_RELEASE_REPO/python_deps/
-                    trusted-host: \$(echo $CORTX_RELEASE_REPO | awk -F '/' '{print \$3}')
-                    ''' >/etc/pip.conf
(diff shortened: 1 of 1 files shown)


andkononykhin2 (approved):

devops/ci/single_node_deploy/deploy.jenkinsfile
```diff
@@ -127,12 +127,11 @@ node ('provisioning'){
...
                     yum-config-manager --add-repo "${CORTX_RELEASE_REPO}/cortx_iso/"
 
-                    echo '''
+                    cat << EOL > /etc/pip.conf
                     [global]
                     timeout: 60
                     index-url: $CORTX_RELEASE_REPO/python_deps/
-                    trusted-host: \$(echo $CORTX_RELEASE_REPO | awk -F '/' '{print \$3}')
-                    ''' >/etc/pip.conf
+                    trusted-host: \$(echo $CORTX_RELEASE_REPO | awk -F '/' '{print \$3}')\nEOL
```
ypise, line 134: The issue here was with use of `'''` instead of `"""`.
The current one is a better approach, just ensure that in
\nEOL
`\n` is not required, so avoid it
`EOL` should be on a new line.

ypise (approved):
Incorporate and test the suggestion.

devops/ci/single_node_deploy/deploy.jenkinsfile
rahuljyoti65: Actually, \n is required
If we use EOL on new line, the cat command doesn't complete because of the indentation in the jenkins script. and this is how it get's stored in the file.
```
[root@node ~]# cat /etc/pip.conf
                    [global]
                    timeout: 60
                    index-url: http://cortx-storage.colo.seagate.com/releases/cortx/github/main/centos-7.8.2003/1345/prod//python_deps/
                    trusted-host: cortx-storage.colo.seagate.com
                    EOL
                    sed -ie 's/^[       ]*//' /etc/pip.conf
```

PR #1380, opened 2021-06-16
EOS-19686: Logic incorporated for unsupported feature

ypise:

### DIFF
--- a/srv/_modules/cortx_unsupported_features.py
+++ b/srv/_modules/cortx_unsupported_features.py
@@ -1,0 +1,150 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+# How to test:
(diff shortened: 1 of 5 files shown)


srv/_modules/cortx_unsupported_features.py
```diff
@@ -0,0 +1,197 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+# How to test:
+# $ salt-call saltutil.clear_cache
+# $ salt-call saltutil.sync_modules
+
```
83bhp, line 21: Have you missed to add the actual test command here?

srv/components/csm/config/config.sls
```diff
@@ -15,6 +15,19 @@
...
 
+
+Register features to be unsuported:
+  module.run:
+    - cortx_unsupported_features.register:
+      - component: provisioner
+      - unsupported_feature_list:
+        - fw_update
+      - setup_types:
+        - VM
```
83bhp, line 26: Since the setup type RBOD is missing here, does that mean the fw update will be supported only for RBOD?

83bhp (approved):
minor comment/question for understanding, patch looks ok. Thanks!

srv/components/csm/config/config.sls
ypise: Yes for now we should maintain support for RBOD. But once it is officially dropped from our supported features list, we can disable it for RBOD as well by listing here.

srv/_modules/cortx_unsupported_features.py
ypise: Actual test command is tricky. However, I'll include the test process here.

srv/components/csm/config/config.sls
ypise: I have made the addition to `test/r2.md` for manual testing.

srv/_modules/cortx_unsupported_features.py
ypise: Updated the basic commands necessary for testing.

83bhp: Thanks!! 
There are some codacy warnings, please address.

83bhp (approved):

PR #1381, opened 2021-06-16
[EOS-21202] API to cleanup local salt configuration

mazinamdar:

### DIFF
--- a/lr-cli/cortx_setup/commands/__init__.py
+++ b/lr-cli/cortx_setup/commands/__init__.py
@@ -17,6 +17,7 @@
 
 from .hostname import Hostname
 from .pillar_sync import PillarSync
+from .salt_cleanup import SaltCleanup
 from .network.config import NetworkConfig
 from .node.initialize import NodeInitalize
 from .node.prepare.server import NodePrepareServer
@@ -72,5 +73,7 @@ __all__ = [
     'ResourceDiscover',
     'SecurityConfig',
     'SetSignature',
-    'StorageEnclosureConfig'
-]
+    'StorageEnclosureConfig',
+    'SaltCleanup',
+    'PillarSync'
+ ]
--- a/lr-cli/cortx_setup/commands/salt_cleanup.py
+++ b/lr-cli/cortx_setup/commands/salt_cleanup.py
@@ -17,24 +17,14 @@
 
 from provisioner.salt import local_minion_id
 from provisioner.utils import run_subprocess_cmd
-from provisioner.commands.grains_get import GrainsGet
 from .command import Command
 
 
 class SaltCleanup(Command):
     _args = {}
 
-    def _get_grains_id(self):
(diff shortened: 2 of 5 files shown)


lr-cli/cortx_setup/commands/salt_cleanup.py
```diff
@@ -0,0 +1,44 @@
...
+from provisioner.commands.grains_get import GrainsGet
+from .command import Command
+
+
+class SaltCleanup(Command):
+    _args = {}
+
+    def _get_grains_id(self):
+        res = GrainsGet().run('id')
+        if res.get(local_minion_id()):
```
83bhp, line 29: [Suggestion] Please add comment about the logic to validate the id.

83bhp (approved):
Looks good to me, added one minor suggestion, also please update the "LR CLI Design" confluence page once this is merged.

lr-cli/cortx_setup/commands/salt_cleanup.py
mazinamdar: I have removed _get_grains_id because local_minion_id works exactly same . It will returns id from grains

lr-cli/cortx_setup/commands/__init__.py
```diff
@@ -72,5 +73,7 @@
     'ResourceDiscover',
     'SecurityConfig',
     'SetSignature',
-    'StorageEnclosureConfig'
-]
+    'StorageEnclosureConfig',
+    'SaltCleanup',
+    'PillarSync'
```
ypise, line 78: PillarSync is repeated twice. Remove one to avoid confusion.

lr-cli/cortx_setup/commands/salt_cleanup.py
```diff
@@ -0,0 +1,34 @@
...
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+from provisioner.salt import local_minion_id
+from provisioner.utils import run_subprocess_cmd
+from .command import Command
+
+
+class SaltCleanup(Command):
+    _args = {}
```
ypise, line 24: Add a comment here that the variable declaration is for the parent Object.
Can this declaration be avoided?

ypise (approved):
Few comments to be addressed.

lr-cli/cortx_setup/commands/__init__.py
mazinamdar: I think this was missed earlier so added here

lr-cli/cortx_setup/commands/salt_cleanup.py
mazinamdar: Yes we can avoid declaring empty _args I have handled in parent class

ypise: Looks good.

PR #1383, opened 2021-06-16
EOS-21185: Increase replication factor for 3-node kafka cluster

ypise:

### DIFF
--- a/srv/components/misc_pkgs/kafka/config.sls
+++ b/srv/components/misc_pkgs/kafka/config.sls
@@ -106,3 +106,34 @@ Update connections in kafka config:
     - pattern: ^zookeeper.connect=.*
     - repl: zookeeper.connect={{ node_hosts|join(',')}}
     - append_if_not_found: True
+
+{% if 1 < server_nodes|length %}
+# As per guidelines in EOS-21185
+Update default replication factor in kafka config:
+  file.replace:
+    - name: /opt/kafka/config/server.properties
+    - pattern: ^default.replication.factor=.*
+    - repl: default.replication.factor=3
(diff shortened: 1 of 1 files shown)


srv/components/misc_pkgs/kafka/config.sls
```diff
@@ -106,3 +106,34 @@ Update connections in kafka config:
     - pattern: ^zookeeper.connect=.*
     - repl: zookeeper.connect={{ node_hosts|join(',')}}
     - append_if_not_found: True
+
+{% if 1 < server_nodes|length %}
```
sachinpunadikar, line 110: Just to reconfirm, this condition is true when number of nodes are 3 (or more than 1) ?

ypise: Yes
```
if 1 less than length of list (server_nodes)
```
So it is not specific to 3 nodes, but any number more than 1.

@sachinpunadikar 
I was thinking should the same logic be used to define the values for 
>default.replication.factor=3
>offsets.topic.replication.factor=3
>transaction.state.log.replication.factor=3

sachinpunadikar: @ypise thanks for explanation.
The config values are 3 till the time we are keeping kafka cluster limited to 3 nodes. Variation in kafka cluster size will have different config values and those will be derived by running some tests.
Hence current logic is good. 
kafka_setup will be useful to handle such cases in future.

sachinpunadikar (approved):
+1

andkononykhin2 (approved):

83bhp (approved):

PR #1390, opened 2021-06-18
EOS-19965: [SW_UPGRADE][HOT_FIX]: GPG signature validation fix & other minor fixes

dmitrydb:
Convert explicitly GPG public key to OpenPGP format. I added some minor fixes and improvements

### DIFF
--- a/api/python/provisioner/commands/validator/validator.py
+++ b/api/python/provisioner/commands/validator/validator.py
@@ -686,13 +686,17 @@ class AuthenticityValidator(PathValidator):
 
         # NOTE: return given path itself if it is in OpenPGP format already
         res = pub_key_path
-        with open(pub_key_path) as fh:
+        with open(pub_key_path, "rb") as fh:
+            # NOTE: read file as binary file since OpenPGP is binary format
             content = fh.readlines()
             armor_header = content[0]
             armor_tail = content[-1]
 
-        if (ARMOR_HEADER in armor_header
-                and ARMOR_TAIL in armor_tail):
+        # NOTE: we check that the armor header and armor tail in binary
+        #  representation exist in the first and the last line of
+        #  the pub key file.
+        if (ARMOR_HEADER.encode() in armor_header
+                and ARMOR_TAIL.encode() in armor_tail):
             # NOTE: it means that provided public key is in ASCII Armor format
             cmd = f"gpg --yes --dearmor {pub_key_path.resolve()}"
             try:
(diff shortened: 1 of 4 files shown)


api/python/provisioner/commands/validator/validator.py
```diff
@@ -661,6 +664,59 @@ class AuthenticityValidator(PathValidator):
...
+            armor_header = content[0]
+            armor_tail = content[-1]
+
+        # NOTE: we check that the armor header and armor tail in binary
+        #  representation exist in the first and the last line of
+        #  the pub key file.
+        if (ARMOR_HEADER.encode() in armor_header
+                and ARMOR_TAIL.encode() in armor_tail):
+            # NOTE: it means that provided public key is in ASCII Armor format
+            cmd = f"gpg --yes --dearmor {pub_key_path.resolve()}"
```
andkononykhin2, line 701: @dmitrydb what if someone does that twice so the output file is already exists? will gpg replace it or fail?

api/python/provisioner/commands/validator/validator.py
```diff
@@ -661,6 +664,59 @@ class AuthenticityValidator(PathValidator):
...
+            try:
+                # NOTE: by default gpg tool converts the given file to the file
+                #  with the same name + '.gpg' extension at the end.
+                #  Directory is the same
+                cmd_run(cmd, targets=local_minion_id(),
+                        fun_kwargs=dict(python_shell=True))
```
andkononykhin2, line 707: I believe python_shell=True is not required here (no pipes or redirections), we need just path a list instead

api/python/provisioner/commands/validator/validator.py
```diff
@@ -661,6 +664,59 @@ class AuthenticityValidator(PathValidator):
...
+            except Exception as e:
+                logger.error("Can't convert ASCII Armor GPG public key "
```
andkononykhin2, line 709: could you mention path to pub key here?

andkononykhin2 (approved):
@dmitrydb Looks good. Thanks.

api/python/provisioner/commands/validator/validator.py
dmitrydb: @andkononykhin2 yes, you are right, the same file will be replaced. I checked this several times. For that reason I used `--yes` option:
```
gpg --dearmor ./RPM-GPG-KEY-Seagate
File `./RPM-GPG-KEY-Seagate.gpg' exists. Overwrite? (y/N)
```

api/python/provisioner/commands/validator/validator.py
dmitrydb: Yes, sure. I added given public key into error message output.

api/python/provisioner/commands/validator/validator.py
dmitrydb: Done

PR #1393, opened 2021-06-21
EOS-21362: Generate resource map

sradhanand-pati:


mazinamdar (approved):
looks good to me

lr-cli/cortx_setup/commands/resource/discover.py
```diff
@@ -15,13 +15,47 @@
...
+import time
 from ..command import Command
+from provisioner.commands import PillarSet
+from cortx.utils.discovery import Discovery
 
 
 class ResourceDiscover(Command):
+
+    def get_resource_map(self, resource_type='nodes[0]'):
+        retries = 30
```
nitin-seagate, line 27: Can we make retries and sleep as configurable values at some common location so that we can tweak them whenever we run into issue (like as the resources grow in size, the time required to fetch resources will take longer)

lr-cli/cortx_setup/commands/resource/discover.py
```diff
@@ -15,13 +15,47 @@
...
+        status = ""
+        self.logger.debug("Running resource discover command")
+        # Currently since this is a bug once fixed
```
nitin-seagate, line 30: Please mention bug-id on which we are blocked

lr-cli/cortx_setup/commands/resource/discover.py
```diff
@@ -15,13 +15,47 @@
...
+        # we wont be needing to provide any parameter
+        request_id = int(Discovery.generate_node_health(resource_type))
```
nitin-seagate, line 32: Do we have one to one mapping of Provisioner related resource type to SSPL related resource type?

nitin-seagate: Also lets log message with generated request ID?

nitin-seagate (changes requested):

lr-cli/cortx_setup/commands/resource/discover.py
ypise: Is `Discovery` a static class or a class method or has initialization been missed?
static would be weird, but can't help if it is.

lr-cli/cortx_setup/commands/resource/discover.py
sradhanand-pati: Done

lr-cli/cortx_setup/commands/resource/discover.py
sradhanand-pati: @ypise Yes it is static.
@nitin-seagate  Done

lr-cli/cortx_setup/commands/resource/discover.py
sradhanand-pati: Done.
Added it in the config.py file.

83bhp (approved):

PR #1395, opened 2021-06-21
EOS-19665: Get/Set release version logic

andkononykhin2:
New logic how to detect / set Cortx release:

- `/opt/seagate/cortx/RELEASE.INFO` now keeps the release data
- it is created by `provisioner set_release <release>` command as part of the upgrade
  -  actually using python API which is available as well (please check `provisioner.commands.release` for more details)
- if the file is missed (TODO the deployment case) the logic fallbacks to a legacy way 
  - compares list of installed rpms with the release info of available upgrade releases
  - if no match - gets factory release as a decision
 
New commands:

- `provisioner release get_release [--factory]`
- `provisioner release set_release <release>`

Deprecating:
- `provisioner get_release_version`
- `provisioner get_factory_version`


83bhp (approved):

api/python/provisioner/commands/release/release.py
```diff
@@ -0,0 +1,291 @@
...
+        converter=release_url_converter,
+        validator=attr.validators.instance_of(ParseResult),
+    )
+
+    def __str__(self) -> str:
+        return self.url.geturl()
+
+    @property
+    def path(self) -> str:
+        return f"{self.url.netloc}{self.url.path}"
```
dmitrydb, line 61: Do we always expect that `self.url.pah` starts with `/` slash symbol? Maybe it is better to leave such a comment to help readers to understand that it is not a mistake.

api/python/provisioner/commands/release/release.py
```diff
@@ -0,0 +1,291 @@
...
+                else:
+                    # FIXME: it may be remote repo
+                    self._iso_version = config.ISOVersion.VERSION1
+
+        return self._iso_version
+
+    @property
+    def metadata_url(self) -> CortxReleaseUrl:
+        """
+        Return URL to a Cortx repository metadata
```
dmitrydb, line 195: Just a suggestion. Maybe it is better to name it 'URI', not the 'URL'. Because URL is a `http`/`https`/`ftp` scheme, but `file://` is a file URI scheme. So users can expect that this method returns the path to the remote resource instead of the path to the local file. URI is more a broader concept than URL.

api/python/provisioner/commands/set_swupdate_repo.py
```diff
@@ -249,6 +249,9 @@ def dynamic_validation(self, params, targets: str, dry_run: bool = False):  # no
                     str(repo.source), f"malformed repo: '{exc}'"
                 )
 
+            # FIXME should respect different version of iso,
```
dmitrydb, line 252: This comment is not clear since the `set_swupgrade_repo` command has support of two existing ISO versions.

api/python/provisioner/commands/upgrade/check_iso_authenticity.py
```diff
@@ -119,15 +119,23 @@ def run(self, iso_path: str, sig_file: str, gpg_pub_key: str = None,
         -------
 
         """
-        if import_pub_key and gpg_pub_key is not None:
-            self._import_gpg_public_key(gpg_pub_key)
-            gpg_pub_key = None  # NOTE: GPG public key is already imported
+        # to make converters and validators work
+        run_args = CheckISOAuthenticityArgs(
```
dmitrydb, line 123: Why do we need this parameters aggregation into the `CheckISOAuthenticityArgs` instance is we have this mechanism at the top of the class definition:
```python
_run_args_type = CheckISOAuthenticityArgs
```
So, maybe, it is better to remove this and leave only one approach?

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
```diff
@@ -177,15 +127,19 @@ def run(self, iso_path: str = None,
             Cortx repo metadata
 
         """
+        run_args = GetSWUpgradeInfoArgs(iso_path, release)
```
dmitrydb, line 130: Why do we need this if we have 
```
  _run_args_type = GetSWUpgradeInfoArgs
```
?

test/api/python/provisioner/test_sw_upgrade.py
```diff
@@ -90,7 +90,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 
 @pytest.mark.unit
-def test_get_swupgrade_info():
+def test_get_swupgrade_info(tmpdir_function):
```
dmitrydb, line 93: Who is responsible to provide this `tmpdir_function` function?

dmitrydb (approved):
I approve but few places are unclear to me

test/api/python/provisioner/test_sw_upgrade.py
andkononykhin2: it's one of our custom fixtures

api/python/provisioner/commands/upgrade/get_swupgrade_info.py
andkononykhin2: `_run_args_types` is used as a declaration of vars but is not called by generic logic so no converters and validators are applied, e.g. relative paths leaves as is at it failed in some cases during testing (gpg could not find a file)

api/python/provisioner/commands/upgrade/check_iso_authenticity.py
andkononykhin2: _run_args_types is used as a declaration of vars but is not called by generic logic so no converters and validators are applied, e.g. relative paths leaves as is at it failed in some cases during testing (gpg could not find a file)

api/python/provisioner/commands/release/release.py
andkononykhin2: yes, it's how urllib works here, I added the comment

api/python/provisioner/commands/set_swupdate_repo.py
andkononykhin2: yes, but that comment was for older logic, I removed that comment since it was wrong actually: legacy `sw_update` logic shouldn't support that

api/python/provisioner/commands/release/release.py
andkononykhin2: correct, I updated that. Thanks

api/python/provisioner/commands/upgrade/check_iso_authenticity.py
dmitrydb: Ok, got it, thank you for your exploration. `_run_args_types` misled me, I thought that the args class is enough not only for parameters declaration and for parameters conversion and validation if setup `convertor` and validator `parameters` for the particular attribute.
Maybe, it is better to refer to `CheckISOAuthenticityArgs` via `self._run_args_type` as a suggestion.

test/api/python/provisioner/test_sw_upgrade.py
dmitrydb: So if I understand correctly, is it safe to miss/do not specify any parameters(fixtures) in the test function definition?

api/python/provisioner/commands/upgrade/check_iso_authenticity.py
andkononykhin2: yes, agree. Generally the common logic should call that and pass the result to `run` methods instead of the current set of arguments.

test/api/python/provisioner/test_sw_upgrade.py
andkononykhin2: yes, sure. If no fixtures are mentioned pytest will not do any preparation steps (but there might be [auto applied](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request)) fixtures

PR #1413, opened 2021-06-25
EOS-20429: SW Upgrade pre validation - upgrade of CORTX components is compatible with the current version

dmitrydb:
I added SW upgrade packages compatibility validation.


api/python/provisioner/commands/validator/validator.py
```diff
@@ -784,3 +787,128 @@ def __attrs_post_init__(self):
...
+
+        error_msg = list()
+        for pkg in iso_info.packages:
+            if (SWUpgradeInfoFields.VERSION_COMPATIBILITY.value
+                    in iso_info.packages[pkg]):
+                compatibility = iso_info.packages[pkg][
+                    SWUpgradeInfoFields.VERSION_COMPATIBILITY.value]
+
+                installed_ver = packages.get(pkg, None)
+                if installed_ver is None:
```
andkononykhin2, line 893: if a package is not installed it should be ok to install any new version, so it's not an error

api/python/provisioner/commands/validator/validator.py
```diff
@@ -784,3 +787,128 @@ def __attrs_post_init__(self):
...
+        Raises
+        ------
+        ValidationError
+            If validation is failed.
+        """
+        packages = list(iso_info.packages.keys())
+
+        # NOTE: the first line of `yum -q list installed` command is
+        #  'Installed Packages' skip it via `tail -n +2`
+        cmd = (f"yum -q list installed {packages} | tail -n +2 | "
```
andkononykhin2, line 860: edge case: packages list is empty

api/python/provisioner/commands/validator/validator.py
```diff
@@ -784,3 +787,128 @@ def __attrs_post_init__(self):
...
+        Returns
+        -------
+        None
+
+        Raises
+        ------
+        ValidationError
+            If validation is failed.
+        """
+        packages = list(iso_info.packages.keys())
```
andkononykhin2, line 856: shouldn't we check Cortx release constraint as well?

api/python/provisioner/commands/validator/validator.py
```diff
@@ -784,3 +787,128 @@ def __attrs_post_init__(self):
...
+@attr.s(auto_attribs=True)
+class CompatibilityValidator:
+
+    """
+    Validator that checks if all SW upgrade packages are compatible with
+    installed ones.
+    """
+
+    @staticmethod
+    def _compare_versions(a: str, b: str, operator: str) -> bool:
```
andkononykhin2, line 801: that custom implemetation looks good, but I would suggest to simplify our life and use [packaging](https://packaging.pypa.io/en/latest/specifiers.html) here

```bash
>>> from packaging.specifiers import Version, SpecifierSet
>>> old_v = Version('1.0.0-321')
>>> new_v = Version('2.0.0-321')
>>> constraint = SpecifierSet('> 2.0.0-0')
>>> old_v in constraint
False
>>> new_v in constraint
True
>>> old_v2 = Version('2.0.0-0')
>>> old_v2 in constraint
False
```

andkononykhin2 (commented):
@dmitrydb Looks good. I have one major comment regarding not existent pkg error. We shouldn't fail for that case. Also please consider to use packaging library for versions comparison. Thanks

api/python/provisioner/commands/validator/validator.py
dmitrydb: Yes, I rewrote this part using the proposed functionality, thank you! I didn't expect that `packaging` supports such API.

api/python/provisioner/commands/validator/validator.py
dmitrydb: Done.

api/python/provisioner/commands/validator/validator.py
dmitrydb: It is not a problem, but it is better to stop validation in this case to save the CPU time. Done

api/python/provisioner/commands/validator/validator.py
dmitrydb: I added that validation for CORTX version

andkononykhin2 (approved):

PR #1417, opened 2021-06-25
fpm based script to build lr-cli pkg

ghost:
pre-requisites: `docker` is installed
run example: `bash devops/rpms/build_lr_cli.sh -d` will build rpm package in the current folder
please check `bash devops/rpms/build_lr_cli.sh -h` for more details

### DIFF
--- a/devops/rpms/build_lr_cli.sh
+++ b/devops/rpms/build_lr_cli.sh
@@ -20,7 +20,7 @@
 set -eu
 
 script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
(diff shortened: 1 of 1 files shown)


nitin-seagate (approved):
Can you help cite output example of this script...
We will need to call this script along with provisioner api install as part of VM/HW startup script...

andkononykhin2:
> Can you help cite output example of this script...
> We will need to call this script along with provisioner api install as part of VM/HW startup script...

Output is a `python3-cortx_setup` package in the directory which is configured as output dir (current dir by default).
If you mean console output as well, then it would include some things but I don't think it's valuable in that context.

PR #1425, opened 2021-06-28
Increment cvg_count by 1 with each successful run of 'storage config' command with 'cvg' option.

83bhp:
Two changes
1. Increment cvg_count by 1 with each successful run of 'storage config' command with 'cvg' option.
2. User data type of cvg_count as int for Pillar and as string for confstore

### DIFF
--- a/lr-cli/cortx_setup/commands/storage/config.py
+++ b/lr-cli/cortx_setup/commands/storage/config.py
@@ -190,7 +190,7 @@ class StorageEnclosureConfig(Command):
         if(key=='password'):
             value = self.encrypt_password(value)
 
-        if pillar_key_map[key] == "cvg":
+        if key == "cvg":
             value = str(value)
 
         self.logger.debug(f"Updating Cortx Confstore with key:{conf_key_map[key]} and value:{value}")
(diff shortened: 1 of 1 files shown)


sradhanand-pati (approved):

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -461,7 +465,13 @@ def run(self, **kwargs):
                     " are missing")
                 raise RuntimeError("Incomplete arguments provided")
 
-            self.update_pillar_and_conf('cvg', self.cvg_count)
+            current_cvg_count = Conf.get (
+                'node_info_index',
+                f'server_node>{self.machine_id}>storage>cvg_count'
+            )
+            if current_cvg_count is MISSED:
```
mazinamdar, line 472: I think we used MISSED for pillar data , we can use None

nitin-seagate, line 472: Probably we don't need this check?

nitin-seagate (approved):
Looks good to me...

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -189,6 +189,10 @@ def update_pillar_and_conf(self, key, value):
 
         if(key=='password'):
             value = self.encrypt_password(value)
+
+        if pillar_key_map[key] == "cvg":
```
nitin-seagate, line 193: This is an issue, we need to match key not value....

nitin-seagate (changes requested):

nitin-seagate (approved):
Looks good to me...

PR #1426, opened 2021-06-28
Create separate package for lr-cli, cortx_setup commands.

83bhp:


shailesh-vaidya (approved):
Looks good to me

PR #1454, opened 2021-07-01
Restructure the storage config command code to support various scenarios.

83bhp:


lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -259,205 +272,224 @@ def run(self, **kwargs):
...
+            # if self.enclosure_id:
+            #     # store ip and port in confstore
+            #     if user and password:
+            #         # store in confstore
+            if (
+                self.enclosure_id is not None
+                and user is not None
+                and password is not None
+            ):
+                # user has provided all the parameters that fetches the
```
nitin-seagate, line 316: Strange condition here? Why would user still reset enclosure_id

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -259,205 +272,224 @@ def run(self, **kwargs):
...
-
-                self.refresh_key_map()
-
-                # store enclosure_id
-                self.store_in_file()
-                self.update_pillar_and_conf('enclosure_id', self.enclosure_id)
 
-                # store user
+        if user is not None or password is not None:
+            if (user and not password) or (password and not user):
```
nitin-seagate, line 378: Please write simple logic to check example
`if user is None or password is None:`
`Please provide 'user' and 'passowrd' together"`

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -259,205 +272,224 @@ def run(self, **kwargs):
...
-                # store ip and port as primary
-                self.update_pillar_and_conf('ip', ip)
-                self.update_pillar_and_conf('port', port)
+        if ip is not None or port is not None:
+            if self.mode is None:
+                self.logger.exception(
+                    f"mode is missing, please provide --mode argument"
+                )
+                raise RuntimeError("Incomplete arguments provided")
             else:
```
nitin-seagate, line 433: Why do  we need else block, please see if this can be handled in outside ese block itself

lr-cli/cortx_setup/commands/storage/config.py
```diff
@@ -259,205 +272,224 @@ def run(self, **kwargs):
...
-                    raise RuntimeError("Imcomplete arguments provided")
+                # This is already handled in 'mode' case
+                pass
+
+        if controller_type is not None:
+            valid_ctrl_type = ['gallium', 'indium']
+            if setup_type == "HW" and controller_type not in valid_ctrl_type:
+                self.logger.error(
+                "Invalid controller provided, please provide the"
+                " supported controller [gallium or indium]")
```
nitin-seagate, line 442: Lets not hard-code it in message, the valid controller types might change later as well

nitin-seagate (approved):
Looks good, see if you can improvise many of if-else conditions (Looks very complicated)...
Hope you have tested thoroughly as case scenarios

lr-cli/cortx_setup/commands/storage/config.py
83bhp: This is if user runs the same command again. In that case the enclosure id would be fetched from confstore and use, but to fetch the fresh enclosure id the enclosure id needs to be reset. This is what this condition is about.

lr-cli/cortx_setup/commands/storage/config.py
83bhp: Addressed.

lr-cli/cortx_setup/commands/storage/config.py
83bhp: It does nothing, just for readability and to understand the code better.

lr-cli/cortx_setup/commands/storage/config.py
83bhp: Addressed

PR #1462, opened 2021-07-02
WhiteSource Local config file

mukul-seagate11:
WhiteSource Local config file will be created to scan & detect vulnerabilities based on local component configurations/Artifacts/code extensions etc.


cortx-admin:

PR #1471, opened 2021-07-05
EOS-22128: remove ldap install step from provisioner

divya-vijay:


sachitanands:
@divya-vijay Could you please verify 1 Node and 3 Node deployment with this change . just to confirm nothing is broken because of this change.
Does this affect on s/w upgrade scenario ?

cortx-admin:

PR #1474, opened 2021-07-05
Fixes related to resource discover and show api

sradhanand-pati:
This PR has fixes for -
1. enclosure password encryption into confstore was using wrong key. should have been 'storage_enclosure' instead of 'storage'
2. not parsing request id received from resource discovery
3. removing srvnode-1 mapping in etc hosts with public data ip


83bhp (approved):

mazinamdar (approved):
looks good to me

nitin-seagate (approved):
Looks good to me...

PR #1480, opened 2021-07-08
Fixing node finalize for health map validation

sradhanand-pati:

### DIFF
--- a/lr-cli/cortx_setup/commands/node/finalize.py
+++ b/lr-cli/cortx_setup/commands/node/finalize.py
@@ -15,7 +15,7 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-from cortx_setup.config import CERT_PATH, HEALTH_PATH, MANIFEST_PATH
+from cortx_setup.config import CERT_PATH, HEALTH_PATH # , MANIFEST_PATH
 from provisioner.salt import local_minion_id
 from provisioner.values import MISSED
 from cortx_setup.validate import CortxSetupError, interfaces, disk_devices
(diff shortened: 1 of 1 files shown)


lr-cli/cortx_setup/commands/node/finalize.py
```diff
@@ -38,7 +38,7 @@ def _validate_health_map(self):
         self.logger.debug("Validating node health check")
         resource_paths = [HEALTH_PATH, MANIFEST_PATH]
         for resource in resource_paths:
-            path = get_pillar_data(resource)
+            path = get_pillar_data(resource).split('://')[1]
```
mazinamdar, line 41: can  you add this split after below check, if path is present then only do it.

lr-cli/cortx_setup/commands/node/finalize.py
```diff
@@ -38,7 +38,7 @@ def _validate_health_map(self):
         self.logger.debug("Validating node health check")
         resource_paths = [HEALTH_PATH, MANIFEST_PATH]
```
mazinamdar, line 39: can you remove MANIFEST_PATH as we don't have manifest in place

sradhanand-pati: Okay.

lr-cli/cortx_setup/commands/node/finalize.py
sradhanand-pati: Yes. i have moved the split after the path check..

mazinamdar (approved):
looks good to me

83bhp (changes requested):
Minor comment. Otherwise looks good.

83bhp (approved):

PR #1482, opened 2021-07-08
EOS-22258: Added Validation check in case of blank in device list

sradhanand-pati:


mazinamdar (approved):
looks good to me

83bhp (approved):

PR #1488, opened 2021-07-12
EOS-22429: Added Alex workflow

venkuppu-chn:
Signed-off-by: Venkatesh K <venkatesh.k@seagate.com>


83bhp (changes requested):
Hi @venkuppu-chn 
Please create a PR against pre-cortx-1.0 branch.
With the current workflow the patch first needs to be there in pre-cortx-1.0 and then it will be propagated to main branch.

83bhp (approved):

PR #1500, opened 2021-07-15
EOS-22267: Create users as part of node finalize

divya-vijay:

### DIFF
--- a/lr-cli/cortx_setup/commands/node/finalize.py
+++ b/lr-cli/cortx_setup/commands/node/finalize.py
@@ -16,7 +16,7 @@
 #
 
 from cortx_setup.config import CERT_PATH, HEALTH_PATH # , MANIFEST_PATH
-from provisioner.salt import local_minion_id
+from provisioner.salt import local_minion_id, StateFunExecuter
 from provisioner.values import MISSED
 from cortx_setup.validate import CortxSetupError, interfaces, disk_devices
 from ..command import Command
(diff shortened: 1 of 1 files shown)


lr-cli/cortx_setup/commands/node/finalize.py
```diff
@@ -35,86 +35,189 @@ class NodeFinalize(Command):
...
     def run(self, force=False):
-        local_minion = local_minion_id()
-        self._validate_cert_installation()
-        self._validate_health_map()
-        self._validate_interfaces(local_minion)
-        s_type = self._validate_server_type(local_minion)
-        self._validate_devices(local_minion, s_type)
-        self.logger.debug("Done")
+        node_id = local_minion_id()
+        val_1 = self._validate_cert_installation()
```
nitin-seagate, line 202: Lets use better names for variables:
val_1 change to validate_certificate
val_2 change to validate_healthmap
val_3 change to validate_interfaces
s_type change to server_type
val_4 change to validate_devices

OR

Instead use case as follows for each of the above checks (For example):
if self._validate_cert_installation():
    # Certificate validation success
elif force:
    # Certificate validation failed but skipped because of force
else:
    # Certificate validation failed (Throw exception)

lr-cli/cortx_setup/commands/node/finalize.py
```diff
@@ -35,86 +35,189 @@ class NodeFinalize(Command):
...
         cert_file = None
         cert_file = Path(CERT_PATH / 'stx.pem')
         if not cert_file.is_file():
-            raise CortxSetupError(
-                f"Validation failed: "
+            check = False
+            self.logger.error(
+                "Validation failed: "
                 f"Cert file not present {cert_file}"
             )
```
nitin-seagate, line 64: After logging error lets return... Why are we printing error as well as success message

lr-cli/cortx_setup/commands/node/finalize.py
```diff
@@ -35,86 +35,189 @@ class NodeFinalize(Command):
...
         self.logger.debug("Certificate installtion check: Success")
+        return check
```
nitin-seagate, line 66: Why not return two values:
**(True | False), "String Message"**
The String Message will be success or error message

lr-cli/cortx_setup/commands/node/finalize.py
```diff
@@ -35,86 +35,189 @@ class NodeFinalize(Command):
...
+        node_id = local_minion_id()
+        val_1 = self._validate_cert_installation()
+        val_2 = self._validate_health_map()
+        val_3 = self._validate_interfaces(node_id)
+        s_type = self._validate_server_type(node_id)
+        val_4 = self._validate_devices(node_id, s_type)
+
+        if (val_1 and val_2 and val_3 and val_4):
+            self.logger.info(
+                "Node validations complete. Proceeding to create users.."
```
nitin-seagate, line 210: We should know exactly what failed?

lr-cli/cortx_setup/commands/node/finalize.py
```diff
@@ -35,86 +35,189 @@ class NodeFinalize(Command):
...
+            )
+        else:
+            if force:
+                self.logger.info(
+                    "One or more node validation(s) failed. Forcibly creating users.."
+                )
+            else:
+                raise CortxSetupError(
+                    "One or more node validation(s) failed. Check logs for more details."
```
nitin-seagate, line 219: We should raise actual Exception exactly what failed?

nitin-seagate (changes requested):

lr-cli/cortx_setup/commands/node/finalize.py
divya-vijay: @nitin-seagate yes sure! this was initial draft to confirm if the flow is correct. Addressed now. Thanks

lr-cli/cortx_setup/commands/node/finalize.py
divya-vijay: Addressed

lr-cli/cortx_setup/commands/node/finalize.py
divya-vijay: Addressed

lr-cli/cortx_setup/commands/node/finalize.py
divya-vijay: Addressed

lr-cli/cortx_setup/commands/node/finalize.py
divya-vijay: Addressed

lr-cli/cortx_setup/commands/node/finalize.py
```diff
@@ -102,19 +110,111 @@ def _validate_devices(self, minion_id, s_type):
...
+            if force:
+                self.logger.info(
+                    f"One or more node validation(s) failed: {str(exc)}.\n"
+                    "Forcibly creating users.."
+                )
+                self._create_field_users(node_id)
+                self.logger.debug(
+                  "Field users created. Check logs for more details on the validations error..")
+            else:
+                raise
```
nitin-seagate, line 220: No new line at the end of file...

lr-cli/cortx_setup/commands/node/finalize.py
nitin-seagate: This comment is not addressed

nitin-seagate (changes requested):
Please address comments

lr-cli/cortx_setup/commands/node/finalize.py
divya-vijay: Earlier, the final validations were written directly to `run` method and exception was thrown when an error occurs. 
Now, we're handling all the validations in `try`-`except` block, and if there's any error, it will check for `force` param and if not, only then will raise an exception to the user.

lr-cli/cortx_setup/commands/node/finalize.py
divya-vijay: yes, added.

mazinamdar (approved):
looks good to me

83bhp (approved):

PR #5998, opened 2021-07-26
EOS-22259 Populate DIX Pool layout parameters to confstore file

zahidsk:
Signed-off-by: zahid.shaikh <zahid.shaikh@seagate.com>

### DIFF
--- a/lr-cli/cortx_setup/commands/storageset/config/durability.py
+++ b/lr-cli/cortx_setup/commands/storageset/config/durability.py
@@ -84,24 +84,24 @@ class DurabilityConfig(Command):
                    "First, set with `cortx_setup storageset create` command."
                 )
             durability_key = f'cluster>{cluster_id}>storage_set[0]>durability'
-            d_data = Conf.get(index, durability_key)
-            if not d_data.get(durability_type, False):
-                d_data[durability_type] = {}
+            durability_info = Conf.get(index, durability_key)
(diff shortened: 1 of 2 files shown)


cortx-admin:

lr-cli/cortx_setup/commands/storageset/config/durability.py
```diff
@@ -76,34 +74,35 @@ def run(self, **kwargs):
                 f'json://{CONFSTORE_CLUSTER_FILE}'
             )
 
-            ss_name = Conf.get(index, f'cluster>{cluster_id}>storage_set[0]>name')
-
-            durability_dict = {durability_type: {}}
+            ss_name = Conf.get(index,
```
nitin-seagate, line 77: How will this work for multiple storage sets?
Tomorrow we may have 2-3 storage sets in which case only storage_set[0] will not work

nitin-seagate (changes requested):

lr-cli/cortx_setup/commands/storageset/config/durability.py
nitin-seagate: Lets leave storage-set for now

nitin-seagate (changes requested):
Not sure this changes are correct

nitin-seagate (approved):
Looks good to me

PR #6026, opened 2021-08-05
EOS-22847: Exclude cvg name from confstore if it's not defined in con…

83bhp:
…fig.ini


cortx-admin:

PR #6033, opened 2021-08-10
EOS-23430: cortx_setup node finalize command failed on 3 node VM depl…

tanujashinde0405:
…oyment.

Signed-off-by: Tanuja Shinde <tanuja.shinde@seagate.com>

### DIFF
--- a/lr-cli/cortx_setup/commands/node/finalize.py
+++ b/lr-cli/cortx_setup/commands/node/finalize.py
@@ -40,7 +40,6 @@ from provisioner.salt import local_minion_id, StateFunExecuter
 from provisioner.utils import run_subprocess_cmd
 from provisioner.values import MISSED
 
-
 class NodeFinalize(Command):
     _args = {
         'force': {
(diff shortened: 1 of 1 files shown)


cortx-admin:

83bhp (approved):

nitin-seagate (changes requested):
Why do we fetch machine-id and enclosure-id to decide if the environment is HW or VM...
Please fetch server-type whether its HW or VM

SomwanshiAnjali (approved):

nitin-seagate (approved):
Approved
@cdeshmukh @mariyappanp Ideally we should have health check for VM as well right?

cdeshmukh:
Yes, Health check on VM would be there, just that it wont return any data for hw resources. I believe we had tested this scenario too, @mariyappanp ?

mariyappanp:
> Yes, Health check on VM would be there, just that it wont return any data for hw resources. I believe we had tested this scenario too, @mariyappanp ?

**Yes**

PR #6038, opened 2021-08-11
EOS-23494:  Re-deployment fails on VM after running cortx_setup cluster reset --type all

tanujashinde0405:


srv/components/ha/corosync-pacemaker/reset.sls
```diff
@@ -0,0 +1,53 @@
...
+  service.dead:
+    - name: {{ serv }}
+    - enable: False
+{% endfor %}
+
+Remove pcs package:
+  cmd.run:
+    - name: "rpm -e --nodeps pcs pacemaker corosync fence-agents-ipmilan"
+
+# Remove configuration directory:
```
nitin-seagate, line 35: Please confirm if we have to remove config directory for services
`"corosync", "pacemaker", "pcsd"`

srv/components/misc_pkgs/consul/reset.sls
```diff
@@ -0,0 +1,72 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+include:
+  - components.misc_pkgs.consul.stop
+
+# Remove Consul:
```
nitin-seagate, line 21: Why are we not removing only consul without Hare dependency?

srv/components/system/teardown.sls
```diff
@@ -20,10 +20,6 @@ Remove cortx-py-utils dependencies:
     - requirements: salt://components/system/files/cortx_py_utils_requirements.txt
     - bin_env: /usr/bin/pip3
 
-Remove cortx-py-utils pip package:
```
nitin-seagate, line 23: Why are we removing this from teardown? I think we are uninstalling all the RPMs for teardown right...

nitin-seagate (changes requested):

srv/components/system/teardown.sls
tanujashinde0405: we should not be removing foundation component rpm right?

srv/components/ha/corosync-pacemaker/reset.sls
tanujashinde0405: We are doing it as part of the same file.

srv/components/system/teardown.sls
tanujashinde0405: this should be done for pre-factory and it is already there in foundation mini provisioner call, this was redundant state which was written. so I have removed it.

83bhp (approved):

nitin-seagate (approved):
LGTM

SomwanshiAnjali (approved):

PR #6141, opened 2021-09-15
Copy generated provisioer rpm as part of docker image creation

shailesh-vaidya:


cortx-admin:

PR #6336, opened 2022-02-24
Cortx-27656 feat: Cluster Upgrade operation shall validate cluster health status before proceeding with Rolling (Non disruptive)  upgrade.

kailashyadav11:
# Problem Statement
- As a healthy cluster is a prerequisite condition for a cluster upgrade operation to take place.
- Provided the upgrade is of type Non-Disruptive category where the cluster functionality is seamless and is unaffected to the upgrade operation.
- There shall be a validation of cluster health status before an upgrade is invoked.

# Design
-  https://seagate-systems.atlassian.net/wiki/spaces/PT/pages/785514832/CORTX+Software+Rolling+Upgrade+POC
[get_cluster_health_feature_testing.TXT](https://github.com/Seagate/cortx-prvsnr/files/8130096/get_cluster_health_feature_testing.TXT)

# Coding
   Checklist for Author
-  [x] Coding conventions are followed and code is consistent

# Testing 
  Checklist for Author
- [ ] Unit and System Tests are added
- [ ] Test Cases cover Happy Path, Non-Happy Path and Scalability
- [x] Testing was performed with RPM

# Impact Analysis
  Checklist for Author/Reviewer/GateKeeper
- [ ] Interface change (if any) are documented
- [x] Side effects on other features (deployment/upgrade)
- [x] Dependencies on other component(s)

# Review Checklist 
  Checklist for Author
- [x] JIRA number/GitHub Issue added to PR
- [x] PR is self reviewed
- [x] Jira and state/status is updated and JIRA is updated with PR link
- [x] Check if the description is clear and explained

# Documentation
  Checklist for Author
- [ ] Changes done to WIKI / Confluence page / Quick Start Guide


cortx-admin:

src/provisioner/const.py
```diff
@@ -20,3 +20,4 @@
 DEFAULT_LOG_LEVEL = "INFO"
 SUPPORTED_LOG_LEVELS = ["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
 RELEASE_INFO_URL = "yaml:///opt/seagate/cortx/RELEASE.INFO"
+DISRUPTIVE_UPGRADE = False
```
ujjwalpl, line 23: Please name disruptive as "cold upgrade". Its small and more meaningful. 

COLD_UPGRADE_KEY = 'COLD_UPGRADE'
COLD_UPGRADE_MODE = 'False'

src/provisioner/provisioner.py
```diff
@@ -282,6 +282,17 @@ def cluster_upgrade(cortx_conf_url: str, force_override: bool = False):
         Paramaters:
         [IN] CORTX Config URL
         """
+        # query to get cluster health
+        _disruptive_upgrade = os.getenv('DISRUPTIVE_UPGRADE', str(const.DISRUPTIVE_UPGRADE))
```
ujjwalpl, line 286:
```
cold_upgrade_mode = os.getenv(const.COLD_UPGRADE_KEY, const.COLD_UPGRADE_MODE).upper()
if cold_upgrade_mode is not "TRUE" and not CortxProvisioner.is_cluster_healthy(): 
    Log.error(....)
    return 1
# No need of else.
# No need to write info about cluster is healthy. It is assumed. 
# Rest of the code.

**Review by ujjwalpl — CHANGES_REQUESTED** [review]

File: src/provisioner/const.py (hunk 1 above)
Comment (nitin-seagate) [reply src/provisioner/const.py:23]: @ujjwalpl 
Instead can we have something like this:
SUPPORTED_UPGRADE_MODES = ["ROLLING", "COLD"]
DEFAULT_UPGRADE_MODE = "ROLLING"

File: src/provisioner/provisioner.py (hunk 2 above)
Comment (nitin-seagate) [reply src/provisioner/provisioner.py:286]: I agree with comment from Ujjwal, but can we check case based on upgrade_types (instead of boolean flag)
Something like this:
upgrade_mode = Fetch value
Check if upgrade_mode is COLD:

File: src/provisioner/provisioner.py [hunk 3]
```diff
@@ -282,6 +282,17 @@ def cluster_upgrade(cortx_conf_url: str, force_override: bool = False):
...
+        if _disruptive_upgrade.upper() != "TRUE":
+            _healthy, _status = CortxProvisioner._get_cluster_health()
```
nitin-seagate, line 288: Change this to below format:
is_healthy, resource_status = CortxProvisioner.get_health_status("cluster")

src/provisioner/provisioner.py
```diff
@@ -398,3 +409,13 @@ def _validate_provisioning_status(cortx_conf: MappedConf, node_id: str, apply_ph
         else:
             Log.info(msg)
         return validate_result[0], validate_result[1]
+
+    @staticmethod
+    def _get_cluster_health():
```
nitin-seagate, line 414: Semantic should be something like this:
`get_health_status(resource)`
So that same API can be used to pull status of any resource i.e cluster, pods etc...

nitin-seagate (changes requested):

src/provisioner/const.py
ujjwalpl: We should not avoid explicit concept for rolling upgrade. Rolling upgrade is just a method. By default upgrade is supposed to be rolling upgrade which will be run 99.99%. COLD is extreme case, enabled through an option.

src/provisioner/provisioner.py
ujjwalpl: For now this condition is like boolean i.e. if cold is disabled, skip health check. We can introduce if there are additional cases in future. For now this is the only exception from the default execution path.

src/provisioner/provisioner.py
ujjwalpl: To simply, we should just have it as 
healthy = CortxProvisioner.is_cluster_healthy() underneath it can make a call to HA for health_status etc.

ujjwalpl (changes requested):

src/provisioner/const.py
nitin-seagate: @ujjwalpl This is internal code logic, not exposed to user... By default it will always be rolling upgrade, with explicit option to execute disruptive upgrade using "cold" keyword...

src/provisioner/provisioner.py
nitin-seagate: I'm keeping an option open to support multiple upgrade types rather than just restricting to rolling and seamless...
With boolean we are adding special meaning to flag rather than keyword to define exact mode which can be exploited to switch case later...

src/provisioner/provisioner.py
nitin-seagate: @ujjwalpl I'm with this API, but guess we discussed that resource type should also be passed as an argument...

nitin-seagate (changes requested):

src/provisioner/provisioner.py
ujjwalpl: There is just one way to upgrade, i.e. Rolling, others are just flags to control. Rolling is not a flag. 

We can keep it this way
```
upgrade_mode = os.getenv(const.UPGRADE_MODE_KEY, const.UPGRADE_MODE).upper() 
if upgrade_mode is not "COLD" and not CortxProvisioner.is_cluster_healthy(): 
   Log.error(....) return 1 
# No need of else. 
# No need to write info about cluster is healthy. It is assumed. 
# Rest of the code.
```

src/provisioner/provisioner.py
ujjwalpl: We need a simple query "Is Cluster Healthy". Other queries can be built into future like below.

```
CortxProvisioer: 
    def is_cluster_healthy(): 
        CortxProvisioer._get_resource_health(resource="cortx") 

    def _get_resource_health(resource): 
        # Send to HA and obtain health
```

ujjwalpl (changes requested):

src/provisioner/provisioner.py
nitin-seagate: > There is just one way to upgrade, i.e. Rolling, others are just flags to control. Rolling is not a flag.
> 
> We can keep it this way
> 
> ```
> upgrade_mode = os.getenv(const.UPGRADE_MODE_KEY, const.UPGRADE_MODE).upper() 
> if upgrade_mode is not "COLD" and not CortxProvisioner.is_cluster_healthy(): 
>    Log.error(....) return 1 
> # No need of else. 
> # No need to write info about cluster is healthy. It is assumed. 
> # Rest of the code.
> ```

I'm fine with this approach

src/provisioner/provisioner.py
nitin-seagate: > Those are internals of the implementation for which arg can be avoided.
> 
> ```
> CortxProvisioer: 
>     def is_cluster_healthy():
>         CortxProvisioer._get_resource_health(resource="cortx")
> 
>     def _get_resource_health(resource):
>        # Send to HA and obtain health
> ```

Agree with this...

src/provisioner/provisioner.py
nitin-seagate: > def is_cluster_healthy(): 
>         CortxProvisioer._get_resource_health(resource="cortx")

Agree with this...

src/provisioner/const.py
nitin-seagate: Lets keep these as below (instead of adding special meaning to boolean values):
`SUPPORTED_UPGRADE_MODES = ["ROLLING", "COLD"]`
`DEFAULT_UPGRADE_MODE = "ROLLING"`

nitin-seagate (changes requested):

src/provisioner/const.py
kailashyadav11: UPGRADE_MODE_KEY = "UPGRADE_MODE"
UPGRADE_MODE = ""

done

src/provisioner/provisioner.py
kailashyadav11: done

src/provisioner/provisioner.py
kailashyadav11: done

src/provisioner/provisioner.py
kailashyadav11: done

src/provisioner/provisioner.py
```diff
@@ -400,3 +405,16 @@ def _validate_provisioning_status(cortx_conf: MappedConf, node_id: str, apply_ph
         else:
             Log.info(msg)
         return validate_result[0], validate_result[1]
+    
+    @staticmethod
+    def is_cluster_healthy():
+        # Checks if cluster is healthy
+        CortxProvisioner._get_resource_health(resource="cortx")
```
ujjwalpl, line 412: This should have the old code i.e. 
```
    health = CortxProvisioner._get_resource_health(resource="cortx")
    return True if health == "OK" else False
```

src/provisioner/provisioner.py
```diff
@@ -400,3 +405,16 @@ def _validate_provisioning_status(cortx_conf: MappedConf, node_id: str, apply_ph
...
+        return True
+
+    @staticmethod
+    def _get_resource_health(resource='cortx'):
```
ujjwalpl, line 416: _get_resource_health is a generic function and must not have a default. 

```
def _get_resource_health(resource: str):
        """API call to get cluster health."""
        # TODO Make a call to HA Health API to get the resource status
        return "OK"
```

ujjwalpl (changes requested):

src/provisioner/provisioner.py
kailashyadav11: done

src/provisioner/provisioner.py
kailashyadav11: done

ujjwalpl (approved):

src/provisioner/const.py
```diff
@@ -20,3 +20,5 @@
 DEFAULT_LOG_LEVEL = "INFO"
 SUPPORTED_LOG_LEVELS = ["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
 RELEASE_INFO_URL = "yaml:///opt/seagate/cortx/RELEASE.INFO"
+UPGRADE_MODE_KEY = "UPGRADE_MODE"
+UPGRADE_MODE = ""
```
nitin-seagate, line 24: Make this:
`UPGRADE_MODE_KEY`
`UPGRADE_MODE_VAL`

nitin-seagate (changes requested):

src/provisioner/const.py
kailashyadav11: done

tanujashinde0405 (approved):

